PR #3, opened 2016-03-02
Add diagnostics for symbolization without mapping filenames

rauls5382:
Add diagnostic directing the user to pass the binary filename to pprof if it is not available and it is needed for local symbolization.

### DIFF
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -272,7 +272,8 @@ func newMapping(prof *profile.Profile, obj plugin.ObjTool, ui plugin.UI, force b
 		mappings[l.Mapping] = true
 	}
 
-	for _, m := range prof.Mapping {
+	missingBinaries := false
+	for mix, m := range prof.Mapping {
 		if !mappings[m] {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


internal/symbolizer/symbolizer.go
```diff
@@ -272,7 +272,8 @@ func newMapping(prof *profile.Profile, obj plugin.ObjTool, ui plugin.UI, force b
 		mappings[l.Mapping] = true
 	}
 
-	for _, m := range prof.Mapping {
+	missingBinaries := false
+	for mix, m := range prof.Mapping {
```
tippjammer: "mix" is confusing.  how about s/mix/midx/

internal/symbolizer/symbolizer.go
```diff
@@ -282,9 +283,19 @@ func newMapping(prof *profile.Profile, obj plugin.ObjTool, ui plugin.UI, force b
 			continue
 		}
 
+		if m.File == "" {
```
tippjammer: What if File is populated but doesn't exist?

rauls5382: We will diagnose missing filenames when we actually try to open them below (line 302).
Previously if the filename was empty we would silently skip it.

PR #4, opened 2016-03-04
Add local symbolization using llvm-symbolizer

rauls5382:
The llvm-symbolizer tool part of llvm appears to be much faster at symbolization than addr2line, and
uses symbol table information when available.

Will still fallback to addr2line if llvm-symbolizer isn't found.

### DIFF
--- a/internal/binutils/addr2liner_llvm.go
+++ b/internal/binutils/addr2liner_llvm.go
@@ -60,7 +60,7 @@ func (a *llvmSymbolizerJob) close() {
 
 // newLlvmSymbolizer starts the given llvmSymbolizer command reporting
 // information about the given executable file. If file is a shared
-// library, base should be the address at which is was mapped in the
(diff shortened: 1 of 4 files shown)


internal/binutils/addr2liner_llvm.go
```diff
@@ -0,0 +1,170 @@
...
+
+// close releases any resources used by the llvmSymbolizer object.
+func (a *llvmSymbolizerJob) close() {
+	a.in.Close()
+	a.cmd.Wait()
+}
+
+// newLlvmSymbolizer starts the given llvmSymbolizer command reporting
+// information about the given executable file. If file is a shared
+// library, base should be the address at which is was mapped in the
```
hk-cho: typo: is -> it

rauls5382: Fixed

PR #18, opened 2016-04-22
Add mechanism for extensions to set variable defaults

rauls5382:
This enables pprof extensions to update the default for a variable
before the command line is processed.


internal/driver/commands.go
```diff
@@ -68,6 +68,14 @@ func AddCommand(cmd string, format int, post PostProcessor, desc, usage string)
 	pprofCommands[cmd] = &command{format, post, false, desc, usage}
 }
 
+// SetDefault sets the default value for a pprof variable. This enables
```
josef-jelinek: s/SetDefault/SetVariableDefault

rauls5382:
Fixed the comments. Thanks

PR #23, opened 2016-06-20
Fix tree trimming

wade-k:


internal/graph/graph.go
```diff
@@ -332,6 +335,61 @@ nextSample:
 	return selectNodesForGraph(nodes, o.DropNegative)
 }
 
+// Trims a Graph that is in forest form to contain only the nodes in kept. This
+// will not work correctly in the case that a node has multiple parents.
+func (g *Graph) TrimTree(kept NodeSet) *Graph {
```
rauls5382: Can you receive a map[*Node]bool as a parameter instead of a NodeSet? That way we wouldn't need to change NodeSet.

rauls5382: There is no need to return a *Graph, and in fact it makes the interface confusing:
- Does it return a new graph? Does it update the incoming graph?

internal/graph/graph.go
```diff
@@ -332,6 +335,61 @@ nextSample:
...
+	// Creates a new list of nodes
+	oldNodes := g.Nodes
+	g.Nodes = make(Nodes, 0, len(kept.Ptr))
+
+	for _, cur := range oldNodes {
+		// A node may not have multiple parents
+		if len(cur.In) > 1 {
+			fmt.Fprintf(os.Stderr, "ERROR: TrimTree only works on trees.\n")
```
rauls5382: Use panic

internal/graph/graph.go
```diff
@@ -332,6 +335,61 @@ nextSample:
...
+		}
+
+		// Get the parent. Since cur.In may only be of size 0 or 1, parent will be
+		// equal to either nil or the only node in cur.In
+		var parent *Node
+		for _, edge := range cur.In {
+			parent = edge.Src
+		}
+
+		if parent != nil {
```
rauls5382: It is a bit more readable to do the exceptional cases first and continue. That way the interesting case flows down.
In this case I'd recommend:

if parent == nil {
  //delete in edges
  continue
}
// update the edges

internal/graph/graph.go
```diff
@@ -332,6 +335,61 @@ nextSample:
...
+				child := outEdge.Dest
+
+				delete(child.In, cur)
+				child.In[parent] = outEdge
+				parent.Out[child] = outEdge
+
+				outEdge.Src = parent
+				outEdge.Residual = true
+				// Any reconfigured edge can no longer be Inline.
+				outEdge.Inline = false
```
rauls5382: Can we mark it as inline if both edges were inline?

internal/graph/graph.go
rauls5382: The check could be: if len(cur.In) == 0 {
...

And then you get the parent only after the continue

internal/graph/graph.go
```diff
@@ -332,6 +334,65 @@ nextSample:
...
+// will not work correctly in the case that a node has multiple parents.
+func (g *Graph) TrimTree(kept nodePtrSet) {
+	// Creates a new list of nodes
+	oldNodes := g.Nodes
+	g.Nodes = make(Nodes, 0, len(kept))
+
+	for _, cur := range oldNodes {
+		// A node may not have multiple parents
+		if len(cur.In) > 1 {
+			panic(errors.New("TrimTree only works on trees.\n"))
```
rauls5382: This is sufficient:
 panic("TrimTree only works on trees")

internal/graph/graph.go
```diff
@@ -601,8 +683,22 @@ func (g *Graph) SortNodes(cum bool, visualMode bool) {
 	}
 }
 
+// selectTopNodePtrs returns a set of the top maxNodes *Node in a graph.
```
rauls5382: Comment Should start with "SelectTopNodePtrs"

internal/graph/graph.go
```diff
@@ -538,15 +599,36 @@ func (g *Graph) DiscardLowFrequencyNodes(nodeCutoff int64) NodeSet {
...
-	kept := make(NodeSet, len(nodes))
+	cutNodes := getNodesWithCumCutoff(nodes, nodeCutoff)
+	kept := make(NodeSet, len(cutNodes))
+	for _, n := range cutNodes {
+		kept[n.Info] = true
+	}
+	return kept
+}
+
+// Returns all the nodes who have a Cum value greater than or equal to cutoff
```
rauls5382: Comment should start with "getNodesWithCumCutoff returns..."

internal/graph/graph.go
```diff
@@ -538,15 +599,36 @@ func (g *Graph) DiscardLowFrequencyNodes(nodeCutoff int64) NodeSet {
...
+func getNodesWithCumCutoff(nodes Nodes, nodeCutoff int64) Nodes {
```
rauls5382: Call this getNodesAboveCumCutoff

internal/graph/graph.go
```diff
@@ -144,6 +145,8 @@ type NodeMap map[NodeInfo]*Node
 // NodeSet maps is a collection of node info structs.
 type NodeSet map[NodeInfo]bool
 
+type nodePtrSet map[*Node]bool
```
rauls5382: Let's export this type. It is awkward to have exported functions return unexported types

rauls5382: Add a comment explaining where this type is useful: To hold sets of nodes for graphs created with the call_tree option, which may have duplicate nodeInfo.

internal/graph/graph_test.go
```diff
@@ -0,0 +1,303 @@
...
+		debug += "\n"
+		debug += fmt.Sprintf("===  Node %d: %p  ===\n", i, node.Node)
+		debug += edgeMapsDebugString(node.In, node.Out)
+	}
+	return debug
+}
+
+// Checks if two edges are equal
+func edgesEqual(this, that *Edge) bool {
+	return this.Src == that.Src && this.Dest == that.Dest &&
```
rauls5382: Should we compare the edge weights?

If so, you can just do:
 *this == *that

internal/symbolizer/symbolizer.go
```diff
@@ -292,7 +292,7 @@ func newMapping(prof *profile.Profile, obj plugin.ObjTool, ui plugin.UI, force b
 			missingBinaries = true
 			continue
 		}
-		
+
```
rauls5382: Unnecessary change

internal/graph/graph_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	for i, node := range Expected {
+		debug += "\n"
+		debug += fmt.Sprintf("===  Node %d: %p  ===\n", i, node.Node)
+		debug += edgeMapsDebugString(node.In, node.Out)
+	}
+	return debug
+}
+
+// Checks if all the edges in this equal all the edges in that.
```
rauls5382: Please start all function comments with the function name.
https://golang.org/doc/effective_go.html#commentary

PR #27, opened 2016-06-29
Speed up proto postprocessing phase in proto decoder

rauls5382:
The proto decoder associates creates pointers to reference each
location/function/mapping based on their id. These ids can be
arbitrary uint64s, but often they are generated in sequence from
1 to N.

The overhead of keeping these indices in a hash is about a 20% of
the cost of decoding a profile. Speed it up by using an array to
track values from 1 to N, and a hash for values outside that range.

### DIFF
--- a/profile/encode.go
+++ b/profile/encode.go
@@ -16,7 +16,6 @@ package profile
 
 import (
 	"errors"
-	"fmt"
 	"sort"
 )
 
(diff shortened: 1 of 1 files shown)


profile/encode.go
```diff
@@ -231,34 +230,54 @@ var profileDecoder = []decoder{
...
-	functions := make(map[uint64]*Function)
+	functions := make(map[uint64]*Function, len(p.Function))
+	functionIds := make([]*Function, len(p.Function)+1)
 	for _, f := range p.Function {
 		f.Name, err = getString(p.stringTable, &f.nameX, err)
 		f.SystemName, err = getString(p.stringTable, &f.systemNameX, err)
 		f.Filename, err = getString(p.stringTable, &f.filenameX, err)
+		if f.ID < uint64(len(functionIds)) {
+			functionIds[f.ID] = f
+			continue
```
josef-jelinek: in this case, it seems to me that explicit if - else is both more readable and safer as it is easier to add code after the blocks or move blocks without reevaluating the loop's short-circuit logic (as demonstrated in lines 261-265).

rauls5382: Thanks. Done.

PR #30, opened 2016-07-21
Instruction-level granularity in callgrind output

prattmic:
When generating callgrind format output, produce cost lines at
instruction granularity. This allows visualizers supporting the
callgrind format to display instruction-level profiling information.

We also need to provide the object file (ob=) in order for tools to find
the object file to disassemble when displaying assembly.

We opportunistically group cost lines corressponding to the same
function together, reducing the number of superfluous description lines.
Subposition compression (relative position numbering) is also used to
reduce the output size.

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -130,12 +130,15 @@ func applyCommandOverrides(cmd []string, v variables) variables {
 	case "proto", "raw":
 		trim, focus, tagfocus, hide = false, false, false, false
 		v.set("addresses", "t")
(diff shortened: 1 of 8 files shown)


prattmic:
I signed it!

rauls5382:
The driver test fails with this change. Please update the verification files.

To repro:
go test github.com/google/pprof/internal/driver

prattmic:
Oops, I made one last change and forgot to update the tests. Fixed now.

rauls5382:
Thank you

PR #32, opened 2016-08-18
Adds hook to allow pprof to handle perf.data files.

wade-k:


internal/driver/fetch.go
```diff
@@ -357,7 +360,7 @@ func fetch(source string, duration, timeout time.Duration, ui plugin.UI) (p *pro
 		f, err = fetchURL(sourceURL, timeout)
 		src = sourceURL
 	} else {
-		f, err = os.Open(source)
+		f, err = profileProtoReader(source)
```
rauls5382: Have profileProtoReader returns a *profile.Profile

Otherwise it isn't really a "proto reader"

internal/driver/fetch.go
```diff
@@ -379,6 +382,74 @@ func fetchURL(source string, timeout time.Duration) (io.ReadCloser, error) {
...
+
+	// If the file is the output of a perf record command, it should begin
+	// with the string PERFILE2.
+	perfHeader := []byte("PERFILE2")
+	actualHeader := make([]byte, len(perfHeader))
+	_, readErr := sourceFile.Read(actualHeader)
+	_, seekErr := sourceFile.Seek(0, 0)
+	if seekErr != nil {
+		return nil, seekErr
+	} else if readErr != nil && readErr != io.EOF{
```
rauls5382: No else after return (also below)

internal/driver/fetch.go
```diff
@@ -379,6 +382,74 @@ func fetchURL(source string, timeout time.Duration) (io.ReadCloser, error) {
...
+// using the perf_to_profile tool. It prints the stderr and stdout of
+// perf_to_profile to the stderr and stdout of this process, then returns the
+// path to a file containing the profile.proto formatted data.
+func convertPerfData(perfPath string) (string, error) {
+	randomBytes := make([]byte, 32)
+	_, randErr := rand.Read(randomBytes)
+	if randErr != nil {
+		return "", randErr
+	}
+	randomFileName := "/tmp/pprof_" +
```
rauls5382: Unlink the file after you're done with it, and in case of errors.

internal/driver/fetch.go
```diff
@@ -379,6 +382,74 @@ func fetchURL(source string, timeout time.Duration) (io.ReadCloser, error) {
...
+	// with the string PERFILE2.
+	perfHeader := []byte("PERFILE2")
+	actualHeader := make([]byte, len(perfHeader))
+	_, readErr := sourceFile.Read(actualHeader)
+	_, seekErr := sourceFile.Seek(0, 0)
+	if seekErr != nil {
+		return nil, seekErr
+	} else if readErr != nil && readErr != io.EOF{
+		return nil, readErr
+	} else if string(actualHeader) == string(perfHeader) {
```
rauls5382: Use bytes.Equal

internal/driver/fetch.go
```diff
@@ -379,6 +382,74 @@ func fetchURL(source string, timeout time.Duration) (io.ReadCloser, error) {
...
+	randomBytes := make([]byte, 32)
+	_, randErr := rand.Read(randomBytes)
+	if randErr != nil {
+		return "", randErr
+	}
+	randomFileName := "/tmp/pprof_" +
+		base64.StdEncoding.EncodeToString(randomBytes)
+	cmd := exec.Command("perf_to_profile", perfPath, randomFileName)
+
+	stdout, stdoutErr := cmd.StdoutPipe()
```
rauls5382: Ignore the stdout/stderr. Perhaps print a substring of the stderr output in case of errors.

internal/driver/fetch.go
```diff
@@ -379,6 +382,74 @@ func fetchURL(source string, timeout time.Duration) (io.ReadCloser, error) {
...
+	actualHeader := make([]byte, len(perfHeader))
+	_, readErr := sourceFile.Read(actualHeader)
+	_, seekErr := sourceFile.Seek(0, 0)
+	if seekErr != nil {
+		return nil, seekErr
+	} else if readErr != nil && readErr != io.EOF{
+		return nil, readErr
+	} else if string(actualHeader) == string(perfHeader) {
+		sourceFile.Close()
+		profileFile, convertErr := convertPerfData(path)
```
rauls5382: Does the conversion take a while? If so, print a message using the ui plugin.

internal/driver/fetch.go
```diff
@@ -400,54 +399,43 @@ func profileProtoReader(path string) (io.ReadCloser, error) {
 	_, seekErr := sourceFile.Seek(0, 0)
 	if seekErr != nil {
 		return nil, seekErr
-	} else if readErr != nil && readErr != io.EOF{
+	}
+	if readErr != nil && readErr != io.EOF{
```
rauls5382: Is there a reason for doing the seek if the read failed? If so, add a comment.
if not, change by combining each check into the if:

if _, readErr := sourceFile.Read(actualHeader); err != nil { // ....

internal/driver/tempfile.go
```diff
@@ -21,16 +21,46 @@ import (
...
+// newTempFile returns a new file with a random name for output files.
+func newTempFile(dir, prefix, suffix string) (*os.File, error) {
+	path, err := newTempFilePath(dir, prefix, suffix)
+	if err != nil {
+		return nil, err
+	}
+	return os.Create(path)
+}
+
+type VolatileFile struct {
```
rauls5382: Why not use the deferDeleteTempFile mechanism below?

internal/driver/fetch.go
```diff
@@ -379,6 +381,63 @@ func fetchURL(source string, timeout time.Duration) (io.ReadCloser, error) {
...
+// profile.proto formatted data.
+func convertPerfData(perfPath string, ui plugin.UI) (string, error) {
+	ui.Print(fmt.Sprintf(
+		"Converting %s to a profile.proto... (May take a few minutes)",
+		perfPath))
+	profilePath, err := newTempFilePath("/tmp", "pprof_", ".pb.gz")
+	if err != nil {
+		return "", err
+	}
+	cmd := exec.Command("perf_to_profile", perfPath, profilePath)
```
rhysh: Where is this command available? It looks like it would be similar to https://godoc.org/rsc.io/perf2pprof, but with a different interface.

wade-k: It is not currently available, I am working on it and plan to opensource it

internal/driver/fetch.go
```diff
@@ -356,6 +358,13 @@ func fetch(source string, duration, timeout time.Duration, ui plugin.UI) (p *pro
 		}
 		f, err = fetchURL(sourceURL, timeout)
 		src = sourceURL
+	} else if isPerf, isPerfErr := isPerfFile(source); isPerf {
+		// Since the if statement is a new scope, if isPerfErr is named
```
rauls5382: If you have isPerfFile only return a bool (==false in case of errors) you'll save a lot of this complexity.

internal/driver/tempfile.go
```diff
@@ -21,16 +21,25 @@ import (
 	"sync"
 )
 
-// newTempFile returns an unused filename for output files.
-func newTempFile(dir, prefix, suffix string) (*os.File, error) {
+// newTempFilePath returns an unused path for output files.
+func newTempFilePath(dir, prefix, suffix string) (string, error) {
 	for index := 1; index < 10000; index++ {
```
rauls5382: Save some duplication by calling newTempFile and run os.File.Name to get the filename (don't forget to close the file).

PR #33, opened 2016-08-22
Add tagshow/taghide commands to select a subset of tags in a profile

rauls5382:


profile/filter.go
```diff
@@ -73,6 +73,36 @@ func (p *Profile) FilterSamplesByName(focus, ignore, hide, show *regexp.Regexp)
 	return
 }
 
+// FilterTagsByName filters the tags in a profile and only keeps
+// tags that match show and not hide.
+func (p *Profile) FilterTagsByName(show, hide *regexp.Regexp) (sm, hm bool) {
+	matchShow := func(name string) bool {
```
josef-jelinek: It is a bit difficult to read code with two matchShow variables with different meanings next to each other.
Maybe in this case, I would choose matchRemove for the function, then return !matchShow || matchHide and change later 2x if matchRemove(lab) { ... } so there are less negations to think in.

PR #40, opened 2016-10-07
Add support for abbreviated commands on interative command like

rauls5382:
Support commands like "top10" or "web5"

### DIFF
--- a/internal/driver/interactive.go
+++ b/internal/driver/interactive.go
@@ -17,6 +17,7 @@ package driver
 import (
 	"fmt"
 	"io"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/driver/interactive.go
```diff
@@ -218,7 +218,20 @@ func parseCommandLine(input []string) ([]string, variables, error) {
 	cmd, args := input[:1], input[1:]
 	name := cmd[0]
 
-	c := pprofCommands[cmd[0]]
+	c := pprofCommands[name]
+	if c == nil {
+		// Attempt splitting digits on abbreviated commands (eg top10)
+		for i := len(name); i > 0; i-- {
```
josef-jelinek: would not a regexp like
re := regexp.MustCompile("[0-9]+$")
match := re.FindStringSubmatch(name)
(and then test match == nil for fail or match[0] for found number) be simpler in this case?

josef-jelinek: Actually rather re.FindString and then test for length == 1 and match[1]..

PR #41, opened 2016-10-07
Reset mapping file to empty string if it was patched to be the source URL

aalexand:
When the source is remote and a mapping doesn't have either build ID or
file field set, the file field is set to the source URL so that the
proper source from the source mapping is used during symbolz processing.
Before this change, the file field would continue to point to the URL
producing the URL in pprof output which confuses users.

With the change it resets the file field back to the empty string. It
also now skips the URL-like paths during local symbolization as
reading at that path is not going to succeed.

We discussed switching to generating more unique IDs. On the second
thought, I propose leaving these to be URLs as that seems unique enough
and in case this field leaks into the tool or log output seeing the URL
seems still friendlier than some arbitrarily prefixed string.

### DIFF
--- a/internal/driver/fetch.go
+++ b/internal/driver/fetch.go
@@ -72,6 +72,7 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
 		return nil, err
 	}
 	p.RemoveUninteresting()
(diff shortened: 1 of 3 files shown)


aalexand:
Done.

PR #44, opened 2016-10-26
Redirect binary outputs away from stdout on interactive mode

rauls5382:


internal/driver/commands.go
```diff
@@ -443,7 +443,9 @@ func invokeVisualizer(format PostProcessor, suffix string, visualizers []string)
 				defer func(t <-chan time.Time) {
 					<-t
 				}(time.After(time.Second))
-				if waitForVisualizer {
+				// On interactive mode, let the visualizer run on the background
```
josef-jelinek: Nit: on -> in?

rauls5382: Done

PR #51, opened 2016-11-07
Allow profiles with no samples

rauls5382:
A previous commit tried to eliminate a panic when reading an empty
profile by stopping support for profiles with no samples. That is
unnecessary and it can obstruct some testing.

The panic was caused by the legacy parser returning a nil profile and
no error when processing an empty profile. Detect that situation and
generate a proper profile instead.

Tested by running 'pprof /dev/null'


rauls5382:
Added test for parsing failures. Without this fix, that test causes a panic

PR #58, opened 2016-12-01
Add source information to disassembly reports

rauls5382:
Disassembly reports generated by pprof -disasm will now include line number
information as generated by objdump. This will make the generated assembly more
readable.

As part of this I've introduced a new assemblyInstruction struct. Previously
the code was reusing the graph.Node to represent assembly instructions but it
seems better to have a dedicated type for this.

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -303,12 +303,12 @@ func printAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
 		flatSum, cumSum := sns.Sum()
 
 		// Get the function assembly.
-		insns, err := obj.Disasm(s.sym.File, s.sym.Start, s.sym.End)
+		insts, err := obj.Disasm(s.sym.File, s.sym.Start, s.sym.End)
 		if err != nil {
 			return err
 		}
 
-		ns := annotateAssembly(insns, sns, s.base)
+		ns := annotateAssembly(insts, sns, s.base)
 
 		fmt.Fprintf(w, "ROUTINE ======================== %s\n", s.sym.Name[0])
 		for _, name := range s.sym.Name[1:] {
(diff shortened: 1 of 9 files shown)


internal/report/source.go
```diff
@@ -204,7 +204,7 @@ func printWebSource(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
 
 // sourceCoordinates returns the lowest and highest line numbers from
 // a set of assembly statements.
-func sourceCoordinates(asm map[int]graph.Nodes) (start, end int) {
+func sourceCoordinates(asm map[int][]assemblyInstruction) (start, end int) {
```
aalexand: Nit (not quite in the scope of the PR): sourceLineRange/Span might be a better name.

internal/report/report.go
```diff
@@ -417,10 +446,34 @@ func nodesPerSymbol(ns graph.Nodes, symbols []*objSymbol) map[*objSymbol]graph.N
...
+	return a.cum
+}
+
 // annotateAssembly annotates a set of assembly instructions with a
 // set of samples. It returns a set of nodes to display. base is an
 // offset to adjust the sample addresses.
-func annotateAssembly(insns []plugin.Inst, samples graph.Nodes, base uint64) graph.Nodes {
+func annotateAssembly(insns []plugin.Inst, samples graph.Nodes, base uint64) []assemblyInstruction {
 	// Add end marker to simplify printing loop.
 	insns = append(insns, plugin.Inst{
```
aalexand: insts since the type is Inst?

aalexand: insts since the type is Inst?

rauls5382: Done here and in several other places.

internal/report/source.go
rauls5382: This is really an assembly instruction, so I prefer the current name.

PR #65, opened 2016-12-09
Combine adjacent mappings even if offsets are unavailable

rauls5382:
Some profile handlers will split mappings into multiple adjacent
entries. This interferes with pprof when overriding the main binary.

Currently there is code that attempts to combine these mappings, but
it fails if the mapping offsets aren't available. Add a check to
combine them in that situation.


rauls5382:
Some legacy tools may write the mappings in a format that does not include the offsets.

PR #66, opened 2016-12-09
Add a message recommending the use of PPROF_BINARY_PATH

rauls5382:
The common mechanism to locate the main binary has several drawbacks:

- The identification of the main mapping is done through heuristics
- It can only override the main binary, no libraries

PPROF_BINARY_PATH is a more robust mechanism, but it isn't widely known.
Add a message when symbolization fails recommending its usage.

### DIFF
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -290,8 +290,7 @@ func newMapping(prof *profile.Profile, obj plugin.ObjTool, ui plugin.UI, force b
 
 		if m.File == "" {
 			if midx == 0 {
-				ui.PrintErr("Main binary filename not available.\n" +
-					"Try passing the path to the main binary before the profile.")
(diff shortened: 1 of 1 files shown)


internal/symbolizer/symbolizer.go
```diff
@@ -313,7 +313,8 @@ func newMapping(prof *profile.Profile, obj plugin.ObjTool, ui plugin.UI, force b
 
 		f, err := obj.Open(m.File, m.Start, m.Limit, m.Offset)
 		if err != nil {
-			ui.PrintErr("Local symbolization failed for ", name, ": ", err)
+			ui.PrintErr("Local symbolization failed for ", name, ": ", err, "\n"+
+				"Try setting PPROF_BINARY_PATH to a path to search for local binaries.")
```
aalexand: Maybe it should only be printed once to avoid spamming the output?

PR #73, opened 2016-12-14
Add a helper function to check if a mapping is unsymbolizable

aalexand:
The local symbolization already skips these mappings. Have the helper which can be used when alternative symbolization flows are used.

### DIFF
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


profile/profile.go
```diff
@@ -565,7 +566,7 @@ func (p *Profile) ScaleN(ratios []float64) error {
 // symbolized function information.
 func (p *Profile) HasFunctions() bool {
 	for _, l := range p.Location {
-		if l.Mapping != nil && !l.Mapping.HasFunctions {
+		if l.Mapping != nil && !l.Mapping.Unsymbolizable() && !l.Mapping.HasFunctions {
```
rauls5382: Not sure that these checks belong in a function that is called HasFunctions()
Perhaps let the symbolizer call Unsymbolizable directly?

aalexand: Done.

PR #77, opened 2017-01-01
README: add file link and fix profile.proto file path

zchee:
Fix file path and add link for README.md.

- fix profile.proto file path
- add file link for profile.proto and some documentation

A happy new year!


rauls5382:
Thanks

PR #81, opened 2017-02-07
profile.parseGoCount: accept non-space characters in profile type.

Sajmani:
In addition to the "goroutine" and "threadcreate" profiles,
Go code can generate custom profiles using the runtime/pprof package.
The user must name these profiles, and the docs recommend using the
convention "import/path" to avoid namespace conflicts.  This CL
updates the pprof tool to be able to parse legacy profiles whose types
contain slashes and other non-space characters.

This is the upstream fix for https://github.com/golang/go/issues/13195.
This change will need to be mirrored to
github.com/golang/go/src/cmd/pprof/internal/profile/legacy_profile.go

### DIFF
--- a/profile/legacy_profile.go
+++ b/profile/legacy_profile.go
@@ -29,7 +29,7 @@ import (
 )
 
 var (
-	countStartRE = regexp.MustCompile(`\A(\w+) profile: total \d+\z`)
+	countStartRE = regexp.MustCompile(`\A(\S+) profile: total \d+\z`)
 	countRE      = regexp.MustCompile(`\A(\d+) @(( 0x[0-9a-f]+)+)\z`)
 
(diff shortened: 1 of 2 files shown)


aalexand:
Assigned Raul to take a look.  My preference would be to just add dot and slash to the allowed characters, rather than whitelist all non-whitespace characters.

Sajmani:
If we're going to allow ., we should also allow - and any other character
that might appear in an import path.  I figure there are arguments in favor
of other characters as well, so let's avoid the arbitrary restrictions and
just disallow spaces.

At the other end would be to just add /, and otherwise require path
components to look like Go identifiers.

Sajmani:
Yes, not allowing whitespace seems fine.  I'll update the runtime/pprof docs in the Go library.  Thanks!

PR #85, opened 2017-02-10
Merge commits from Go distribution

rauls5382:
These are the remaining relevant commits identified from the Go distribution.
The others in Hana's list are either already applied or not applicable.

### DIFF
--- a/internal/driver/fetch.go
+++ b/internal/driver/fetch.go
@@ -16,6 +16,7 @@ package driver
 
 import (
 	"bytes"
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


internal/driver/fetch.go
```diff
@@ -474,12 +475,26 @@ func adjustURL(source string, duration, timeout time.Duration) (string, time.Dur
...
+			InsecureSkipVerify: true,
+		}
+		url.Scheme = "https"
+		source = url.String()
+	}
+
 	client := &http.Client{
 		Transport: &http.Transport{
 			ResponseHeaderTimeout: timeout + 5*time.Second,
-			Proxy: http.ProxyFromEnvironment,
```
aalexand: This appears to undo https://github.com/google/pprof/commit/d87190945c4dace166442593ffe062aebd0a3c3d?

rauls5382: Good catch! Thanks

rauls5382:
yw. Thank you for digging and finding these.

I agree we need to combine the tests. I plan to do an overall review of the tests after we vendor the pprof code.

PR #86, opened 2017-02-10
Clean-up handling of output flag

dsnet:
For certain formats (eog, evince, gv, web, weblist), the visualizer
would be invoked regardless of whether the user specified a specific
output file to write to. In order to ensure that output is properly
written, the invokeVisualizer function has a check for whether the
output is os.Stdout and alters behavior based on that. This is the
wrong abstraction layer to do that work.

At the core of the problem is that PostProcess is used to both
do data post-processing that is inherent to the format, and also
to invoke some visualization for the data (to Stdout or browser).
We split these two steps apart and make it obvious which is which
by adding a visualizer field to Command.

This seperation of concerns allows us to simplify the code.

### DIFF
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -41,6 +41,7 @@ type commands map[string]*command
 type command struct {
 	format      int           // report format to generate
 	postProcess PostProcessor // postprocessing to run on report
+	visualizer  PostProcessor // display output using some callback
 	hasParam    bool          // collect a parameter from the CLI
 	description string        // single-line description text saying what the command does
 	usage       string        // multi-line help text saying how the command is used
@@ -64,7 +65,7 @@ func (c *command) help(name string) string {
 // specialized visualization formats. If the command specified already
 // exists, it is overwritten.
 func AddCommand(cmd string, format int, post PostProcessor, desc, usage string) {
-	pprofCommands[cmd] = &command{format, post, false, desc, usage}
+	pprofCommands[cmd] = &command{format, post, nil, false, desc, usage}
 }
 
 // SetVariableDefault sets the default value for a pprof
@@ -76,7 +77,7 @@ func SetVariableDefault(variable, value string) {
 }
 
 // PostProcessor is a function that applies post-processing to the report output
-type PostProcessor func(input []byte, output io.Writer, ui plugin.UI) error
+type PostProcessor func(input io.Reader, output io.Writer, ui plugin.UI) error
 
 // interactiveMode is true if pprof is running on interactive mode, reading
 // commands from its shell.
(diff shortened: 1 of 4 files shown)


dsnet:
This was originally https://go-review.googlesource.com/c/36698/, which ensures that setting the output flag will cause pprof to always dump to that file.

When I transfered that logic here, I noticed that there was a fix for this, by [adding a check in `invokeVisualizer` to check if the output is `os.Stdout`](https://github.com/google/pprof/blame/058580598d0dd053a2e14770f759937fa726a997/internal/driver/commands.go). I believe this is the wrong layer of abstraction to do the check.

internal/driver/commands.go
```diff
@@ -59,14 +60,6 @@ func (c *command) help(name string) string {
 	return message + "\n"
 }
 
-// AddCommand adds an additional command to the set of commands
-// accepted by pprof. This enables extensions to add new commands for
-// specialized visualization formats. If the command specified already
-// exists, it is overwritten.
-func AddCommand(cmd string, format int, post PostProcessor, desc, usage string) {
```
rauls5382: Please do not remove this interface. There are users of the pprof infrastructure that use it to introduce their own commands.

internal/driver/commands.go
```diff
@@ -360,70 +353,52 @@ var kcachegrind = []string{"kcachegrind"}
...
 		if err := cmd.Run(); err != nil {
 			return fmt.Errorf("Failed to execute dot. Is Graphviz installed? Error: %v", err)
 		}
-		return divert(buf.Bytes(), output, ui)
+		return nil
 	}
 }
 
-func saveSVGToFile() PostProcessor {
+func invokeDotSVG() PostProcessor {
```
rauls5382: Please add a comment explaining why this is different from invokeDot("svg"): mainly that it also massages the SVG to add the svgpan functionality.

Also maybe rename it, something like massageDotSVG, or invokeDotSvgpan ?

internal/driver/commands.go
dsnet: I don't understand. How is that possible? Isn't this under an internal package path?

rauls5382: This code is also used in some environments where the internal package rules are not enforced.

aalexand: @dsnet, Can you add the function back please?

dsnet: I added this back in. Note that the function signature is strange since it makes no distinction between a data post-processor and a visualizer.

I looked at the various uses cases of custom commands, and it seems to me that many of them have a distinct separation of data-processing vs visualization (or output storage).

internal/driver/commands.go
dsnet: Done.

aalexand:
@dsnet, could you squash the commits?

dsnet:
Done. You do realize that GitHub offers that feature now?

aalexand:
Didn't know that.  TIL.

PR #88, opened 2017-02-17
Detect invalid addresses when parsing profile and report as error

aalexand:


rauls5382:
This is fine, but is it addressing a practical issue? There are many other regexps in that file that use /d+ to match to a decimal, and [0-9a-f]+ to match a hex number. Should we restrict those as well?

An alternative would be to better handle error conditions when parsing the string into an integer -- I see we currently panic. Could we catch the error there instead?

aalexand:
The crash we are having has the stack below so the fix should address that. Other patterns could be updated, but I'm not sure whether making sure to tolerate any invalid input is worth it now.

I'll update the PR to recognize the error and bail out instead of skipping the address.  I had same though, but convinced myself it was not worth it as I wanted to stop the crashes first, but you convinced me otherwise now.

    0x3a727bb               pprof/profile/profile.extractHexAddresses+0x3a727bb
    0x3a72845               pprof/profile/profile.parseHexAddresses+0x3a72845
    0x3a754fe               pprof/profile/profile.parseThreadSample+0x3a754fe
    0x3a74a67               pprof/profile/profile.parseThread+0x3a74a67
    0x3a7b1f3               pprof/profile/profile.parseLegacy+0x3a7b1f3

aalexand:
Updated, how does it look?

PR #89, opened 2017-02-17
Allow binary override for profiles with no mappings

rauls5382:
The go runtime generates profiles in profile.proto format
without symbols or mappings. The expectation is that these
can be symbolized by passing the binary name to pprof.

The mechanism pprof uses relies to override the binary relies
on there being a mapping, and previously we moved the creation
of fake mappings to the legacy profile handlers, so profiles
parsed from profile.proto with no mappings can no longer be
symbolized.

Special case this situation to create a fake mapping and associate
all locations to it if there is a command line override but no
mappings.


aalexand:
The change looks good. Can you add a comment in the source code about this specificity of profiles generated by Go runtime?

As a more general question: these profiles won't symbolize for any shared objects or if the executable is PIC & ASLR'ed?

rauls5382:
That's correct, symbolization with a fake mapping is only possible with a non-ASLR binary.
Added comments explaining this.

PR #102, opened 2017-03-01
More govet fixes

rauls5382:
Surprisingly these don't show up with go vet, but still cause issues
for the Go distribution.

Related to https://github.com/golang/go/issues/19322


rauls5382:
Merging trivial change myself.

Issue #107, opened 2017-03-07
pprof tests leave orphaned files in $HOME/pprof 


rauls5382:
These should either not be generated, or generated in a test directory and cleaned up.

rsc:
People keep saying that, but I thought it was an intentional feature of pprof, dating back to the original perl script, to keep the downloaded profiles around after pprof exits, so that if you want to get back to them you can.

rauls5382:
Sorry, I wasn't clear. It is an intentional feature of pprof; I don't intend to change it.
I meant we shouldn't generate these **during tests** or clean them up after testing.

There is already some logic to save these files only when the profile is fetched remotely (if you pprof a local file we do not copy it into $HOME/pprof). I can probably extend that logic to disable the creation of local copies during testing.

aalexand:
Could we set HOME env var to point to a temporary directory in tests?

PR #108, opened 2017-03-08
pprof: clean up temp files created during tests

josharian:


aalexand:
Fixes https://github.com/google/pprof/issues/107 ?

rauls5382:
This is good, but it doesn't fix #107, which is about the files created in $HOME during tests.

PR #115, opened 2017-03-15
Enable continuous integration with Travis CI

aalexand:
This is to enable continuous build for the master and PRs.


.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: go
+go:
+  - 1.6
+  - 1.7
```
tamird: You should use .x on all of these, unless you really want to test with 1.6 and 1.7 (and not their highest patch levels).

PR #118, opened 2017-03-15
Clarify that Go 1.7 or newer is required.

aalexand:
Go 1.6 currently fails the tests with build error

`profile/legacy_profile_test.go:309: t.Run undefined (type *testing.T has no field or method Run)`

Alternatively, we can fix it but I am not sure there is a strong reason
to support Go 1.6 or older.

Also add build status badge in the README.md file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: go
 go:
-  - 1.6.x
   - 1.7.x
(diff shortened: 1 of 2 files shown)


aalexand:
@rauls5382 could you approve as well?  want to make sure you are OK with the change.

PR #119, opened 2017-03-15
Handle shared libraries with nonzero program headers.

gmarin13:
Generalize GetBase for shared libraries to handle cases when both process
mapping offset and program header addresses are non-zero.

Based on process mapping information, a sample at a virtual address x maps to
a file offset fx = x - map_start + map_offset.

The program header for the segment with the .text section may have non-zero
mapping information. Thus, a file offset fx maps to a symbol address
sx = fx - ph_offset + ph_virt_addr.

Thus, sx = x - map_start + map_offset - ph_offset + ph_virt_addr, and the base
address for symbolization is map_start - map_offset + ph_offset - ph_virt_addr.


internal/elfexec/elfexec.go
```diff
@@ -240,17 +240,10 @@ func GetBase(fh *elf.FileHeader, loadSegment *elf.ProgHeader, stextOffset *uint6
...
-			}
-			return 0, fmt.Errorf("Don't know how to handle mapping. Offset=%x, vaddr=%x",
-				offset, loadSegment.Vaddr)
-		}
 		if loadSegment == nil {
-			return start, nil
+			return start - offset, nil
 		}
-		return start - loadSegment.Vaddr, nil
+		return start - offset + loadSegment.Off - loadSegment.Vaddr, nil
```
aalexand: The logic looks correct to me.  I agree that loadSegment.Off should be accounted.  Basically, (start - offset) is the virtual address of where the file offset of zero is (as if the file is mapped into memory as a whole).  Then we add the offset of the segment from the program header which gives the virtual address of where the segment starts.  Then we subtract the virtual address of where the program headers the segment should start which gives the adjustment that we return.

In practice, the executable segment of shared libraries has typically file offset of zero I think which is why we didn't notice it until we started dealing with non-standard ELF files.

Can you put all of this reasoning into the code as a comment?  I think it's subtle enough and deserves more than just being a PR discussion.

gmarin13: I added some comments. PTAL.

gmarin13: I just pushed the commit again to fix something in the comment.

PR #123, opened 2017-03-16
Measure coverage of tests using codecov.io.

aalexand:


tamird:
I think you need to do something more sophisticated: https://github.com/codecov/example-go#caveat-multiple-files

.travis.yml
```diff
@@ -18,4 +18,7 @@ script:
   - staticcheck ./...
   - unused ./...
   - go test -v ./...
```
tamird: nit: this should probably be in `test.sh`, too. Note that `test.sh` correctly handles `vendor`, while this does not (and this would be obvious if this line were in that file).

aalexand:
@tamird LGTY?

tamird:
Yep ship it

PR #124, opened 2017-03-16
Improve weblist assembly output

aclements:


aalexand:
For the record, can you please include before and after screenshots / outputs for some example profile?

No tests?  Because no prior tests for this?

internal/report/source.go
```diff
@@ -236,11 +236,21 @@ func assemblyPerSourceLine(objSyms []*objSymbol, rs graph.Nodes, src string, obj
...
+	var prevline = 0
 	for _, an := range anodes {
 		if filepath.Base(an.file) == srcBase {
 			lineno = an.line
 		}
 		if lineno != 0 {
+			if lineno != prevline && len(assembly[lineno]) > 0 {
+				// Place a separator between
+				// non-contiguous instructions.
+				sep := assemblyInstruction{
```
rauls5382: Could we add a startBlock bool to assemblyInstruction instead of reusing the instruction field?
Or perhaps have this function return a map to a list of lists of assembly instructions.

rauls5382:
There is a high level test on internal/driver/driver_test.go that prints a weblist, but it doesn't seem to trigger this new function. Could we extend that testcase to cover this functionality?

aclements:
> For the record, can you please include before and after screenshots / outputs for some example profile?

I assume you mean for the alignment change, since I did include before/after outputs for the discontiguous blocks change. Sure, I can do that.

> There is a high level test on internal/driver/driver_test.go that prints a weblist, but it doesn't seem to trigger this new function. Could we extend that testcase to cover this functionality?

Yeah, I just found that. I've got a change to that test that exercises this code. Currently that test doesn't change because there's no file/line info on the instructions.

aalexand:
> I assume you mean for the alignment change, since I did include before/after outputs for the discontiguous blocks change. Sure, I can do that.

Ah, sorry, didn't look into commit description.  For the discontiguous blocks change - can it make the display somewhat confusing if the lines are not ordered with respect to assembly instructions?

aalexand:
> Ah, sorry, didn't look into commit description. For the discontiguous blocks change - can it make the display somewhat confusing if the lines are not ordered with respect to assembly instructions?

Ah nevermind, it's per line.

internal/report/source.go
aclements: Added a startBlock field to the first assemblyInstruction of each contiguous block.

aclements:
Ping.

aclements:
No worries. Thanks!

PR #125, opened 2017-03-16
internal/report: change the format of tags command output

hyangah:
This change also includes:
 - Instead of printing and sorting based on the first value in the
 internal value slice, use the sample type user selected.
 - Use the unit and print numbers in human friendly version.
 - Add a flag '-update' to internal/driver/driver_test to update the
 golden files.

 TODO: apply to other tests that use other golden files.


internal/report/report.go
```diff
@@ -620,24 +621,22 @@ func printTags(w io.Writer, rpt *Report) error {
...
-					valueMap[val] = valueMap[val] + s.Value[0]
-					continue
+				valueMap, ok := tagMap[key]
+				if !ok {
+					valueMap = make(map[string]int64)
 				}
-				valueMap := make(map[string]int64)
-				valueMap[val] = s.Value[0]
+				valueMap[val] += o.SampleValue(s.Value)
 				tagMap[key] = valueMap
```
tamird: throughout: maps are references, so I believe this can move inside the `if !ok {` scope

internal/report/report.go
```diff
@@ -656,18 +656,21 @@ func printTags(w io.Writer, rpt *Report) error {
...
+				fmt.Fprintf(tabw, " \t%.1f%s (%s):\t %s\n", f, u, percentage(t.FlatValue(), total), t.Name)
 			} else {
-				fmt.Fprintf(w, "  %8d: %s\n", t.FlatValue(), t.Name)
+				fmt.Fprintf(tabw, " \t%.1f%s:\t %s\n", f, u, t.Name)
 			}
 		}
-		fmt.Fprintln(w)
+		fmt.Fprintln(tabw)
 	}
+	tabw.Flush()
```
tamird: `return tabw.Flush()`

internal/report/report.go
hyangah: done

internal/report/report.go
hyangah: done

rauls5382:
I **love** that -update flag: where have you been all of my life? ;-)

About the tag formatting, this assumes that the unit of the tags is the same as the sample unit. This is problematic because different sample values may have different units. This changes makes us interpret tags using the units of whatever sample value we're printing. To fully solve this we probably should have a unit field for numeric tags (eg a map from tag name to a unit name).

Previously as a hack we have used the tag name as the unit if we can recognize it as such. That would also work in the short term, but it will not be sufficient as we start having samples with multiple numeric tags.

hyangah:
@rauls5382 Thanks for the review but I am afraid I don't understand what you mean by 'the unit of the tags'. 

I probably don't fully understand the goal of 'tags' command tries to report. Based on the original code, I read 'total' is the sum of sample.Value[0] with the same tag key, and the command outputs the sum of sample.Value[0] with a key:tag and its fraction to the 'total'. 

The confusion arose when analyzing the Go's memory profile. Go's memory profile stores values in the (alloc_count, alloc_space, inuse_count, inuse_space) order. https://go.googlesource.com/go/+/master/src/runtime/pprof/protomem.go#19
pprof in default mode uses inuse_space, but tags report numbers taken from alloc_count. These numbers look random and don't provide any meaningful info to users who are looking into inuse_space profile data.

What I am trying to propose here is the current use of sample.Value[0] is probably not right; instead we should use the Value the tool is currently analyzing. So, instead of sample.Value[0], I am using o.SampleValue(s.Value) and thus am using the sample unit.

If you prefer to simply provide the sample counts (per tags) instead, I think I can change to look up the associated counter Value (e.g. for inuse_space mode, we should look at inuse_count value (s.Value[2]) in go's memory profile) with heuristics. Then, there will be no 'unit'.

rauls5382:
No, I'm sorry, I misunderstood before what you were trying to do.

Previously the tags command was printing SampleValue[0] as a proxy for the number of samples, which is a convention we've used when generating some profiles, but I agree that using the selected sample value and its associated unit is a better approach.

I thought you were trying to deal with numeric tags, which currently do not have an associated unit (we're using the tag name as the unit -- see report.formatTag), but now I see that is unchanged.

Sorry for the noise.

Issue #136, opened 2017-04-27
Disambiguate memory units in favor of IEC names


aalexand:
On reports, pprof uses B/kB/MB/GB to represent 1/2^10/2^20/2^30, which as at least one user pointed out is ambiguous at best. We should probably remap these to B/KiB/MiB/GiB, which should be more accurate. The only compat concerns would be for people that parse pprof's output, and we offer no guarantees not to break them.

The other place where we use units is on the unit fields in profile.proto and the command line override, where we recognize kb/mb/gb to mean kiB, MiB, GiB, etc... This is also ambiguous at best, and problematic because changing it could cause us to stop parsing existing profiles. In reality I don't know if we have any profile sources that currently generate profiles with anything other than 'b', but if there are we would break them. A quick search over the accessible code shows only profiles generated using 'bytes', so this is probably worthwhile to do in general. It would only affect people that are used to specifying units by hand to pprof. For example, pprof -unit=mb would change from MiB to MB, and instead users would have to say pprof -unit=MiB (or maybe pprof -unit=mib).

rsc:
Please please _please_ don't do that. It's just needlessly pedantic. Just accept that on computers MB means 2^20 and so on.

ahh:
Not everyone does accept that, and when I read a pprof profile I have *no idea* which sort of person the author is.

Saying KiB makes it absolutely clear.

ahh:
@aalexand Also, should I file a separate issue for needing larger units (TiB and PiB) or is that worth dealing with in one fell swoop?

aalexand:
Please file a separate bug for the larger units.

aalexand:
I filed this request at ahh@'s ask to discuss it.

In practice there isn't a lot of ambiguity about MB / GB for memory units - assuming powers of 2 are used is a safe bet.  E.g. `top` shows RSS as like "123M", and that expectedly means MiB.

That said, the clarity becomes more important when a profile represents larger amounts of memory as at TB vs. TiB the difference becomes ~10%.

rauls5382:
How about if we leave the current defaults as-is (as broken as they can pedantically be considered), but we support users specifying IEC units on the command line (eg -unit=GiB or even perhaps even -unit=IEC)?

Would that please/appease/at least not annoy everyone?

rauls5382:
Closing this due to lack of interest. Reopen if you care about it.

PR #138, opened 2017-04-30
Add support for terabyte and petabyte memory units.

### LINKED ISSUE #137: Larger space units needed
Currently pprof tops out at GB (Well, GiB: https://github.com/google/pprof/issues/136 :P).   I have profiles containing quantities much larger than 2^30 bytes.  

Please add TiB and PiB.

aalexand:
This fixes #137 (but not #136, this will come separately once we decide
what we want to do if anything).

### DIFF
--- a/internal/measurement/measurement.go
+++ b/internal/measurement/measurement.go
@@ -170,12 +170,16 @@ func memoryLabel(value int64, fromUnit, toUnit string) (v float64, u string, ok 
 
 	switch fromUnit {
 	case "byte", "b":
-	case "kilobyte", "kb":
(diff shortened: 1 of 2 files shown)


tamird:
FYI this functionality exists in https://github.com/dustin/go-humanize, if you are able to take on an external dependency.

aalexand:
@tamird pprof is [used][1] by Go as a vendor dependency so not willing to take the dep.

[1]: https://github.com/golang/go/tree/master/src/cmd/vendor/github.com/google/pprof

rauls5382:
Thank you

PR #140, opened 2017-05-24
When pruning unsimplified names, don't trim "(anonymous namespace)".

### LINKED ISSUE #139: Drop frames with an anonymous namespace don't get dropped
If a profile has a drop frame matching against an anonymous namespace, the match never happens. I think the culprit is this line (https://github.com/google/pprof/blob/master/profile/prune.go#L39), which checks for brackets before applying the drop frame RE, and only applies the RE to the symbol before the brackets. I'm assuming this is for symbols that still have parameters.

Here's an example [1], where I create a profile with a drop frame RE "Foo::.*::Bar". If I comment that line out, things behave as expected [2].

Since this is already a hack, I'll just special-case "(anonymous namespace)", but there are probably other things with a legit "(" as a part of the function name (std::function objects?).

[1] $ pprof -traces tmp.pb 
...
-----------+-------------------------------------------------------
       15B   
             Foo::(anonymous namespace)::Test::Bar
             Foo::(anonymous namespace)::Test::Baz
             main
             __libc_start_main
             _start
-----------+-------------------------------------------------------

[2] $ ./pprof -traces tmp.pb 
...
-----------+-------------------------------------------------------
       15B   Foo::(anonymous namespace)::Test::Baz
             main
             __libc_start_main
             _start
-----------+-------------------------------------------------------

s-kanev:
Fixes #139.

### DIFF
--- a/profile/prune.go
+++ b/profile/prune.go
@@ -22,6 +22,38 @@ import (
 	"strings"
 )
 
+var (
+	reservedNames = []string{"(anonymous namespace)", "operator()"}
+	bracketRx     = rx.MustCompile("(" + strings.Replace(strings.Replace(
+		strings.Join(append(reservedNames, "("), "|"),
+		"(", "\\(", -1), ")", "\\)", -1) + ")")
+)
+
+// simplifyFunc does some primitive simplification of function names.
+func simplifyFunc(f string) string {
+	fmt.Println(rx)
+	// Account for leading '.' on the PPC ELF v1 ABI.
+	funcName := strings.TrimPrefix(f, ".")
+	// Account for unsimplified names -- try  to remove the argument list by trimming
+	// starting from the first '(', but skipping reserved names that have '('.
+	if indeces := bracketRx.FindAllStringSubmatchIndex(funcName, -1); indeces != nil {
+		for _, ind := range indeces {
+			foundReserved := false
+			for _, res := range reservedNames {
+				if funcName[ind[0]:ind[1]] == res {
+					foundReserved = true
+					break
+				}
+			}
+			if !foundReserved {
+				funcName = funcName[:ind[0]]
+				break
+			}
+		}
+	}
+	return funcName
+}
+
 // Prune removes all nodes beneath a node matching dropRx, and not
 // matching keepRx. If the root node of a Sample matches, the sample
 // will have an empty stack.
@@ -33,15 +65,7 @@ func (p *Profile) Prune(dropRx, keepRx *regexp.Regexp) {
 		var i int
 		for i = len(loc.Line) - 1; i >= 0; i-- {
 			if fn := loc.Line[i].Function; fn != nil && fn.Name != "" {
-				// Account for leading '.' on the PPC ELF v1 ABI.
-				funcName := strings.TrimPrefix(fn.Name, ".")
-				// Account for unsimplified names -- trim starting from the first '('.
-				if index := strings.Index(funcName, "("); index > 0 {
-					// ... unless the thing in brackets is an anonymous namespace marker
-					if !strings.HasPrefix(funcName[index:], "(anonymous namespace)") {
-						funcName = funcName[:index]
-					}
-				}
+				funcName := simplifyFunc(fn.Name)
 				if dropRx.MatchString(funcName) {
 					if keepRx == nil || !keepRx.MatchString(funcName) {
 						break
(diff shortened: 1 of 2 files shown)


profile/prune.go
```diff
@@ -37,6 +37,10 @@ func (p *Profile) Prune(dropRx, keepRx *regexp.Regexp) {
 				funcName := strings.TrimPrefix(fn.Name, ".")
 				// Account for unsimplified names -- trim starting from the first '('.
 				if index := strings.Index(funcName, "("); index > 0 {
+					// ... unless the thing in brackets is an anonymous namespace marker
+					if !strings.HasPrefix(funcName[index:], "(anonymous namespace)") {
+						funcName = funcName[:index]
```
aalexand: Is it intentional that "funcName = funcName[:index] will execute twice (here and two lines below) if there is no prefix match?

profile/prune.go
```diff
@@ -37,6 +37,10 @@ func (p *Profile) Prune(dropRx, keepRx *regexp.Regexp) {
 				funcName := strings.TrimPrefix(fn.Name, ".")
 				// Account for unsimplified names -- trim starting from the first '('.
 				if index := strings.Index(funcName, "("); index > 0 {
+					// ... unless the thing in brackets is an anonymous namespace marker
```
aalexand: Could you add some tests for the change?

profile/prune.go
s-kanev: Not at all, just sloppy. Thanks for catching.

profile/prune.go
s-kanev: Done.

profile/prune.go
```diff
@@ -130,7 +133,10 @@ func (p *Profile) PruneFrom(dropRx *regexp.Regexp) {
 				funcName := strings.TrimPrefix(fn.Name, ".")
 				// Account for unsimplified names -- trim starting from the first '('.
 				if index := strings.Index(funcName, "("); index > 0 {
-					funcName = funcName[:index]
+					// ... unless the thing in brackets is an anonymous namespace marker
+					if !strings.HasPrefix(funcName[index:], "(anonymous namespace)") {
```
aalexand: It feels that we will unexpectedly match a function like `Hello::(anonymous namespace)::World(const Foo::(anonymous namespace)::Test::Bar` with the `Foo::.*::Bar` pattern you're using. Can you please add such double-parens-pair test case to check whether it's so? If it is, I think it can be fixed by using LastIndex instead of Index above.

aalexand: LastIndex won't help it seems, it's more convoluted than that :)

s-kanev: This should be a little more general -- just search for the first '(' that's not part of one of these reserved cases (operator() is another one). It will still fail for '(' inside a template, but I'm not sure there's an easy way around that other than actually parsing the AST.

profile/prune.go
```diff
@@ -22,6 +22,37 @@ import (
 	"strings"
 )
 
+var (
+	reservedNames = []string{"(anonymous namespace)", "operator()"}
+	bracketRx     = regexp.MustCompile("(" + strings.Replace(strings.Replace(
```
tamird: is this `strings.Replace` dance really just [`regexp.QuoteMeta`](https://godoc.org/regexp#QuoteMeta)?

s-kanev: Unfortunately, almost. QuoteMeta would also escape the "|" between reserved words, which we don't want here.

tamird: You might write:
```golang
	bracketRx     = func() *regexp.Regexp {
		var quotedNames []string
		for _, name := range append(reservedNames, "(") {
			quotedNames = append(quotedNames, regexp.QuoteMeta(name))
		}
		return regexp.MustCompile(strings.Join(quotedNames, "|"))
	}()
```

The opening and closing parentheses you wrap the regexp in are also unnecessary, I believe (so I've omitted them above).

s-kanev: Indeed, this is more readable, thanks.

profile/prune.go
```diff
@@ -22,6 +22,41 @@ import (
...
+	}()
+)
+
+// simplifyFunc does some primitive simplification of function names.
+func simplifyFunc(f string) string {
+	// Account for leading '.' on the PPC ELF v1 ABI.
+	funcName := strings.TrimPrefix(f, ".")
+	// Account for unsimplified names -- try  to remove the argument list by trimming
+	// starting from the first '(', but skipping reserved names that have '('.
+	if indeces := bracketRx.FindAllStringSubmatchIndex(funcName, -1); indeces != nil {
```
aalexand: indeces -> indices

s-kanev: Done.

profile/prune.go
```diff
@@ -22,6 +22,41 @@ import (
...
+	}()
+)
+
+// simplifyFunc does some primitive simplification of function names.
+func simplifyFunc(f string) string {
+	// Account for leading '.' on the PPC ELF v1 ABI.
+	funcName := strings.TrimPrefix(f, ".")
+	// Account for unsimplified names -- try  to remove the argument list by trimming
+	// starting from the first '(', but skipping reserved names that have '('.
+	if indices := bracketRx.FindAllStringSubmatchIndex(funcName, -1); indices != nil {
```
aalexand: By the way, nil slice == empty slice in Go so you could avoid the check for nil here saving one indentation level.

s-kanev: Thanks for the pointer.

profile/prune.go
```diff
@@ -22,6 +22,40 @@ import (
...
+	}()
+)
+
+// simplifyFunc does some primitive simplification of function names.
+func simplifyFunc(f string) string {
+	// Account for leading '.' on the PPC ELF v1 ABI.
+	funcName := strings.TrimPrefix(f, ".")
+	// Account for unsimplified names -- try  to remove the argument list by trimming
+	// starting from the first '(', but skipping reserved names that have '('.
+	indices := bracketRx.FindAllStringSubmatchIndex(funcName, -1)
```
tamird: Somewhat golfing, but this local is also not needed; you can just `for _, ind := range bracketRx.FindAll....`

s-kanev: Fair point.

PR #150, opened 2017-07-06
Added files to enable fuzz testing

nolanmar511:
Files in the new directory, fuzz, are allow one to conduct fuzz testing using github.com/dvyukov/go-fuzz


aalexand:
Looks like appveyor is enabled but not configured yet. Could you disable it for now so that we don't get the failures on PRs?

aalexand:
Regarding the fuzz testing, we need to support two flows:

1) Running "go test ./..." in the top-level pprof source directory should run against the known representative corpus to detect any regressions.
2) Some manual steps (document them in `fuzz/README.md` perhaps) can be followed to do more fuzzing and to augment the corpus.

It's not clear whether the current layout addresses the two flows?

nolanmar511:
For the fuzz testing, the first flow is not addressed -- I had intended to add some testing when adding the bug fix (so that I didn't add a fail test) but I can add testing that will all pass now. How would you recommend testing which files in the corpus to test against? There are currently 2060 things (~445 kb) in the corpus, but there isn't really a way to check which ones are representative. Should I include them all or maybe do a random sample?

Adding a README to explain how to do the fuzz testing is an excellent idea -- I'll work on that now.

aalexand:
Submitting all of the corpus is probably too much. Should we include the crashers at least?

nolanmar511:
I've added the one crasher and a test for that crasher. 
So the test didn't fail, I also added in a fix for this error.

PR #151, opened 2017-07-07
Update legend to show show/hide/focus/ignore settings

nolanmar511:
Updated legend to show settings related to the options show, hide, focus, ignore, tagfocus, tagignore, tagshow, and taghide

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -229,7 +229,7 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error) 
 	for _, k := range []string{"focus", "ignore", "hide", "show", "tagfocus", "tagignore", "tagshow", "taghide"} {
 		v := vars[k].value
 		if v != "" {
-			filters = append(filters, "   " + k+"="+v)
+			filters = append(filters, "   "+k+"="+v)
 		}
 	}
 
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -238,6 +238,13 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error) 
 		NodeFraction: vars["nodefraction"].floatValue(),
 		EdgeFraction: vars["edgefraction"].floatValue(),
 
+		Focus:				vars["focus"].value,
+		Ignore:				vars["ignore"].value,
+		Hide:					vars["hide"].value,
+		Show:					vars["show"].value,
+		TagFocus:			vars["tagfocus"].value,
+		TagIgnore:		vars["tagignore"].value,
+
 		SampleValue:       value,
 		SampleMeanDivisor: meanDiv,
 		SampleType:        stype,
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -229,7 +229,7 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error) 
 	for _, k := range []string{"focus", "ignore", "hide", "show", "tagfocus", "tagignore", "tagshow", "taghide"} {
 		v := vars[k].value
 		if v != "" {
-			filters = append(filters, k+"="+v)
+			filters = append(filters, "   " + k+"="+v)
 		}
 	}
 
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -238,12 +238,12 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error) 
 		NodeFraction: vars["nodefraction"].floatValue(),
 		EdgeFraction: vars["edgefraction"].floatValue(),
 
-		Focus:				vars["focus"].value,
-		Ignore:				vars["ignore"].value,
-		Hide:					vars["hide"].value,
-		Show:					vars["show"].value,
-		TagFocus:			vars["tagfocus"].value,
-		TagIgnore:		vars["tagignore"].value,
+		Focus:     vars["focus"].value,
+		Ignore:    vars["ignore"].value,
+		Hide:      vars["hide"].value,
+		Show:      vars["show"].value,
+		TagFocus:  vars["tagfocus"].value,
+		TagIgnore: vars["tagignore"].value,
 
 		SampleValue:       value,
 		SampleMeanDivisor: meanDiv,
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -225,6 +225,14 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error) 
 		return nil, fmt.Errorf("zero divisor specified")
 	}
 
+	var filters []string
+	for _,k := range []string{"focus", "ignore", "hide", "show", "tagfocus", "tagignore", "tagshow", "taghide"} {
+		v := vars[k].value
+		if v != "" {
+			filters = append(filters, k+"="+v)
+		}
+	}
+
 	ropt := &report.Options{
 		CumSort:             vars["cum"].boolValue(),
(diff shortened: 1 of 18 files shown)


internal/report/report.go
```diff
@@ -1089,6 +1130,30 @@ func reportLabels(rpt *Report, g *graph.Graph, origCount, droppedNodes, droppedE
 	return label
 }
 
+func parseTagFocusIgnore(tag string) string {
```
aalexand: I think trying to parse the tag filter expression makes things more complicated than needed. Also, the tag ranges can be used for tags other than allocations related so the "with allocations" part may easily get wrong.

internal/report/report.go
```diff
@@ -1070,6 +1078,39 @@ func reportLabels(rpt *Report, g *graph.Graph, origCount, droppedNodes, droppedE
 		flatSum = flatSum + n.FlatValue()
 	}
 
+	if rpt.options.Focus != "" {
+		label = append(label, fmt.Sprintf("Using only samples with locations matching %s",
+			rpt.options.Focus))
+	}
+	if rpt.options.Ignore != "" {
+		label = append(label, fmt.Sprintf("Removed profile samples with locations matching %s",
```
aalexand: The filtering verb here is "ignore", so use that in the text? You do for tagignore below. Please make it consistent. In other places too.

internal/driver/testdata/pprof.cpu.cum.lines.text.show
```diff
@@ -1,3 +1,4 @@
+Using only nodes matching [12]00
```
aalexand: To make the output and the legend simpler and more compact, would it be better to just do something like:

Active filters: focus=W, ignore=X, show=Y, hide=Z

?

aalexand:
Please also make sure the PR does not make the test coverage worse. It does now.

internal/driver/testdata/pprof.cpu.cum.lines.text.focus.hide
```diff
@@ -0,0 +1,6 @@
+Active Filters: focus=[12]00, hide=line[X3]0
```
aalexand: "Active filters"? Without capitalization, because "Showing nodes" is that way.

internal/report/report.go
```diff
@@ -1078,37 +1078,35 @@ func reportLabels(rpt *Report, g *graph.Graph, origCount, droppedNodes, droppedE
 		flatSum = flatSum + n.FlatValue()
 	}
 
+	var filters []string
 	if rpt.options.Focus != "" {
-		label = append(label, fmt.Sprintf("Using only samples with locations matching %s",
-			rpt.options.Focus))
+		filters = append(filters, fmt.Sprintf("focus=%s", rpt.options.Focus))
```
aalexand: Just `filters = append(filters, "focus=" + rpt.options.Focus)` perhaps?

internal/driver/driver_test.go
```diff
@@ -164,6 +167,7 @@ func TestParse(t *testing.T) {
 			t.Errorf("Failed to read profile %s: %v", outputTempFile.Name(), err)
 		}
 
+		fmt.Println(solution)
```
aalexand: Is this a debugging leftover?

internal/report/report.go
```diff
@@ -64,6 +64,14 @@ type Options struct {
 	Ratio               float64
 	Title               string
 	ProfileLabels       []string
+	Focus               string
```
aalexand: It might be simpler to have instead something like `ActiveFilters []string`, put there `key=value` entries like `focus=foo` and print them out as is. The code would be shorter and it would also be clearer that these are passed in just for display purposes, not to be used for the actual report construction.

internal/driver/testdata/pprof.cpu.cum.lines.text.focus.hide
nolanmar511: Done!

internal/report/report.go
nolanmar511: Good idea! Thanks! (and done)

internal/driver/driver_test.go
nolanmar511: Yes -- thanks!

internal/driver/testdata/pprof.cpu.cum.lines.text.show
nolanmar511: Done!

internal/driver/driver.go
```diff
@@ -225,6 +225,14 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
 		return nil, fmt.Errorf("zero divisor specified")
 	}
 
+	var filters []string
+	for _,k := range []string{"focus", "ignore", "hide", "show", "tagfocus", "tagignore", "tagshow", "taghide"} {
```
aalexand: "_,k" is missing a space after ",". Run "go fmt", don't add the space manually.

aalexand:
The only question I have about the PR is whether specifying practically long filters could make the legend too wide and not looking good.

Can you try grabbing a profile from one of the services and verify the legend looks OK with some combinations of focus / ignore / show / hide filters? You may patch in https://github.com/google/pprof/pull/154 and run the interactive web interface which will give you an easy way to construct such filters by clicking on a node in the browser and clicking respective filter buttons.

nolanmar511:
The legend was really wide when filters were long, so I did need to change the formatting. 
It now looks like this when the expressions are long:
image

internal/report/report.go
```diff
@@ -1070,6 +1071,16 @@ func reportLabels(rpt *Report, g *graph.Graph, origCount, droppedNodes, droppedE
 		flatSum = flatSum + n.FlatValue()
 	}
 
+	if len(rpt.options.ActiveFilters) > 0 {
```
aalexand: Maybe split this fragment into a dedicated function and add a test for it.

internal/driver/driver.go
```diff
@@ -225,6 +225,14 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
 		return nil, fmt.Errorf("zero divisor specified")
 	}
 
+	var filters []string
+	for _, k := range []string{"focus", "ignore", "hide", "show", "tagfocus", "tagignore", "tagshow", "taghide"} {
+		v := vars[k].value
+		if v != "" {
+			filters = append(filters, "   "+k+"="+v)
```
aalexand: I think adding the leading spaces should be pushed down to where the legend formatting is done. The caller here doesn't know that the entries are going to be displayed one per line, they could as well be semicolon separated or something, where the leading spaces are unwanted.

internal/report/report.go
```diff
@@ -1070,6 +1071,16 @@ func reportLabels(rpt *Report, g *graph.Graph, origCount, droppedNodes, droppedE
...
+		label = append(label, "Active filters:")
+		for i, s := range rpt.options.ActiveFilters {
+			if len(s) > 80 {
+				rpt.options.ActiveFilters[i] = s[:70] + "...(regex continues)"
```
aalexand: Maybe use the unicode ellipsis character (…) and drop the "(regex continues)" part. Using the ellipsis should make things clear enough that these dots are not part of the regex but rather represent the omission.

internal/report/report.go
nolanmar511: Done!

internal/report/report.go
nolanmar511: Done

internal/driver/driver.go
nolanmar511: Done

PR #154, opened 2017-07-10
Added an interactive web interface

ghemawat:
The web interface is triggered by passing -http=port on the command line.
The interface is available by visiting localhost:port in a browser.

Requirements:
* Graphviz must be installed.
* Browser must support Javascript.
* Tested in recent stable versions of chrome and firefox.

Features:
* The entry point is a dot graph display (equivalent to "web" output).
* Nodes in the graph can be selected by clicking.
* A regular expression can also be typed in for selection.
* The current selection (either list of nodes or a regexp)
  can be focused, ignored, shown, or hidden.
* Source code or disassembly of the current selection can be displayed.


internal/driver/cli.go
```diff
@@ -58,6 +59,7 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flagTools := flag.String("tools", os.Getenv("PPROF_TOOLS"), "Path for object tool pathnames")
 
 	flagTimeout := flag.Int("timeout", -1, "Timeout in seconds for fetching a profile")
+	flagHTTPPort := flag.Int("http", 0, "Present interactive web based UI at the specified http port")
```
aalexand: Would it be possible to make the "http" another command here - https://github.com/google/pprof/blob/1b8137e88907127c25b226f856452f42d08ec336/internal/driver/commands.go#L87?

I felt this would integrate more naturally, but maybe I am missing something.

ghemawat: That's what I had tried initially, but it became awkward for a few reasons:

1. The usage message for the new command showed up under `Output formats` which seemed odd.

2. Commands only support string parameters, whereas I wanted a port number.

3. Making it a command would allow `http _port_` as a command in the interactive shell, which I wasn't sure we wanted to support.

So possible, but not sure it would be better/simpler than the current approach.

README.md
```diff
@@ -77,6 +77,18 @@ This will open a simple shell that takes pprof commands to generate reports.
 Type 'help' for available commands/options.
 ```
 
+## Run pprof via a web interface
+
+If the `-http=port` option is specified, pprof starts a web server at
+the specified port that you can visit in your browser to get an interactive
```
aalexand: Maybe mention here that the browser will auto-open.

README.md
```diff
@@ -77,6 +77,18 @@ This will open a simple shell that takes pprof commands to generate reports.
...
+interface to pprof.
+
+```
+pprof -http=[port] [main_binary] profile.pb.gz
+
+This will run a server that can be visited in the browser.
```
aalexand: Nit: this line seems redundant given the above text.

internal/driver/cli.go
aalexand: I see, makes sense to me.

internal/driver/webui.go
```diff
@@ -0,0 +1,234 @@
...
+		host, _, err := net.SplitHostPort(req.RemoteAddr)
+		if err != nil || ((host != "127.0.0.1") && (host != "::1")) {
+			http.Error(w, "permission denied", http.StatusForbidden)
+			return
+		}
+		h.ServeHTTP(w, req)
+	})
+}
+
+// disasm generates a web page containing an svg diagram..
```
aalexand: "disasm" -> "dot"
also fix double full stop at the end.

internal/driver/webui.go
```diff
@@ -0,0 +1,234 @@
...
+	g, config := report.GetDOT(rpt)
+	legend := config.Labels
+	config.Labels = nil
+	dot := &bytes.Buffer{}
+	graph.ComposeDot(dot, g, &graph.DotAttributes{}, config)
+
+	// Convert to svg.
+	svg, err := dotToSvg(dot.Bytes())
+	if err != nil {
+		reportHandlerError(w, "Failed to execute dot. Is Graphviz installed?", err)
```
aalexand: Maybe change to "failed to execute dot: is Graphviz installed?" to start with lowercase like in other messages.

internal/driver/webui.go
```diff
@@ -0,0 +1,234 @@
...
+		reportHandlerError(w, "error generating report", err)
+		return
+	}
+	w.Header().Set("Content-Type", ctype)
+	io.Copy(w, out)
+}
+
+// getFromLegend returns the suffix of an entry in legend that starts
+// with param.  It returns def if no such entry is found.
+func getFromLegend(legend []string, param, def string) string {
```
aalexand: Extracting the data from the legend seems a bit fragile. Maybe make pprof fail early in tests if the parameter to extract is not found? Or something else that the tests would detect.

internal/driver/cli.go
rauls5382: I'm OK either way, but I think having http be listed under Output Formats wouldn't be terrible: we have (web|evince|gv) in there already. We could always explicitly disable the command in the UI if we really wanted to (is there really a problem with it?).

In fact, how about calling this "web" instead of "http", and eliminating the current "web"?
There's always "svg" for people that really want a static svg.

ghemawat: 1. I think reusing the name "web" will be problematic since it will change the behavior substantially without warning.  E.g., consider somebody who has a script that does "pprof web" a few times to open a bunch of views in a row.  The interactive UI will block on the first call.  I also am not so excited about "There's always "svg" for people that really want a static svg."  If we do that, we are going to force some people to learn a new behavior after they get confused a bit about why things don't work as they used to; and I don't think the win here is big enough to justify that.

2. W.r.t. putting "http" under "Output Formats", I still think that it is better to keep things as they are for the reasons mentioned above (usage message, port number not a string, interactive command misuse).

So unless I hear a stronger complaint than "I'm OK either way", I will keep "http" out of Output Formats.

README.md
ghemawat: Done.

README.md
ghemawat: Done.

internal/driver/webui.go
ghemawat: Done.

internal/driver/webui.go
ghemawat: Done.

internal/driver/webui.go
ghemawat: I augmented webui_test.go to check that the binary name and profile type are in the generated output.

I didn't do more than that because I was thinking that it would be best to eventually change lower-level code to produce legend info in a struct and leave the stringification to a helper routine.  In this code we would avoid the stringification altogether and just get the pieces we need directly from the struct.  E.g., I think the header we generate here could be augmented to include info about the displayed counts vs. the overall count.

internal/driver/webui.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/google/pprof/internal/graph"
+	"github.com/google/pprof/internal/plugin"
+	"github.com/google/pprof/internal/report"
+	"github.com/google/pprof/profile"
+)
+
+// webUI holds the state needed for serving a browser based interface.
+type webUI struct {
+	prof    *profile.Profile
+	options *plugin.Options
```
rauls5382: The motivation behind the UI field in plugin.Options is that it would allow errors to be redirected to an interactive UI instead of always printing to stderr.

Would it be possible to redefine options.UI so that it would show such errors somewhere in the UI (either a pop-up or maybe in red in a corner)?

For instance, if you type a function name that doesn't exist and click "focus" we get a print on stderr saying "Focus expression matched no samples". It would be better if that showed up on the browser.

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+    } else {
+      window.location.href = url.toString()
+    }
+  }
+
+  var updateButtons = function() {
+    var enable = (currentRe != "" || selected.size != 0)
+    if (buttonsEnabled == enable) return
+    buttonsEnabled = enable
+    var d = enable ? false : true
```
josef-jelinek: Why not equivalent "!enable" instead

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+  var resetButton = document.getElementById("reset")
+  var focusButton = document.getElementById("focus")
+  var showButton = document.getElementById("show")
+  var ignoreButton = document.getElementById("ignore")
+  var hideButton = document.getElementById("hide")
+  var search = document.getElementById("searchbox")
+  var graph0 = document.getElementById("graph0")
+  var svg = graph0.parentElement
+
+  var currentRe = ""
```
josef-jelinek: FYI, Map is an ES6 feature. Relying on it means that the entire code can utilize ES6 (const, let, lambdas, for-of, destructuring, etc.). The same goes for the const at the top, also ES6 only.

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+  // State needed to implement zooming.
+  var currentScale = 1.0
+  var initWidth = svg.viewBox.baseVal.width
+  var initHeight = svg.viewBox.baseVal.height
+
+  // State needed to implement panning.
+  var panLastX = 0      // Last event X coordinate
+  var panLastY = 0      // Last event Y coordinate
+  var moved = false     // Have we seen significant movement
+  var touchid = null;   // Current touch identifier
```
josef-jelinek: Although the code tries to avoid semicolons, there are still some left, which are not necessary (the only necessary semicolon is when the next line starts with "(" or "[" AFAIR).

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+
+  var handleSearch = function() {
+    // Delay processing so a flurry of key strokes is handled once.
+    if (searchAlarm != null) {
+      clearTimeout(searchAlarm)
+    }
+    searchAlarm = setTimeout(doSearch, 300)
+  }
+
+  doSearch = function() {
```
josef-jelinek: missing "var" (I recommend using "use strict" directive in global functions, which catch this). The same e.g. for the select and unselect functions below.

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+</div>
+
+</div>
+</body>
+</html>
+<script>
+
+// Make svg pannable and zoomable.
+// Call clickHandler(t) if a click event is caught by the pan event handlers.
+function initPanAndZoom(svg, clickHandler) {
```
josef-jelinek: function statements are used here for global functions, but assignment of function expressions are used for the inner functions, is there any reason?
Unless reassigning a variable with another function, the statements have the advantage of being order independent and can be called before they are defined in the source, so the code can be top-down. The assignments have to be in a correct order and defined before used (unless used in another function and called later), which enforced bottom-up order).

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+    }
+    return null
+  }
+
+ // Return distance between two touch points
+  var touchGap = function(t1, t2) {
+    var dx = t1.clientX - t2.clientX
+    var dy = t1.clientY - t2.clientY
+    return Math.sqrt(Math.pow(dx, 2) + Math.pow(dy, 2))
+  }
```
josef-jelinek: There is Math.hypot() (ES6+)

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+// limitations under the License.
+
+package driver
+
+import "html/template"
+
+var graphTemplate = template.Must(template.New("graph").Parse(
+	`<!DOCTYPE html>
+<html>
+<head>
```
josef-jelinek: nit: trailing "/" in tags is a HTML error (that is being gracefully ignored, it was used only for now dead XHTML). Only pair tags have to be paired.

internal/driver/webhtml.go
```diff
@@ -0,0 +1,555 @@
...
+</div>
+
+<div id="graph">
+{{.Svg}}
+</div>
+
+</div>
+</body>
+</html>
+<script>
```
josef-jelinek: script should be before closing body (browsers generally ignore closing body and html and insert then at the end anyway, but AFAIK IE 10 for example did not and could ignore the script.

internal/driver/webhtml.go
ghemawat: Switched to Math.hypot.

internal/driver/webhtml.go
ghemawat: Switched to const where possible, let elsewhere.
Switched to "for..of" where possible.
I didn't spot any place where lambdas or destructuring assingment would make things easier to read.

internal/driver/webhtml.go
ghemawat: Added 'use strict'; very helpful.  I put it just inside my two global functions instead of the top-level, since I presume you had some reason in mind when you mentioned global functions,

internal/driver/webhtml.go
ghemawat: Remnant of older code.  Fixed to use !enable instead.

PR #160, opened 2017-07-14
Modified to improve style of fuzz.go

nolanmar511:


fuzz/main.go
```diff
@@ -1,3 +1,5 @@
+// Package is used in conjunction with
```
aalexand: This is incomplete. Also, a Go package should only be documented once. It's probably the best to document it here, not in fuzz_test.go.

nolanmar511: Thanks -- I've fixed this

PR #161, opened 2017-07-16
Added peek command to web interface.

ghemawat:
Detailed list of changes:
* Since the list of actions is now long, move it into
  an action box that is displayed only when applicable.
* Display help messages when hovering over action buttons.
* Hitting return in the search box invokes "focus" action.
* Cleaned up stale css entries.

I have attached one of the resulting pages to see what things look like easily.  It is static, so won't allow navigation to other pages, but should be interactive enough.

[benchcpu.zip](https://github.com/google/pprof/files/1151238/benchcpu.zip)

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -113,13 +113,13 @@ button {
 <div id="graph">
 
 <div id="actionbox">
-<button title="{{.Help.list}}" id="list">List</button>
-<button title="{{.Help.disasm}}" id="disasm">Disassemble</button>
-<button title="{{.Help.peek}}" id="peek">Peek</button>
 <button title="{{.Help.focus}}" id="focus">Focus</button>
 <button title="{{.Help.ignore}}" id="ignore">Ignore</button>
 <button title="{{.Help.hide}}" id="hide">Hide</button>
 <button title="{{.Help.show}}" id="show">Show</button>
+<button title="{{.Help.peek}}" id="peek">Peek</button>
+<button title="{{.Help.list}}" id="list">List</button>
+<button title="{{.Help.disasm}}" id="disasm">Disassemble</button>
 </div>
 
 {{.Svg}}
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -32,13 +32,6 @@ body {
   width: 100%;
   overflow: hidden;
 }
-h1 {
-  font-weight: normal;
-  font-size: 24px;
-  padding: 0em;
-  margin-top: 5px;
-  margin-bottom: 5px;
-}
 #page {
   display: flex;
   flex-direction: column;
@@ -78,16 +71,32 @@ button {
   background-color: #ffffff;
   border: 1px solid black;
   box-shadow: 2px 2px 2px 0px #aaa;
-  padding: 1em;
   top: 20px;
   right: 20px;
+  z-index: 1;
+}
+.actionhdr {
+  background-color: #ddd;
+  width: 100%;
+  border-bottom: 1px solid black;
+  border-top: 1px solid black;
+  font-size: 16pt;
 }
 #actionbox > button {
   display: block;
   width: 100%;
+  margin: 0px;
+  text-align: left;
+  padding-left: 0.5em;
+  background-color: #fff;
+  border: none;
+  font-size: 14pt;
+}
+#actionbox > button:hover {
+  background-color: #ddd;
 }
 #home {
-  font-size: 24px;
+  font-size: 20pt;
   padding-left: 0.5em;
   padding-right: 0.5em;
 }
@@ -113,10 +122,12 @@ button {
 <div id="graph">
 
 <div id="actionbox">
+<div class="actionhdr">Refine graph</div>
 <button title="{{.Help.focus}}" id="focus">Focus</button>
 <button title="{{.Help.ignore}}" id="ignore">Ignore</button>
 <button title="{{.Help.hide}}" id="hide">Hide</button>
 <button title="{{.Help.show}}" id="show">Show</button>
+<div class="actionhdr">Show Functions</div>
 <button title="{{.Help.peek}}" id="peek">Peek</button>
 <button title="{{.Help.list}}" id="list">List</button>
 <button title="{{.Help.disasm}}" id="disasm">Disassemble</button>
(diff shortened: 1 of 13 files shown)


internal/driver/webui.go
```diff
@@ -215,6 +226,11 @@ func (ui *webInterface) weblist(w http.ResponseWriter, req *http.Request) {
 	ui.output(w, req, "weblist", "text/html")
 }
 
+// peek generates a web page listing callers/callers.
+func (ui *webInterface) peek(w http.ResponseWriter, req *http.Request) {
```
aalexand: Add a test to cover this new code?

internal/driver/webhtml.go
```diff
@@ -91,34 +72,56 @@ button {
...
+
+<div id="errors">{{range .Errors}}<div>{{.}}</div>{{end}}</div>
 
 <div id="graph">
+
+<div id="actionbox">
+<button title="{{.Help.list}}" id="list">List</button>
+<button title="{{.Help.disasm}}" id="disasm">Disassemble</button>
+<button title="{{.Help.peek}}" id="peek">Peek</button>
+<button title="{{.Help.focus}}" id="focus">Focus</button>
```
aalexand: Since focus is the default action, maybe it should go first.

internal/driver/webhtml.go
```diff
@@ -91,34 +72,56 @@ button {
...
+
+<input id="searchbox" type="text" placeholder="Search regexp" autocomplete="off" autocapitalize="none" size=40>
+
+<div id="page">
+
+<div id="errors">{{range .Errors}}<div>{{.}}</div>{{end}}</div>
 
 <div id="graph">
+
+<div id="actionbox">
```
aalexand: I find the UI where the action menu suddenly shows on the very right after you start typing into the search box a bit weird. It would be better if the menu would show closer to the search box. Or (even better IMO) if there would be some sort of a drop-down-looking button (like the Details one) with the actions (maybe disabled until there is something typed into the box).

This is not a deal breaker for the PR, just mentioning something that we could probably do better.

ghemawat: Two reasons I can think of to keep things as they are:
1. Positioning and transitioning become more complicated to handle if we try to make things look like a menu.
2. The current positioning on the right hand side avoids obscuring the middle of the svg display.

Let's see if others have opionions about this.

internal/driver/webhtml.go
ghemawat: I think we should keep the following groups together {list,disasm,peek}, {focus,ignore,hide,show}.  (Everything in the former group generates a new report; everything in the latter group refines the current graph.)

So if we move focus to the top, I would suggest the resulting order should be:

focus
ignore
hide
show
list
disasm
peek

Does that seem reasonable?

internal/driver/webui.go
ghemawat: Will do.  Waiting to get some more feedback on the UI questions.

internal/driver/webhtml.go
aalexand: Yes. As a minor detail, I would move peek to come after show since it too operates on function graph like {focus,ignore,hide,show}, while list and disasm produce source and assembly views.

internal/driver/webhtml.go
rauls5382: Perhaps we could leave the box on all the time, producing an error if they are clicked if no node is selected and no text is chosen?

rauls5382:
UI looks pretty good.

I personally often find peek most useful when looking at line number granularity.
If adding granularity options is too much, perhaps consider having peek turn on line-number granularity?

internal/driver/webhtml.go
ghemawat: A reason to keep the box hidden until its actions are enabled is that it reduces clutter; users don't see something they can't interact with.

Though I don't speak from much experience with UI design. How about we submit something and if we can find an interface expert, ask them to improve things or at least suggest changes.

internal/driver/webhtml.go
ghemawat: Done.  I also added titles to the two different sections in the command list.

internal/driver/webui.go
ghemawat: Done.

aalexand:
There is a merge conflict, please resolve.

ghemawat:
Not sure where the following message is coming from "The bad news is that it appears that one or more commits were authored by someone other than the pull request submitter. We need to confirm that they're okay with their commits being contributed to this project. Please have them confirm that here in the pull request."

Is it because I merged upstream changes into my branch?

aalexand:
It's likely because of the merge. You should have rebased the changes onto the master instead I think. Something like "git checkout master", "git pull --ff-only master", "git checkout peek", "git rebase -i master", <resolve conflicts>, git rebase --continue (as needed), git push --force ghemawat.

aalexand:
A possible way to get the right commits into this PR now is I think something like

```
git checkout master
git pull --ff-only
git checkout peek
git pull --ff-only
git reset --hard b82dbbd972240fb7c3042950f8574914c04f41ca
git rebase -i master
```

In the editor that shows up leave these commits removing 6 bottom ones:

```
pick 8ea3b7d Added peek command to web interface.
pick 59f70e8 added test for /peek urls
pick 577a2c5 rearranged actions so Focus comes first
pick ee43d4a Added action titles and change styling
pick ee8820e Use smaller fonts for action menu
pick b82dbbd line granularity for peek command
```

Then `git push --force ghemawat`. This should result in something like https://github.com/google/pprof/compare/master...aalexand:peek?expand=1, which appears to be right.

I can also just try pressing the green "Squash and merge" button, it appears to be enabled, so maybe things will figure themselves out, but I am not sure.

aalexand:
(the last command may need to be `git push --force origin`, depending on what's the right local name of the remote.

ghemawat:
On Wed, Jul 19, 2017 at 10:07 PM, Alexey Alexandrov <
notifications@github.com> wrote:

> (the last command may need to be git push --force origin, depending on
> what's the right local name of the remote.
>

So many ways to move commits from one place to another...

I think I had gotten everything right except for the --force on the last
command.

So what's the deal with this PR?  Can you override and merge it anyway, or
should I close and reopen?

aalexand:
I merged it, it seems to be OK. Maybe take a look at the resulting [commit](https://github.com/google/pprof/commit/7209e76b006cc26137cb00fd4184fd7d7f937c7f) to make sure it got everything, just in case.

PR #162, opened 2017-07-17
Added option to normalize by base profile

nolanmar511:
When calling a command like "pprof -normalize -base this_profile_has_5_seconds_of_cpu_time_and _200_samples_total.pb.gz this_profile_has_8_seconds_of_cpu_time_and_500_samples_total.pb.gz" all time metrics in the non-base profile would be multiplied by 5/8, and  all sample metrics would be multiplies by 200/500  before subtracting the base profile from the source profile.

This makes it easier to compare profiles with a different number of samples or different total duration.


internal/driver/fetch.go
```diff
@@ -41,36 +41,119 @@ import (
...
-	}
-	if want, got := len(sources), cnt; want != got {
-		o.UI.PrintErr(fmt.Sprintf("fetched %d profiles out of %d", got, want))
+	normalize := s.Normalize
+
+	var p *profile.Profile
+	var save bool
+	var msrcs plugin.MappingSources
+
+	if normalize {
```
aalexand: Can you limit this condition to only whether the profile is normalized or not? There is no reason for the fetch of the profiles to be different that much, is there?

nolanmar511: I refactored a bit -- which hopefully helps this.

internal/driver/fetch.go
```diff
@@ -41,36 +41,82 @@ import (
...
-	if err != nil {
-		return nil, err
-	}
-	if cnt == 0 {
-		return nil, fmt.Errorf("failed to fetch any profiles")
-	}
-	if want, got := len(sources), cnt; want != got {
-		o.UI.PrintErr(fmt.Sprintf("fetched %d profiles out of %d", got, want))
+
+	if normalize {
```
aalexand: There is still more code under this condition than I'd expect. Can only "psrc.Normalize(pbase)" call be under the condition? I am trying to ensure that the same code exercises for the most part when normalization is on or off so that any bugs can be caught easier.

nolanmar511: Okay -- I think I changed the code to do this.
Still looking at having chunkedGrab return 2 profiles.

internal/driver/fetch.go
```diff
@@ -110,7 +156,7 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
 // chunkedGrab fetches the profiles described in source and merges them into
 // a single profile. It fetches a chunk of profiles concurrently, with a maximum
 // chunk size to limit its memory usage.
-func chunkedGrab(sources []profileSource, fetch plugin.Fetcher, obj plugin.ObjTool, ui plugin.UI) (*profile.Profile, plugin.MappingSources, bool, int, error) {
+func chunkedGrab(sources []profileSource, fetch plugin.Fetcher, obj plugin.ObjTool, ui plugin.UI) (*profile.Profile, plugin.MappingSources, int, bool, error) {
```
aalexand: Curious: why did the order of bool and int ret values change?

internal/driver/fetch.go
```diff
@@ -41,32 +41,78 @@ import (
...
 		})
 	}
+
+	bases := make([]profileSource, 0, len(s.Base))
 	for _, src := range s.Base {
-		sources = append(sources, profileSource{
+		bases = append(bases, profileSource{
 			addr:   src,
 			source: s,
 			scale:  -1,
```
aalexand: Since we are fetching input and base profiles independently now, we could also remove setting up the scaling factor here and just scale the combined base profile by -1 below in this function.

internal/driver/fetch.go
nolanmar511: Oops -- I'd briefly gotten rid of the in return when I tried to move some of the error checking, realized that wouldn't work, and then added back the return incorrectly. 

I'll change it back.

internal/driver/fetch.go
```diff
@@ -41,32 +41,78 @@ import (
...
 		})
 	}
-	p, msrcs, save, cnt, err := chunkedGrab(sources, o.Fetch, o.Obj, o.UI)
-	if err != nil {
-		return nil, err
+
+	// retrieve source and base profiles
+	wg := sync.WaitGroup{}
```
aalexand: I'd move the code which grabs the input and base profiles in parallel here into a separate function like grabSourcesAndBases().

internal/driver/fetch.go
```diff
@@ -107,6 +137,36 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
...
+	}()
+	go func() {
+		defer wg.Done()
+		pb, mbase, savebase, countbase, errbase = chunkedGrab(bases, fetch, obj, ui)
+	}()
+	wg.Wait()
+	save := savesrc || savebase
+
+	var err error
+	if errsrc != nil && errbase != nil {
```
aalexand: How about simplifying this: if there is a problem fetching source profiles - return that. Else if there is a problem returning base profiles - return that. Else return no error. I agree that printing out both errors when both errors happen may be nice, but I think it's very rarely required so it's better to keep the code simple.

internal/driver/fetch.go
```diff
@@ -107,6 +137,36 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
 	return p, nil
 }
 
+func grabSourceAndBases(sources, bases []profileSource, fetch plugin.Fetcher, obj plugin.ObjTool, ui plugin.UI) (*profile.Profile, *profile.Profile, plugin.MappingSources, plugin.MappingSources, bool, int, int, error) {
+	wg := sync.WaitGroup{}
+	wg.Add(2)
+	var p, pb *profile.Profile
```
aalexand: Nit: p -> psrc for consistency?

internal/driver/fetch.go
```diff
@@ -107,6 +137,36 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
 	return p, nil
 }
 
+func grabSourceAndBases(sources, bases []profileSource, fetch plugin.Fetcher, obj plugin.ObjTool, ui plugin.UI) (*profile.Profile, *profile.Profile, plugin.MappingSources, plugin.MappingSources, bool, int, int, error) {
```
aalexand: Can we avoid returning the profile counts and instead return an error if zero number of source profiles was fetched or if (non-zero number of base profiles was requested and zero was fetched)?

internal/driver/fetch.go
```diff
@@ -107,6 +137,36 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
...
+		pb, mbase, savebase, countbase, errbase = chunkedGrab(bases, fetch, obj, ui)
+	}()
+	wg.Wait()
+	save := savesrc || savebase
+
+	var err error
+	if errsrc != nil && errbase != nil {
+		err = fmt.Errorf("Problem fetching source profiles: %s, Problem fetching base profiles: %s", errsrc.Error(), errbase.Error())
+	} else if errsrc != nil {
+		err = fmt.Errorf("Problem fetching source profiles: %s,", errsrc)
```
aalexand: Nit: we sometimes capitalize the error message (like here - "Problem ...") and sometimes we use lowercase (like above - "failed to ..."). Should we try to be more consistent?

internal/driver/fetch.go
aalexand: Should the name of the function be grabSourcesAndBases as input is a list of sources?

internal/driver/fetch.go
```diff
@@ -41,30 +41,60 @@ import (
...
+		o.UI.PrintErr(fmt.Sprintf("fetched %d source profiles out of %d", got, want))
 	}
-	if want, got := len(sources), cnt; want != got {
-		o.UI.PrintErr(fmt.Sprintf("fetched %d profiles out of %d", got, want))
+	if want, got := len(bases), countbase; want != got {
+		o.UI.PrintErr(fmt.Sprintf("fetched %d base profiles out of %d", got, want))
+	}
+
+	// normalize by base, if necessary
+	if normalize {
```
aalexand: This should probably be under `if pb != nil` condition below, just in case.

internal/driver/fetch.go
```diff
@@ -41,30 +41,60 @@ import (
...
+		if countbase == 0 {
```
aalexand: This won't be needed if we make grabSourcesAndBases return an error if base profiles were requested but neither were fetched.

internal/driver/fetch.go
```diff
@@ -41,30 +41,60 @@ import (
 // there are some failures. It will return an error if it is unable to
 // fetch any profiles.
 func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
-	sources := make([]profileSource, 0, len(s.Sources)+len(s.Base))
+	normalize := s.Normalize
```
aalexand: Curious: why do you need this local variable? This only saves two characters.

nolanmar511: I also only use it once -- so really good point.
Fixed this!

internal/driver/fetch.go
nolanmar511: Done!

internal/driver/fetch.go
nolanmar511: Good point -- I did that.

internal/driver/fetch.go
nolanmar511: Done!

nolanmar511: done!

internal/driver/fetch.go
nolanmar511: Thanks! -- changed this

internal/driver/fetch.go
nolanmar511: I switched error messages to uppercase for errors I added/modified. 
Should I go through everything to make it consistent? (And would lowercase be better?)

nolanmar511:
Just to note this, I believe this change will impact the behavior when -base is used, and either no base or no source profiles are fetched. Prior to this change, I don't believe an error was returned. Now an error will be returned.

internal/driver/fetch.go
```diff
@@ -107,6 +127,47 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
...
+	}()
+	go func() {
+		defer wg.Done()
+		pbase, mbase, savebase, countbase, errbase = chunkedGrab(bases, fetch, obj, ui)
+	}()
+	wg.Wait()
+	save := savesrc || savebase
+
+	if errsrc != nil {
+		return nil, nil, nil, nil, false, fmt.Errorf("Problem fetching source profiles: %s", errsrc.Error())
```
aalexand: Instead of `fmt.Errorf("Problem fetching source profiles: %s", errsrc.Error())` the usual way to do it is `fmt.Errorf("Problem fetching source profiles: %v", errsrc)`

profile/merge.go
```diff
@@ -85,6 +86,42 @@ func Merge(srcs []*Profile) (*Profile, error) {
 	return p, nil
 }
 
+// Normalize normalizes the source profiles by multiplying each value in profiles by the
```
aalexand: I think the comment is out of date since it still talks about multiple profiles.

profile/merge.go
```diff
@@ -85,6 +86,42 @@ func Merge(srcs []*Profile) (*Profile, error) {
...
+// absolute value of the ratio of the sum of the base profile value's of that sample type
+// to the sum of the source profiles value of that sample type.
+func (p *Profile) Normalize(pb *Profile) error {
+
+	err := p.compatible(pb)
```
aalexand: Use one line if statement here:

```
if err := p.compatible(pb); err != nil {
  return err
}

File: profile/profile_test.go [hunk 16]
```diff
@@ -420,14 +497,94 @@ func TestMergeAll(t *testing.T) {
...
 			t.Errorf("merge got wrong value at %s : %d instead of %d", tb, samples[tb], s.Value[0]*10)
 		}
 	}
 }
 
+func TestNormalizeBySameProfile(t *testing.T) {
+	pb := testProfile1.Copy()
+	p := testProfile1.Copy()
+
+	err := p.Normalize(pb)
```
aalexand: Use one line if.

profile/profile_test.go
```diff
@@ -420,14 +497,94 @@ func TestMergeAll(t *testing.T) {
...
+			}
+		}
+	}
+}
+
+func TestNormalizeByDifferentProfile(t *testing.T) {
+	p := testProfile1.Copy()
+	pb := testProfile2.Copy()
+
+	err := p.Normalize(pb)
```
aalexand: Use one line if.

profile/profile_test.go
```diff
@@ -420,14 +497,94 @@ func TestMergeAll(t *testing.T) {
...
+			}
+		}
+	}
+}
+
+func TestNormalizeIncompatibleProfiles(t *testing.T) {
+	p := testProfile1.Copy()
+	pb := testProfile3.Copy()
+
+	err := p.Normalize(pb)
```
aalexand: Use one line if.

profile/merge.go
nolanmar511: thanks!

internal/driver/fetch.go
```diff
@@ -41,34 +41,44 @@ import (
...
-		sources = append(sources, profileSource{
+		bases = append(bases, profileSource{
 			addr:   src,
 			source: s,
-			scale:  -1,
 		})
 	}
-	p, msrcs, save, cnt, err := chunkedGrab(sources, o.Fetch, o.Obj, o.UI)
+
+	// retrieve source and base profiles
```
aalexand: For my taste this comment is redundant as well.

internal/driver/fetch_test.go
```diff
@@ -206,6 +206,138 @@ func TestFetch(t *testing.T) {
 	}
 }
 
+func TestFetchWithBase(t *testing.T) {
+	const path = "testdata/"
+	// set up testcases
```
aalexand: Remove  comment.

internal/driver/fetch_test.go
```diff
@@ -206,6 +206,138 @@ func TestFetch(t *testing.T) {
...
+	testcases := []testcase{
+		{
+			[]string{
+				path + "cppbench.cpu",
+			},
+			[]string{
+				path + "cppbench.cpu",
+			},
+			false,
+			"not normalized, base is same as non-base",
```
aalexand: Maybe word it as "not normalized, base is same as source profile"? We call these "source profiles" in fetch.go.

internal/driver/fetch_test.go
```diff
@@ -206,6 +206,138 @@ func TestFetch(t *testing.T) {
...
+			[]string{
+				path + "cppbench.cpu",
+			},
+			[]string{
+				path + "cppbench.cpu",
+				path + "cppbench.cpu",
+			},
+			false,
+			"not normalized, single non-base, multiple base (all profiles same)",
+			true,
```
aalexand: Instead of just validating whether there are samples or not, would it be better to validate the specific value count for one of functions in the profile? Feel free to decline if it seems redundant / not useful.

internal/driver/fetch_test.go
```diff
@@ -206,6 +206,138 @@ func TestFetch(t *testing.T) {
...
+			},
+			[]string{
+				path + "cppbench.cpu",
+			},
+			false,
+			"not normalized, base is same as non-base",
+			false,
+		},
+		{
+			[]string{
```
aalexand: Nit: you could shift prepending the path prefix from the test cases to the code of the test. And make the array declarations here one liners then, it would make the source more compact, like

```
		{
			[]string{"cppbench.cpu"},
			[]string{"cppbench.cpu", "cppbench.cpu"},
			false,
			"description",
			true,
		}
```

internal/driver/fetch_test.go
```diff
@@ -206,6 +206,138 @@ func TestFetch(t *testing.T) {
...
 }
 
+func TestFetchWithBase(t *testing.T) {
+	const path = "testdata/"
+	// set up testcases
+	type testcase struct {
+		sources         []string
+		bases           []string
+		normalize       bool
+		testDescription string
```
aalexand: Nit: testDescription field typically goes first and is named just "desc".

internal/driver/fetch_test.go
```diff
@@ -206,6 +206,109 @@ func TestFetch(t *testing.T) {
...
+		{
+			"normalized, different base and source",
+			[]string{path + "cppbench.cpu"},
+			[]string{path + "go.crc32.cpu"},
+			true,
+			true,
+		},
+	}
+
+	for _, tc := range testcases {
```
aalexand: Not sure if you want to change this now, but sometimes running test cases using https://golang.org/pkg/testing/#T.Run makes the test code more readable. It allows using t.Fatal when a check fails that the test shouldn't skip, instead of using t.Error() + continue. And will make defer() statement execute for each test case individually which is sometimes expected.

internal/driver/fetch_test.go
```diff
@@ -206,6 +206,109 @@ func TestFetch(t *testing.T) {
...
+		baseVars := pprofVariables
+		pprofVariables = baseVars.makeCopy()
+		defer func() { pprofVariables = baseVars }()
```
aalexand: This usage of defer() here might be a bit misleading. It defers things until the test function exits, not until the for loop iteration finishes. So when the test finishes, there will be a bunch of deferred calls where the last one will be the first call deferred, which will restore the pprof variables to the right, oldest value. But it would probably enough to restore pprofVariables once, with a single defer call at the beginning of the test function?

aalexand:
Looks like you need to run `go fmt` - https://travis-ci.org/google/pprof/builds/255072054

PR #167, opened 2017-07-19
Removed tag scaling from dotgraph.go

nolanmar511:
When viewing certain graphs for heapz data, users saw that some graphs had boxes (nodelets) containing "0..0".

This was because, when minimum (the default value) is specified for the -unit option and there are enough tags (more that 4) that some tags have to be merged with other tags, then unit used to display values in the graph is selected to be as large as possible while still capturing values in the nodes, but without consideration to the size of the values in the nodelets. So, if the smallest node has a value of 10MB, and nodelet has a value of 10B, (and is a sibling node to a node containing a range) the value for that nodelet would be reported in MB (so it would look like zero, after rounding).

When there are 4 or less tags already, then the nodelets are labelled with the tag's name, which had the units scaled in a manner more consistent with what one would expect. 

So, this change modifies the code so that all nodelets are labelled based on tag names.


internal/graph/dotgraph_test.go
```diff
@@ -299,32 +299,32 @@ func TestTagCollapse(t *testing.T) {
...
 
 	tagSource := []*Tag{
-		makeTag("12mb", "mb", 12, 100, 100),
-		makeTag("1kb", "kb", 1, 1, 1),
-		makeTag("1mb", "mb", 1, 1000, 1000),
-		makeTag("2048mb", "mb", 2048, 1000, 1000),
-		makeTag("1b", "b", 1, 100, 100),
-		makeTag("2b", "b", 2, 100, 100),
-		makeTag("7b", "b", 7, 100, 100),
+		makeTag("12MB", "mb", 12, 100, 100),
```
aalexand: It is strange that you had to rename the inputs. I would expect that we still re-format "12mb" as "12MB" so that the output is consistent.

internal/graph/dotgraph.go
```diff
@@ -481,7 +475,7 @@ func (b *builder) tagGroupLabel(g []*Tag) (label string, flat, cum int64) {
 	if dc != 0 {
 		c = c / dc
 	}
-	return formatTag(min.Value, min.Unit) + ".." + formatTag(max.Value, max.Unit), f, c
+	return min.Name + ".." + max.Name, f, c
```
aalexand: It is not clear from the test changes whether you added a test case which would reproduce the bug before the fix?

internal/graph/dotgraph.go
```diff
@@ -450,14 +449,9 @@ func tagDistance(t, u *Tag) float64 {
...
 
 func (b *builder) tagGroupLabel(g []*Tag) (label string, flat, cum int64) {
-	formatTag := b.config.FormatTag
-	if formatTag == nil {
-		formatTag = measurement.Label
-	}
-
 	if len(g) == 1 {
 		t := g[0]
-		return formatTag(t.Value, t.Unit), t.FlatValue(), t.CumValue()
```
aalexand: I can see that formatTag() is invoked even for a single tag (when len(g) == 1). So I am not entirely clear why the bug only manifested for grouped tags, could you elaborate?

nolanmar511:
Doing this as suggested was different enough that I created a new branch, so will make new pull request for this issue.

PR #168, opened 2017-07-20
Modified dotgraph tag labels so they aren't scaled to graph's unit

nolanmar511:
Before this change, some tags (tags which were grouped into ranges or were siblings to nodes grouped into ranges) would be scaled so that their units matched the units automatically selected for the graph's nodes -- sometimes making ranges like "0..0" appear.

This changes makes it so that tag labels are automatically scaled to reasonable units, rather than scaled to the units of the rest of the graph.


internal/driver/driver_test.go
```diff
@@ -796,6 +799,166 @@ func heapProfile() *profile.Profile {
 	}
 }
 
+func heapProfileMedium() *profile.Profile {
```
nolanmar511: I switched to making the profile used to test from heapProfile(), and now call it "heap_sizetags"

internal/graph/dotgraph.go
```diff
@@ -481,7 +475,11 @@ func (b *builder) tagGroupLabel(g []*Tag) (label string, flat, cum int64) {
 	if dc != 0 {
 		c = c / dc
 	}
-	return formatTag(min.Value, min.Unit) + ".." + formatTag(max.Value, max.Unit), f, c
+
+	// tags are not scaled with the selected output unit because tags are often much smaller
```
aalexand: Nits:

* Capitalize the first word ("Tags") so that the first sentence is a proper one.
* It's common to format the comment text in 80 columns.

PR #175, opened 2017-07-25
Removed tests which failed on some systems

nolanmar511:
These tests passed in Appveyor and in Travis, but failed when using blaze, preventing internal release.

The test with hostport  equaling "127.0.0.1" likely fails because "127.0.0.1" is interpreted as ipv4, so will fail on systems where only ipv6 is supported.

The test with hostport equaling "localhost:12344" likely fails with blaze because the allowable port range is controlled. This test will likely also fail if the port is bound to other processes when the test runs.

### DIFF
--- a/internal/driver/webui_test.go
+++ b/internal/driver/webui_test.go
@@ -209,16 +209,6 @@ func TestNewListenerAndURL(t *testing.T) {
 			wantLocal: true,
 		},
 		{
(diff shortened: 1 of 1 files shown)


aalexand:
Can you update the description to explain on which systems the tests were failing and why you think removing these tests is a good idea?  E.g. on the systems where the tests passed, they were testing something useful perhaps?

PR #176, opened 2017-07-26
Fix unexpected webui permission error on an IPv6 configured machine

### LINKED ISSUE #174: local web UI returning `permission denied` on localhost
I ran `pprof -http :9090` on a MacOS host, and when it opened in my browser it just shows "permission denied".

Adding the host to the error message reveals that `host` is `::1`, not `[::1]` as expected by `isLocalhost`.

ghemawat:
When net.SplitHostPort is called, it will return the string "::1",
but the code was only checking for "[::1]".

Fixes #174.

### DIFF
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -105,7 +105,7 @@ func newListenerAndURL(hostport string) (ln net.Listener, url string, isLocal bo
 }
 
 func isLocalhost(host string) bool {
-	for _, v := range []string{"localhost", "127.0.0.1", "[::1]"} {
+	for _, v := range []string{"localhost", "127.0.0.1", "[::1]", "::1"} {
 		if host == v {
 			return true
 		}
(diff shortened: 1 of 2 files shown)


nolanmar511:
Is something like `pprof -http [::1]:` a valid user input? When I tried that with this, basically  just a new tab opened in my browser. 
If this is valid, I think newListenerAndURL() should maybe use net.JoinHostPort() to make the URL (line 102)

ghemawat:
> Is something like pprof -http [::1]: a valid user input? When I tried that with this, basically just a new tab opened in my browser.

Probably.  We seem to accept "localhost:", so yes, I think we should make sure that "[::1]:" and "[::1]:<someport>" also work.

Seems separable from this pull request though?

nolanmar511:
Totally true -- mistakenly grouped the problems since they had the same-ish cause, but the fixes are distinct.

Issue #152, opened 2017-07-07
-traces should include tags


ahh:
Attached a profile (zipped gzip is silly, but Github won't let me upload "heapz".)
```
pprof -raw heapz
...

Samples:
objects/count space/bytes[dflt]
       7438   60936269: 1 2 3 4 5 6
                bytes:[8192] request:[8191]
        238   62466710: 7 8 9 10 5 6
                bytes:[262144] request:[262143]
       3341   54739221: 7 8 9 10 5 6
                bytes:[16384] request:[16383]
       2064   67634517: 1 2 3 4 5 6
                bytes:[32768] request:[32767]
      29710   60847108: 1 2 3 4 5 6
                bytes:[2048] request:[2047]
        396   51920894: 1 2 3 4 5 6
                bytes:[131072] request:[131071]
      63503   65027585: 7 8 9 10 5 6
                bytes:[1024] request:[1023]
       1105   72423082: 7 8 9 10 5 6
                bytes:[65536] request:[65535]
        108   56884975: 11 12 13 3 4 5 6
                bytes:[524288] request:[524287]
      15887   65075220: 7 8 9 10 5 6
                bytes:[4096] request:[4095]
...
pprof -traces heapz
...

-----------+-------------------------------------------------------
   69.07MB   [UNKNOWN]
             [UNKNOWN]
             [UNKNOWN]
             [UNKNOWN]
             __libc_start_main
             [UNKNOWN]
-----------+-------------------------------------------------------
   54.25MB   [UNKNOWN]
             [UNKNOWN]
             [UNKNOWN]
             [UNKNOWN]
             [UNKNOWN]
             __libc_start_main
             [UNKNOWN]
-----------+-------------------------------------------------------
   62.06MB   [UNKNOWN]
             [UNKNOWN]
             [UNKNOWN]
             [UNKNOWN]
             __libc_start_main
             [UNKNOWN]
-----------+-------------------------------------------------------
...etc
```

the "traces" output (which is much more useful (as it can be filtered, etc)) should contain the bytes/request tags.
[heapz.zip](https://github.com/google/pprof/files/1131685/heapz.zip)

ahh:
ping?

aalexand:
@nolanmar511 could you look at it?

aalexand:
Should be fixed by #179

PR #179, opened 2017-07-27
Print numeric tags when printing the trace

nolanmar511:
Addresses #152


internal/driver/driver_test.go
```diff
@@ -398,6 +399,18 @@ func (testFetcher) Fetch(s string, d, t time.Duration) (*profile.Profile, string
 		for _, s := range p.Sample {
 			s.NumLabel["bytes"] = append(s.NumLabel["bytes"], tags...)
 		}
+	case "heap_tags":
+		p = heapProfile()
+		sizeTags := []int64{2, 4, 8, 16, 32, 64, 128, 256}
```
aalexand: I think that the common usage of numeric tags like "bytes" in the heap traces is that each sample has just one value for the given tag, recording separate samples as needed. Otherwise it would not be possible to break the allocations down precisely. The test should probably be reflective of that.

nolanmar511: Thanks -- I've changed this!

internal/driver/driver_test.go
```diff
@@ -398,6 +399,17 @@ func (testFetcher) Fetch(s string, d, t time.Duration) (*profile.Profile, string
...
 		}
+	case "heap_tags":
+		p = heapProfile()
+
+		for i := 0; i < len(p.Sample); i += 2 {
+			s := p.Sample[i]
+			if s.Label == nil {
+				s.Label = make(map[string][]string)
+			}
+			s.NumLabel["request"] = (s.NumLabel["bytes"])
```
aalexand: Remove parenthesis around `s.NumLabel["bytes"]` ?

internal/report/report.go
```diff
@@ -755,6 +755,15 @@ func printTraces(w io.Writer, rpt *Report) error {
 		}
 		sort.Strings(labels)
 		fmt.Fprint(w, strings.Join(labels, ""))
+
+		// Print any numeric labels for the sample
+		var numLabels []string
+		for k, v := range sample.NumLabel {
+			numLabels = append(numLabels, fmt.Sprintf("%10s:  %s\n", k, strings.Trim(fmt.Sprintf("%v", v), "[]")))
```
aalexand: Curious: why is strings.Trim(..., "[]") needed?

aalexand: I think "%v" should not be used in the output to the user. It's easy to leak something unexpected there. I think it should be "%s".

internal/report/report.go
```diff
@@ -755,6 +755,15 @@ func printTraces(w io.Writer, rpt *Report) error {
...
+		}
+		sort.Strings(numLabels)
```
aalexand: Sorting numbers as strings doesn't sound right. "100" < "2"?

nolanmar511: I'm sorting the keys -- like "bytes" and "requests". This makes sure I'll get bytes before requests.

Is my naming unclear?

internal/report/report.go
nolanmar511: I switched it to %d. Does that work, or should I iterate over the array to build a string for the values?

nolanmar511: v is an array, and I didn't like it when the value was printed out in square brackets.

internal/report/report.go
aalexand: Got it. No, it's fine, I misread.

internal/report/report.go
aalexand: Is %d happy to print an array? Hmm, if not maybe leave what you had with %v. It's weird but not enough to unroll into a loop here. The only string thing is that if the list is nil it would print nil which is weird. But apparently this should never happen.

nolanmar511: %d was able to print the array.

aalexand:
By the way, use "Fixes #152" syntax, rather than "Addresses #152", because [this](https://help.github.com/articles/closing-issues-using-keywords).

PR #186, opened 2017-08-08
Modified Prune to account for unsimplified names with return type

nolanmar511:


profile/prune_test.go
```diff
@@ -228,3 +230,50 @@ Locations
      5: 0x0 Foo::operator()(::Bar) fun.c:1 s=0
 Mappings
 `
+
+var locs3 = []*Location{
```
aalexand: Instead of keeping extending the test cases here, I'd add a table-based test for just simplifyFunc. It would be easier to add more test cases as simplifyFunc gets more complex.

nolanmar511: I've added a test for simplifyFunc

nolanmar511:
Closing this -- we want to figure out why a few functions had their return type after symbolization, but most didn't.

PR #194, opened 2017-08-19
Extend web UI to contain a tabular "top" display

ghemawat:
Details:

* New /top URL that generates a tabular display.
* Split-up template into reusable templates (shared by / and /top).
* Sanitized web page header to contain a single menubar.
* Fixed bug in javascript node name handling (names were regexp-quoted, but were being treated as plain strings).

Attached zip file contains sample output.
[sample.zip](https://github.com/google/pprof/files/1236737/sample.zip)

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -195,7 +195,6 @@ Functions
 </div>
 </div>
 
-{{if (eq .Type "dot")}}
 <div class="menu-header">
 Refine
 <div class="menu">
@@ -205,7 +204,6 @@ Refine
 <button title="{{.Help.show}}" id="show">Show</button>
 </div>
 </div>
-{{end}}
 
 <input id="searchbox" type="text" placeholder="Search regexp" autocomplete="off" autocapitalize="none" size=40>
 
@@ -234,7 +232,7 @@ Refine
 
 </div>
 {{template "script" .}}
-<script>viewer({{.Nodes}})</script>
+<script>viewer({{.BaseURL}}, {{.Nodes}})</script>
 </body>
 </html>
 {{end}}
@@ -457,7 +455,7 @@ function initPanAndZoom(svg, clickHandler) {
   svg.addEventListener("wheel", handleWheel, true)
 }
 
-function viewer(nodes) {
+function viewer(baseUrl, nodes) {
   'use strict';
 
   // Elements
@@ -482,16 +480,16 @@ function viewer(nodes) {
     if (detailsText != null) detailsText.style.display = "none"
   }
 
-  function handleReset() { window.location.href = "/" }
+  function handleReset() { window.location.href = baseUrl }
   function handleTop() { navigate("/top", "f", false) }
   function handleGraph() { navigate("/", "f", false) }
   function handleList() { navigate("/weblist", "f", true) }
   function handleDisasm() { navigate("/disasm", "f", true) }
   function handlePeek() { navigate("/peek", "f", true) }
-  function handleFocus() { navigate("/", "f", false) }
-  function handleShow() { navigate("/", "s", false) }
-  function handleIgnore() { navigate("/", "i", false) }
-  function handleHide() { navigate("/", "h", false) }
+  function handleFocus() { navigate(baseUrl, "f", false) }
+  function handleShow() { navigate(baseUrl, "s", false) }
+  function handleIgnore() { navigate(baseUrl, "i", false) }
+  function handleHide() { navigate(baseUrl, "h", false) }
 
   function handleKey(e) {
     if (e.keyCode != 13) return
@@ -760,7 +758,7 @@ function viewer(nodes) {
 </div>
 
 {{template "script" .}}
-<script>viewer({{.Nodes}})</script>
+<script>viewer({{.BaseURL}}, {{.Nodes}})</script>
 </body>
 </html>
 {{end}}
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -144,9 +144,8 @@ button {
   text-align: right;
   padding-left: 1em;
 }
-#toptable tr th:nth-child(6) {
-  text-align: left;
-}
+#toptable tr th:nth-child(6) { text-align: left; }
+#toptable tr th:nth-child(7) { text-align: left; }
 #toptable tr td {
   padding-left: 1em;
   font: monospace;
@@ -154,9 +153,8 @@ button {
   white-space: nowrap;
   cursor: default;
 }
-#toptable tr td:nth-child(6) {
-  text-align: left;
-}
+#toptable tr td:nth-child(6) { text-align: left; }
+#toptable tr td:nth-child(7) { text-align: left; }
 .hilite {
   background-color: #ccf;
 }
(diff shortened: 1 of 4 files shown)


internal/driver/webhtml.go
```diff
@@ -511,6 +615,11 @@ function dotviewer(nodes) {
     return null
   }
 
+  // convert a string to a regexp that matches that string.
+  function quotemeta(str) {
+    return str.replace(/([\\\.?+*\[\](){}])/g, '\\$1')
```
rauls5382: I selected a function that showed "std::abs (inline)" and hit "show", and the profile came back empty. I think it we should strip the "(inline)" part for the selected expressions (or even better have it on a separate column.

internal/driver/webhtml.go
```diff
@@ -53,90 +57,187 @@ button {
...
+<div class="menu">
+{{if (ne .Type "top")}}
+  <button title="{{.Help.top}}" id="topbtn">Top</button>
+{{end}}
+{{if (ne .Type "dot")}}
+  <button title="{{.Help.graph}}" id="graphbtn">Graph</button>
+{{end}}
+<hr>
+<button title="{{.Help.details}}" id="details">Details</button>
+<button title="{{.Help.reset}}" id="reset">Reset</button>
```
aalexand: Would it be better to put "Reset" into the "Refine" menu, since its primary job appears to be throwing away the current focus, ignore, show and hide filters?

internal/driver/webhtml.go
```diff
@@ -53,90 +57,187 @@ button {
...
-  display: none;
-  position: fixed;
-  background-color: #ffffff;
-  border: 1px solid black;
-  box-shadow: 2px 2px 2px 0px #aaa;
-  top: 20px;
-  right: 20px;
-  z-index: 1;
+#closedetails {
+  display: inline;
```
josef-jelinek: display style is overridden by float, so has no effect. It is often easier now to use display: flex on a container for element alignment than dealing with floats, and clearing them.

internal/driver/webhtml.go
```diff
@@ -491,6 +584,17 @@ function dotviewer(nodes) {
   }
 
   function setBackground(elem, set) {
+    // Handle table row highlighting.
+    if (elem.nodeName == "TR") {
+      if (set) {
```
josef-jelinek: depending if not supporting IE10 is not an issue (Edge works fine), this can be simplified to
elem.classList.toggle("hilite", set)

internal/driver/webhtml.go
josef-jelinek: Are we concerned about other characters like $ ^ |, or are they supposed to be a part of the regexp (then it would be hard to escape them as \$ would turn into \\$)?

internal/driver/webhtml.go
```diff
@@ -547,38 +656,110 @@ function dotviewer(nodes) {
...
   }
 
+  function handleTopClick(e) {
+    // Walk back until we find TR and then get the Name column (index 5)
+    let elem = e.target
+    while (elem != null && elem.nodeName != "TR") {
+      elem = elem.parentElement
+    }
+    if (elem == null) return
+    if (elem.children.length != 6) return
```
josef-jelinek: why not use ||?

internal/driver/webhtml.go
ghemawat: This is for the close button on the details pop up which I do want to flow independently of any other info in the pop up (it floats at the top-right corner of the pop-up).  The "float" seems pretty natural here (also no clearing is being done here)?

Not sure how flex would work to achieve this, but if you think flex would be better and can give me a pointer, I can try to incorporate that.

internal/driver/webhtml.go
ghemawat: Good point. Done.

internal/driver/webhtml.go
ghemawat: Done.

internal/driver/webhtml.go
ghemawat: Done (added a new column for the inline label).

ghemawat: Added ^, $, |.

Also double-checked with pcre syntax that I am not missing any other special characters.

internal/driver/webhtml.go
ghemawat: Too much Go style on my brain.  Fixed.

internal/driver/webhtml.go
josef-jelinek: flex is probably too different from the behavior described (the button would need to be part of some header).
However, the display: inline is still redundant as float takes over.

ghemawat: Done.

ghemawat:
Travis build seems to have failed due to an environmental problem (github.com not resolved).  Any way to retry it without pushing another commit?

$ go get -t -v ./...
github.com/ianlancetaylor/demangle (download)
# cd .; git clone https://github.com/ianlancetaylor/demangle /Users/travis/gopath/src/github.com/ianlancetaylor/demangle
Cloning into '/Users/travis/gopath/src/github.com/ianlancetaylor/demangle'...
fatal: unable to access 'https://github.com/ianlancetaylor/demangle/': Could not resolve host: github.com
package github.com/ianlancetaylor/demangle: exit status 128
The command "eval go get -t -v ./... " failed. Retrying, 2 of 3.
github.com/ianlancetaylor/demangle (download)
# cd .; git clone https://github.com/ianlancetaylor/demangle /Users/travis/gopath/src/github.com/ianlancetaylor/demangle
Cloning into '/Users/travis/gopath/src/github.com/ianlancetaylor/demangle'...
fatal: unable to access 'https://github.com/ianlancetaylor/demangle/': Could not resolve host: github.com
package github.com/ianlancetaylor/demangle: exit status 128
The command "eval go get -t -v ./... " failed. Retrying, 3 of 3.
github.com/ianlancetaylor/demangle (download)
# cd .; git clone https://github.com/ianlancetaylor/demangle /Users/travis/gopath/src/github.com/ianlancetaylor/demangle
Cloning into '/Users/travis/gopath/src/github.com/ianlancetaylor/demangle'...
fatal: unable to access 'https://github.com/ianlancetaylor/demangle/': Could not resolve host: github.com
package github.com/ianlancetaylor/demangle: exit status 128
The command "eval go get -t -v ./... " failed 3 times.

nolanmar511:
Thanks! -- the failed travis test is re-running now.

internal/driver/webhtml.go
rauls5382: Thanks. However, my profile has one function with a very long name, so the inline column gets pushed to the right so much that I cannot see it anymore. I could scroll-right to see it, but now it is so far away that it is hard to see which function is which.

Could we place the Inline column to the left of the Name column so that it always fits, or cap the width of the Name column?

ghemawat: I put it on the right so that it matches the text display.

We could easily cap the width of the Name column (I think a CSS max-width property might suffice).

Could you save the html file that contains the long name and place it somewhere accessible so we can play around with different possible layouts?

Also, should we deal with the long Name column issue in a separate change?

PR #210, opened 2017-08-24
Stop treating -symbolize=force as the default

rauls5382:
The code parsing symbolize options was treating the empty option as
"force", effectively making -symbolize=force the default. This would
unnecessarily symbolize profiles that were already symbolized.

### DIFF
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -42,20 +42,23 @@ type Symbolizer struct {
 // test taps for dependency injection
 var symbolzSymbolize = symbolz.Symbolize
 var localSymbolize = doLocalSymbolize
+var demangleFunction = Demangle
 
 // Symbolize attempts to symbolize profile p. First uses binutils on
 // local binaries; if the source is a URL it attempts to get any
 // missed entries using symbolz.
 func (s *Symbolizer) Symbolize(mode string, sources plugin.MappingSources, p *profile.Profile) error {
-	remote, local, force, demanglerMode := true, true, false, ""
+	remote, local, fast, force, demanglerMode := true, true, false, false, ""
 	for _, o := range strings.Split(strings.ToLower(mode), ":") {
 		switch o {
 		case "":
 			continue
 		case "none", "no":
 			return nil
-		case "local", "fastlocal":
+		case "local":
 			remote, local = false, true
+		case "fastlocal":
+			remote, local, fast = false, true, true
 		case "remote":
 			remote, local = true, false
 		case "force":
@@ -77,7 +80,7 @@ func (s *Symbolizer) Symbolize(mode string, sources plugin.MappingSources, p *pr
 	var err error
 	if local {
 		// Symbolize locally using binutils.
-		if err = localSymbolize(mode, p, s.Obj, s.UI); err != nil {
+		if err = localSymbolize(p, fast, force, s.Obj, s.UI); err != nil {
 			s.UI.PrintErr("local symbolization: " + err.Error())
 		}
 	}
@@ -87,7 +90,7 @@ func (s *Symbolizer) Symbolize(mode string, sources plugin.MappingSources, p *pr
 		}
 	}
 
-	Demangle(p, force, demanglerMode)
+	demangleFunction(p, force, demanglerMode)
 	return nil
 }
 
@@ -136,18 +139,10 @@ func statusCodeError(resp *http.Response) error {
 // doLocalSymbolize adds symbol and line number information to all locations
 // in a profile. mode enables some options to control
 // symbolization.
-func doLocalSymbolize(mode string, prof *profile.Profile, obj plugin.ObjTool, ui plugin.UI) error {
-	force := false
-	// Disable some mechanisms based on mode string.
-	for _, o := range strings.Split(strings.ToLower(mode), ":") {
-		switch {
-		case o == "force":
-			force = true
-		case o == "fastlocal":
-			if bu, ok := obj.(*binutils.Binutils); ok {
-				bu.SetFastSymbolization(true)
-			}
-		default:
+func doLocalSymbolize(prof *profile.Profile, fast, force bool, obj plugin.ObjTool, ui plugin.UI) error {
+	if fast {
+		if bu, ok := obj.(*binutils.Binutils); ok {
+			bu.SetFastSymbolization(true)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


rauls5382:
Added more testing (and minor refactoring based on it). PTAL

internal/symbolizer/symbolizer.go
```diff
@@ -75,7 +80,7 @@ func (s *Symbolizer) Symbolize(mode string, sources plugin.MappingSources, p *pr
 	var err error
 	if local {
 		// Symbolize locally using binutils.
-		if err = localSymbolize(mode, p, s.Obj, s.UI); err != nil {
+		if err = localSymbolize(p, fast, force, s.Obj, s.UI); err != nil {
```
aalexand: Curious: looks like there is no way to force remote symbolization so it's always either forced or always skips already symbolized mappings?

internal/symbolizer/symbolizer_test.go
```diff
@@ -149,11 +167,28 @@ func symbolzMock(sources plugin.MappingSources, syms func(string, string) ([]byt
...
 
-func localMock(mode string, p *profile.Profile, obj plugin.ObjTool, ui plugin.UI) error {
-	p.Comments = append(p.Comments, "local="+mode)
+func localMock(p *profile.Profile, fast, force bool, obj plugin.ObjTool, ui plugin.UI) error {
+	var args []string
+	if fast {
+		args = append(args, "fast")
+	}
+	if force {
+		args = append(args, "force")
```
aalexand: Maybe use "force-local" string here to make the wanted comment string in the test cases a bit clearer?

internal/symbolizer/symbolizer_test.go
```diff
@@ -149,11 +167,28 @@ func symbolzMock(sources plugin.MappingSources, syms func(string, string) ([]byt
...
+	}
+	p.Comments = append(p.Comments,
+		"local="+strings.Join(args, ","))
 	return nil
 }
 
+func demangleMock(p *profile.Profile, force bool, mode string) {
+	if force {
+		p.Comments = append(p.Comments, "force")
```
aalexand: Same with "force-demangle"?

internal/symbolizer/symbolizer.go
rauls5382: Not sure I understand the question.

-symbolize=force will force both remote and local symbolization.
-symbolize=force:remote will force remote symbolization.

rauls5382: Oh I see what you mean. Yes, symbolz will always skip mappings that already have function names. 
I'll propagate the option and check there too, just for consistency.

internal/symbolizer/symbolizer_test.go
rauls5382: local=force-local seems redundant to me. Can we leave as is?

internal/symbolizer/symbolizer_test.go
rauls5382: Same

internal/symbolizer/symbolizer_test.go
aalexand: Was looking to make the string in tests a bit more readable (currently have to know that the first force is for symbolization and the second for demangling), but leaving as is SGTM, it's a minor thing.

rauls5382:
Many changes, PTAL

internal/symbolz/symbolz.go
```diff
@@ -39,9 +39,9 @@ var (
 // separated by '+') and returns the symbolz output in a string. It
```
aalexand: Update docs?

internal/symbolz/symbolz_test.go
```diff
@@ -59,33 +96,25 @@ func TestSymbolize(t *testing.T) {
...
-		t.Errorf("unexpected symbolization for %#x: %v", l.Address, l.Line)
-	}
+	return p
+}
 
-	for _, l := range p.Location[1:] {
-		if len(l.Line) != 1 {
-			t.Errorf("failed to symbolize %#x", l.Address)
-			continue
+func checkSymbolized(locs []*profile.Location, hasSymbols bool) error {
```
aalexand: subjectively, "wantSymbolized" might be a better name than "hasSymbols".

internal/symbolz/symbolz_test.go
```diff
@@ -41,12 +41,49 @@ func TestSymbolzURL(t *testing.T) {
...
+		for _, force := range []bool{false, true} {
+			p := testProfile(hasFunctions)
+
+			if err := Symbolize(p, force, s, fetchSymbols, &proftest.TestUI{T: t}); err != nil {
+				t.Errorf("symbolz: %v", err)
+				continue
+			}
+			var wantSym, wantNoSym []*profile.Location
+			if force || !hasFunctions {
+				wantNoSym = p.Location[1:1]
```
aalexand: Empty slice looks suspicious, [1:1] -> [0:1]?

internal/symbolz/symbolz.go
```diff
@@ -39,9 +39,9 @@ var (
...
 // symbolizes all locations based on their addresses, regardless of
 // mapping.
-func Symbolize(sources plugin.MappingSources, syms func(string, string) ([]byte, error), p *profile.Profile, ui plugin.UI) error {
+func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, syms func(string, string) ([]byte, error), ui plugin.UI) error {
 	for _, m := range p.Mapping {
-		if m.HasFunctions {
+		if m.HasFunctions && !force {
```
aalexand: newMapping() in symbolizer.go does this check as ```if !force && (m.HasFunctions || m.HasFilenames || m.HasLineNumbers) {```, should these be consistent?

rauls5382: No, because symbolz can only provide functions. If we have functions but not filenames there is no point on redoing symbolz, while it might be worthwhile doing local symbolization.

internal/symbolz/symbolz.go
rauls5382: Done

internal/symbolz/symbolz_test.go
rauls5382: Ugh, My Go is getting rusty. Thanks!

internal/symbolz/symbolz_test.go
rauls5382: Done

internal/symbolz/symbolz.go
aalexand: Ah, got it. It's somewhat subtle, consider adding a comment.

Issue #153, opened 2017-07-07
tagfocus issues with size


ahh:
(again, silly zipped profile for github naming)

```
pprof -raw heapz
Samples:
objects/count space/bytes[dflt]
       7438   60936269: 1 2 3 4 5 6
                bytes:[8192] request:[8191]
        238   62466710: 7 8 9 10 5 6
                bytes:[262144] request:[262143]
       3341   54739221: 7 8 9 10 5 6
                bytes:[16384] request:[16383]
       2064   67634517: 1 2 3 4 5 6
                bytes:[32768] request:[32767]
      29710   60847108: 1 2 3 4 5 6
                bytes:[2048] request:[2047]
        396   51920894: 1 2 3 4 5 6
                bytes:[131072] request:[131071]
      63503   65027585: 7 8 9 10 5 6
                bytes:[1024] request:[1023]
       1105   72423082: 7 8 9 10 5 6
                bytes:[65536] request:[65535]
        108   56884975: 11 12 13 3 4 5 6
                bytes:[524288] request:[524287]
      15887   65075220: 7 8 9 10 5 6
                bytes:[4096] request:[4095]
```

Now let's try to filter to some of these. Each of the below queries should match exactly one entry and do (this is good)

```
pprof -traces -tagfocus=4096bytes heapz
pprof -traces -tagfocus=4095request heapz
pprof -traces -tagfocus=4kb heapz
pprof -traces -tagfocus=512kb heapz
pprof -traces -tagfocus=524288bytes: heapz
```

But this matches everything:
```
pprof -traces -tagfocus=512kb: heapz
```

Note that `524288 bytes = 512kb`.  But for some reason when we have a non-empty *range* that breaks. That's a flat bug.  `:1kb` seems to work, otoh, as does `1kb:4kb`.

A slightly different problem is that there's no easy way to specify request ranges without having to write out full numbers.  I'd be very happy if I could specify, say:

```-tagfocus=/request/256kb:512kb```

to mean "I want to see a tag, whose name matches "request", whose value is between 256kb and 512kb.

[heapz.zip](https://github.com/google/pprof/files/1131710/heapz.zip)

aalexand:
@nolanmar511  Is this issue still relevant?

nolanmar511:
It's still not resolved, but I think it should be fixed when/by adding units to tags.

aalexand:
@nolanmar511 Should we mark this issue as fixed now?

nolanmar511:
I think so.

PR #213, opened 2017-08-29
Add support for units for numeric tags

### LINKED ISSUE #181: tags *have* units, they aren't units
Related to #178: a problem with the current format is that a tag is just a label with a name and either a string or an int value...but the *name*, while often implying a unit, isn't actually a unit.  

It would be good if we could extend the profile.proto format to support tags having both a name *and* an associated unit scale.

nolanmar511:
Fixes #181, #153

Currently, a numeric tag's units are assumed to be given by the key of the tag. This is problematic, because we have numeric tags which represent different things but should have the same units (for example,  the actual size of memory allocations is noted with the tag "bytes" and requested size of memory allocations with the tag "request").

This change adds a unit field to Label in profile.proto and enables this field to be used by pprof. Generally, if a numeric tag does not have a unit specified, the unit will still be assumed to be the key. There are two exceptions to this: if the key is "request" or "alignment", the units will be assumed to be "bytes".


profile/legacy_profile.go
```diff
@@ -541,7 +541,7 @@ func parseHeap(b []byte) (p *Profile, err error) {
 		p.Sample = append(p.Sample, &Sample{
 			Value:    value,
 			Location: sloc,
-			NumLabel: map[string][]int64{"bytes": {blocksize}},
+			NumLabel: map[string]NumValues{"bytes": {Unit: "bytes", Values: []int64{blocksize}}},
```
rauls5382: Should we now call this "size" instead of bytes?

profile/encode.go
```diff
@@ -288,15 +292,31 @@ func (p *Profile) postDecode() error {
...
 			if l.strX != 0 {
 				value, err = getString(p.stringTable, &l.strX, err)
 				labels[key] = append(labels[key], value)
 			} else if l.numX != 0 {
-				numLabels[key] = append(numLabels[key], l.numX)
+				var unit string
+				if l.unitX != 0 {
+					unit, err = getString(p.stringTable, &l.unitX, err)
+				} else if key == "bytes" || key == "requests" || key == "alignment" {
+					unit = "bytes"
```
rauls5382: I don't think it is right for pprof to force "bytes" as the units when writing profiles. I think we should preserve whatever was in the incoming profile.

It is OK for pprof to internally guess the units based on heuristics, but I would prefer to not expose those outside pprof.

nolanmar511: I could see guessing the units everywhere that tag units are used getting a bit tricky, but I can look more into it if that seems like the best option.

Would it work to add a field to indicate that units have been guessed and then not write out the tag units if this field indicates that units were guessed?

profile/legacy_profile.go
nolanmar511: That would make sense to me. 
I did keep this as bytes for now to be consistent with other profiles being generated. 
Should I change this  to "size" anyways?

profile/encode.go
rauls5382: I'd like to at least explore how bad it is. I'd like to keep this as a self-contained hack that we can eventually easily remove.

In particular I worry that if somebody has in the future a profile with tags {key =requests, unit=count}, we'd override their unit to bytes incorrectly.

profile/legacy_profile.go
rauls5382: I'd like to avoid "bytes"/"bytes" as it makes little sense. "size"/"bytes", or just continue to generate "bytes"/"" as before might be preferable -- we will have to deal with profiles generated by handles with "bytes"/"" for the foreseeable future anyway.

profile/encode.go
nolanmar511: Right now, this won't ever override units that were already assigned -- just add units when the units are empty.  (I'll update the tests to capture this)
It is true, if someone specified the units as "", then this would infer incorrect units.

profile/legacy_profile.go
nolanmar511: I had briefly changed the key to "size", but then realized that graphs display information from the "bytes" tag. So, while I agree "size""bytes" would make more sense, I think making a change to this tag name might require thinking a bit more about how to use numeric tags in graphs.

I also think that the units field should be set within pprof. Otherwise, it would be necessary to infer units in multiple places in the code, (I know it would be necessary at least within tag filtering and within some of the code for different graphs), which doesn't seem as maintainable. Would it be clear enough if code were modified so that units were not written out if the profile were saved?

rauls5382: Yes, bytes/bytes is confusing, and is being introduced here.

I'd rather either go to "size"/"bytes", or if we need more thinking, leave as "bytes"/"" for now. We need to continue supporting that combination for the foreseeable future anyway.

profile/encode.go
rauls5382: Yes, I think it is OK for pprof to infer incorrect units in some cases, but I'm weary of baking units derived from heuristics into profiles we generate.

internal/driver/driver.go
```diff
@@ -277,6 +281,70 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
 	return ropt, nil
 }
 
+// inferNumericLabelUnits returns a map of numeric label keys to the units
+// associated with those keys.
+// Unit for a given key is the first encountered unit for that key. If multiple
+// units are encountered for values paired with a particular key, then the first
+// unit encountered is used and an error message is displayed.
+// if units are encountered for a particular key, the unit is then inferred
```
aalexand: Nit: if -> If

internal/driver/driver.go
```diff
@@ -277,6 +281,70 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
+// based on the key.
+func inferNumericLabelUnits(p *profile.Profile, ui plugin.UI) map[string]string {
+	inferredNumericLabelUnits := map[string]string{}
```
aalexand: This and two other vars below - prefer using shorter names as these are local vars.

internal/driver/driver.go
```diff
@@ -277,6 +281,70 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
+	unusedUnits := map[string]map[string]bool{}
+	encounteredKeys := map[string]bool{}
+
+	// determine units based on numeric tags for each sample
+	for _, s := range p.Sample {
+		for key, vs := range s.NumLabel {
+			encounteredKeys[key] = true
+			for _, v := range vs {
+				unit := v.Unit
+				if unit != "" {
```
aalexand: Use
```
if unit == "" {
  continue
}
```
to save an indentation level?

internal/driver/driver.go
```diff
@@ -277,6 +281,70 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
 }
 
+// inferNumericLabelUnits returns a map of numeric label keys to the units
+// associated with those keys.
+// Unit for a given key is the first encountered unit for that key. If multiple
+// units are encountered for values paired with a particular key, then the first
+// unit encountered is used and an error message is displayed.
+// if units are encountered for a particular key, the unit is then inferred
+// based on the key.
+func inferNumericLabelUnits(p *profile.Profile, ui plugin.UI) map[string]string {
```
aalexand: "infer" is a bit misleading since we only infer the unit if it's not specified explicitly and actually use the unit from the profile when it's there?

internal/driver/driver.go
```diff
@@ -277,6 +281,70 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
+			}
+		}
+	}
+
+	// infer units for keys without any units associated with
+	// numeric tag values
+	for key := range encounteredKeys {
+		if _, ok := inferredNumericLabelUnits[key]; !ok {
+			switch key {
+			case "alignment":
```
aalexand:
```
case "alignment", "bytes", "requests":
   ...
```
You also don't need to list "bytes", it falls under the default case.

profile/profile.go
```diff
@@ -31,13 +31,14 @@ import (
...
-	Function          []*Function
-	Comments          []string
+	SampleType            []*ValueType
+	DefaultSampleType     string
+	Sample                []*Sample
+	Mapping               []*Mapping
+	Location              []*Location
+	Function              []*Function
+	Comments              []string
+	InferredNumLabelUnits map[string]string
```
aalexand: It is a bit strange that we attach this to the profile only using it in the reporting flow, not in the parsing. Can we pass the information about the inferred units separately from the profile?

profile/testdata/cppbench.growth.string
```diff
@@ -3,175 +3,175 @@ Period: 1
 Samples:
 objects/count space/bytes
           1    2097152: 1 2 3 4 5 6 7 8 9 10 11 12 13 
-                bytes:[2097152]
+                bytes:[{ 2097152}]
```
aalexand: Why there are the curly braces now? Also, why there is a space after the opening brace but not before the closing one?

nolanmar511: The curly brackets are because the value for the numeric label is now represented as an object, rather than an int. The space is because there is an empty string for the unit field that is printed out.

aalexand: Can we make it print exactly as before when the unit is not specified?

internal/driver/driver.go
```diff
@@ -277,6 +282,69 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
 }
 
+// getNumLabelUnits returns a map of numeric label keys to the units
+// associated with those keys.
+// Unit for a given key is the first encountered unit for that key. If multiple
+// units are encountered for values paired with a particular key, then the first
+// unit encountered is used and an error message is displayed.
+// If units are encountered for a particular key, the unit is then inferred
+// based on the key.
+func getNumLabelUnits(p *profile.Profile, ui plugin.UI) map[string]string {
```
aalexand: Nit: I think Go style typically does not use "get" in function names.

aalexand:
Looks like some comments are still not addressed, so I'll wait for a heads up when the change is ready to be reviewed again.

profile/legacy_profile.go
nolanmar511: I've modified this so it will be "bytes"/"" for now.

profile/encode.go
nolanmar511: I've modified this so inferred units will not modify the profile (if the profile is, for example, saved)

profile/profile.go
```diff
@@ -68,18 +68,27 @@ type Sample struct {
 	Location []*Location
 	Value    []int64
 	Label    map[string][]string
-	NumLabel map[string][]int64
+	NumLabel map[string][]NumValue
```
aalexand: Suggestion: this change technically breaks the API of Sample which is public. I don't know if this is going to break any existing clients, but could consider doing

```
NumLabel map[string][]int64
NumLabelUnit map[string][]string
```

instead. That is less memory efficient though so maybe save that only for the case there are clients we can't upgrade easily.

profile/profile_test.go
```diff
@@ -505,6 +549,60 @@ func TestMergeAll(t *testing.T) {
 	}
 }
 
+func TestNumLabelMerge(t *testing.T) {
+	for _, tc := range []struct {
+		Name         string
```
aalexand: Name -> name, Profs -> profs and so on, these don't need to be exported.

profile/profile_test.go
```diff
@@ -505,6 +549,60 @@ func TestMergeAll(t *testing.T) {
...
+		Profs        []*Profile
+		ExpNumLabels []map[string][]NumValue
```
aalexand: exp -> want (golang uses "got" and "want" speak, not "actual" and "expected")

profile/profile_test.go
```diff
@@ -505,6 +549,60 @@ func TestMergeAll(t *testing.T) {
...
+			}
+			for i, expLabels := range tc.ExpNumLabels {
+				numLabels := prof.Sample[i].NumLabel
+				if want, got := len(expLabels), len(numLabels); want != got {
+					t.Errorf("for sample %d, want %d numeric tags, got %d", i, want, got)
+					continue
+				}
+				for k, expVs := range expLabels {
+					if vs, ok := numLabels[k]; ok {
+						if len(vs) != len(expVs) {
```
aalexand: Would it be possible to just use reflect.DeepEqual?

profile/profile.go
```diff
@@ -455,85 +464,135 @@ func (p *Profile) String() string {
...
+	ls := []string{}
+	for k, v := range labels {
+		ls = append(ls, fmt.Sprintf("%s:%v", k, v))
+	}
+	sort.Strings(ls)
+	return strings.Join(ls, " ")
+}
+
+// numLablesToString returns a string representation of a map
+// representing numeric labels
```
aalexand: Full stop at the end.

profile/profile.go
```diff
@@ -68,18 +68,27 @@ type Sample struct {
...
 	Label    map[string][]string
-	NumLabel map[string][]int64
+	NumLabel map[string][]NumValue
 
 	locationIDX []uint64
 	labelX      []label
 }
 
+// NumValue used to store numeric value with unit for numeric values
+// of Profile.Label
```
aalexand: Full stop at the end.

profile/legacy_profile.go
```diff
@@ -541,7 +541,7 @@ func parseHeap(b []byte) (p *Profile, err error) {
 		p.Sample = append(p.Sample, &Sample{
 			Value:    value,
 			Location: sloc,
-			NumLabel: map[string][]int64{"bytes": {blocksize}},
+			NumLabel: map[string][]NumValue{"bytes": {{Unit: "", Value: blocksize}}},
```
aalexand: Remove `Unit: ""`, string fields are empty strings by default.

profile/legacy_java_profile.go
```diff
@@ -215,7 +215,7 @@ func parseJavaSamples(pType string, b []byte, p *Profile) ([]byte, map[uint64]*L
 				if s.Value[0] == 0 {
 					return nil, nil, fmt.Errorf("parsing sample %s: second value must be non-zero", line)
 				}
-				s.NumLabel = map[string][]int64{"bytes": {s.Value[1] / s.Value[0]}}
+				s.NumLabel = map[string][]NumValue{"bytes": {{Unit: "", Value: s.Value[1] / s.Value[0]}}}
```
aalexand: Remove `Unit: ""`, string fields are empty strings by default.

internal/report/report.go
```diff
@@ -759,7 +767,15 @@ func printTraces(w io.Writer, rpt *Report) error {
 		// Print any numeric labels for the sample
 		var numLabels []string
 		for k, v := range sample.NumLabel {
-			numLabels = append(numLabels, fmt.Sprintf("%10s:  %s\n", k, strings.Trim(fmt.Sprintf("%d", v), "[]")))
+			formatTag := func(vv int64, key string) string {
```
aalexand: Why is this local function needed?

internal/report/report.go
```diff
@@ -759,7 +767,15 @@ func printTraces(w io.Writer, rpt *Report) error {
...
+				return measurement.ScaledLabel(vv, key, o.OutputUnit)
```
aalexand: Using o.OutputUnit seems a bit weird. The output unit for the metric has nothing to do with the tag output units? There is only one output metric (set by sample_index), but there are multiple tags, so if anything I'd expect having a separate option like `-tag_output_unit=bytes:kb,alignment:bytes`. We don't need to add that flag now, but using o.OutputUnit appears strange to me.

internal/driver/driver.go
```diff
@@ -277,6 +282,69 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
+// associated with those keys.
+// Unit for a given key is the first encountered unit for that key. If multiple
+// units are encountered for values paired with a particular key, then the first
+// unit encountered is used and an error message is displayed.
+// If units are encountered for a particular key, the unit is then inferred
+// based on the key.
+func identifyNumLabelUnits(p *profile.Profile, ui plugin.UI) map[string]string {
+	numLabelUnits := map[string]string{}
+	unusedUnits := map[string]map[string]bool{}
+	encounteredKeys := map[string]bool{}
```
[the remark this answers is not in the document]
aalexand: I mean dropping encounteredKeys.

internal/driver/driver.go
```diff
@@ -277,6 +282,69 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
+	// determine units based on numeric tags for each sample
+	for _, s := range p.Sample {
+		for key, vs := range s.NumLabel {
+			encounteredKeys[key] = true
+			for _, v := range vs {
+				unit := v.Unit
+				if unit == "" {
+					continue
+				}
+				if expUnit, ok := numLabelUnits[key]; !ok {
```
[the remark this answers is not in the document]
aalexand: Or wantUnit

profile/profile.go
```diff
@@ -464,98 +464,135 @@ func (p *Profile) String() string {
...
-			bits = bits + "[FL]"
+		ss = append(ss, m.String())
+	}
+
+	return strings.Join(ss, "\n") + "\n"
+}
+
+// String dumps a text representation of a mapping. Intended mainly
+// for debugging purposes.
+func (m *Mapping) String() string {
```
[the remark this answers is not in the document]
nolanmar511: I figured that might be the case, but didn't really think through the consequences of that.
Should I change the function name to prevent any side effects?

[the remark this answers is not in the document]
aalexand: Yes, maybe just do it as `string()` for now.

internal/report/report.go
nolanmar511: I agree -- updated so tag units are auto-scaled

nolanmar511:
I believe I've addressed all comments -- PTAL

internal/driver/driver_test.go
```diff
@@ -1020,33 +1018,247 @@ func TestTagFilter(t *testing.T) {
 	}
 }
 
+func TestInferUnits(t *testing.T) {
```
aalexand: TestIdentifyNumLabelUnits

internal/driver/driver.go
```diff
@@ -277,6 +283,71 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
+	// a single key
+	for k, v := range unusedUnits {
+		units := make([]string, len(v))
+		i := 0
+		for unit := range v {
+			units[i] = unit
+			i++
+		}
+		sort.Strings(units)
+		ui.PrintErr("For tag ", k, " used unit ", numLabelUnits[k], " also encountered unit(s) ", strings.Join(units, ","))
```
aalexand: Use `ui.PrintErr(fmt.Sprintf(...` instead of string concatenation maybe? I think that would be easier to read.

internal/driver/driver.go
```diff
@@ -277,6 +283,71 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
+// identifyNumLabelUnits returns a map of numeric label keys to the units
+// associated with those keys.
+// Unit for a given key is the first encountered unit for that key. If multiple
+// units are encountered for values paired with a particular key, then the first
+// unit encountered is used and an error message is displayed.
+// If units are encountered for a particular key, the unit is then inferred
+// based on the key.
+func identifyNumLabelUnits(p *profile.Profile, ui plugin.UI) map[string]string {
+	numLabelUnits := map[string]string{}
+	unusedUnits := map[string]map[string]bool{}
```
aalexand: Maybe `ignoredUnits` would be a bit clearer.

internal/driver/driver.go
```diff
@@ -277,6 +283,71 @@ func reportOptions(p *profile.Profile, vars variables) (*report.Options, error)
...
 }
 
+// identifyNumLabelUnits returns a map of numeric label keys to the units
+// associated with those keys.
+// Unit for a given key is the first encountered unit for that key. If multiple
+// units are encountered for values paired with a particular key, then the first
+// unit encountered is used and an error message is displayed.
+// If units are encountered for a particular key, the unit is then inferred
+// based on the key.
+func identifyNumLabelUnits(p *profile.Profile, ui plugin.UI) map[string]string {
```
aalexand: Looking more at this code, it seems that we should have this as a public method on Profile so that clients of pprof API can use it to get the mapping of tag keys to their units. We discussed before that we want to print the warnings / errors about the ignored units for the same tag key, but maybe we can do something like

```
func (p *Profile) NumLabelUnits() (map[string]string, map[string][]string) {
  ...
}
```

where the function returns the detected units per tag key and also the list of ignored duplicated units per tag key. The implementation here then would call that function and report the ignored units to the plugin UI.

nolanmar511:
PTAL

nolanmar511:
PTAL

internal/driver/driver_test.go
```diff
@@ -1020,34 +1018,221 @@ func TestTagFilter(t *testing.T) {
...
 }
 
+func TestIdentifyNumLabelUnits(t *testing.T) {
+	var tagFilterTests = []struct {
+		name               string
+		tagVals            []map[string][]int64
+		tagUnits           []map[string][]string
+		wantUnits          map[string]string
+		ignoreRX           string
+		wantIgnoreErrCount int
```
aalexand: Nit: *Ignore* -> *Ignored*

internal/driver/driver_test.go
```diff
@@ -1020,34 +1018,221 @@ func TestTagFilter(t *testing.T) {
 	}
 }
 
+func TestIdentifyNumLabelUnits(t *testing.T) {
+	var tagFilterTests = []struct {
+		name               string
```
aalexand: This is typically called "desc" and should be more descriptive than "test1", "test2" and so on. Describe what the actual test case the test covers.

internal/driver/driver_test.go
```diff
@@ -1020,34 +1018,221 @@ func TestTagFilter(t *testing.T) {
...
+				NumUnit:  test.tagUnits[i],
+			}
+			p.Sample[i] = &s
+		}
+		testUI := &proftest.TestUI{T: t, IgnoreRx: test.ignoreRX}
+		units := identifyNumLabelUnits(&p, testUI)
+		for key, wantUnit := range test.wantUnits {
+			unit := units[key]
+			if wantUnit != unit {
+				t.Errorf("%s: for key %s, got unit %s, want unit %s", test.name, key, unit, wantUnit)
```
aalexand: Since you got the test.name field (which should be test.desc, see above), I'd recommend using t.Run() to execute the test cases, see https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks.

internal/driver/driver_test.go
```diff
@@ -1020,34 +1018,221 @@ func TestTagFilter(t *testing.T) {
...
+		}
+	}
+}
+
 func TestNumericTagFilter(t *testing.T) {
 	var tagFilterTests = []struct {
-		name, value string
-		tags        map[string][]int64
-		want        bool
+		name, value   string
```
aalexand: desc

internal/driver/driver_test.go
```diff
@@ -1020,34 +1018,221 @@ func TestTagFilter(t *testing.T) {
...
+		tags          map[string][]int64
+		inferredUnits map[string]string
```
aalexand: Nit: identifiedUnits

internal/proftest/proftest.go
```diff
@@ -74,9 +74,10 @@ func EncodeJSON(x interface{}) []byte {
 // TestUI implements the plugin.UI interface, triggering test failures
 // if more than Ignore errors not matching IgnoreRx are printed.
 type TestUI struct {
-	T        *testing.T
-	Ignore   int
-	IgnoreRx string
+	T               *testing.T
+	Ignore          int
+	IgnoreRx        string
+	IgnoredErrCount int
```
aalexand: Maybe just IgnoreCount to be consistent as we don't include the "Err" string in Ignore and IgnoreRx names.

internal/proftest/proftest.go
```diff
@@ -74,9 +74,10 @@ func EncodeJSON(x interface{}) []byte {
 // TestUI implements the plugin.UI interface, triggering test failures
 // if more than Ignore errors not matching IgnoreRx are printed.
```
aalexand: Update the docs wrt IgnoredCount.

internal/proftest/proftest.go
```diff
@@ -92,14 +93,16 @@ func (ui *TestUI) Print(args ...interface{}) {
...
+		if matched, err := regexp.MatchString(ui.IgnoreRx, fmt.Sprint(args...)); matched || err != nil {
 			if err != nil {
 				ui.T.Errorf("failed to match against regex %q: %v", ui.IgnoreRx, err)
 			}
+			ui.IgnoredErrCount++
 			return
 		}
 	}
 	if ui.Ignore > 0 {
+		ui.IgnoredErrCount++
```
aalexand: I would only expect the counter to be increased when the regex is matched. I think there is a naming issue. Perhaps IgnoreRx should be AllowRx and IgnoredErrCount should be NumAllowRxMatches or something like that.

internal/report/report.go
```diff
@@ -241,15 +242,25 @@ func (rpt *Report) newGraph(nodes graph.NodeSet) *graph.Graph {
 	for _, f := range prof.Function {
 		f.Filename = trimPath(f.Filename)
 	}
-	// Remove numeric tags not recognized by pprof.
+	// Remove numeric tags not recognized by pprof label with inferred units.
```
aalexand: This "not recognized by pprof label with inferred units." is not quite clear. What is "pprof label"? Can you rephrase?

internal/report/report.go
```diff
@@ -759,7 +771,12 @@ func printTraces(w io.Writer, rpt *Report) error {
 		// Print any numeric labels for the sample
 		var numLabels []string
 		for k, v := range sample.NumLabel {
```
aalexand: Nit: "key, vals" like in line 645?

aalexand: Or change to "k, v" there.

profile/encode.go
```diff
@@ -289,21 +296,35 @@ func (p *Profile) postDecode() error {
...
 			}
 		}
 		if len(labels) > 0 {
 			s.Label = labels
 		}
 		if len(numLabels) > 0 {
 			s.NumLabel = numLabels
+			for k := range numUnits {
+				if !unitEncountered[k] {
+					delete(numUnits, k)
```
aalexand: Maybe we can optimize a bit for the case of no tag units by just skipping append if the unit is an empty string and catching up for the desired length when we do need to append a non-empty unit.

nolanmar511:
PTAL

internal/report/report.go
```diff
@@ -241,15 +242,27 @@ func (rpt *Report) newGraph(nodes graph.NodeSet) *graph.Graph {
 	for _, f := range prof.Function {
 		f.Filename = trimPath(f.Filename)
 	}
-	// Remove numeric tags not recognized by pprof.
+	// Removes all numeric tags except for the bytes tag prior
+	// to making graph.
+	// TODO : modify to select first numeric tag if no bytes tag
```
aalexand: Nit: remove the space before the colon make the text a proper sentence. I.e. "TODO: Do something."

profile/encode.go
```diff
@@ -289,13 +296,24 @@ func (p *Profile) postDecode() error {
...
 			if l.strX != 0 {
 				value, err = getString(p.stringTable, &l.strX, err)
 				labels[key] = append(labels[key], value)
 			} else if l.numX != 0 {
+				numValues := numLabels[key]
+				units := numUnits[key]
+				if l.unitX != 0 {
+					var unit string
+					unit, err = getString(p.stringTable, &l.unitX, err)
+					if len(units) < len(numValues) {
```
aalexand: I'd move the condition into padStringArray, it would simplify the calling code.

profile/encode.go
```diff
@@ -340,6 +364,17 @@ func (p *Profile) postDecode() error {
 	return err
 }
 
+// padStringArray pads arr with enough empty strings to make arr
+// length l when arr's length is less than l.
+func padStringArray(arr []string, l int) []string {
```
aalexand: Nit: I'd simplify to
```
func padStringArray(arr []string, l int) []string {
  if l <= len(arr) {
    return arr
  }
  return append(arr, make([]string, l - len(arr))
}
```

profile/profile.go
```diff
@@ -426,6 +429,75 @@ func (p *Profile) Aggregate(inlineFrame, function, filename, linenumber, address
...
+			default:
+				numLabelUnits[key] = key
+			}
+		}
+	}
+
+	// Copy ignored units into more readable format
+	unitsIgnored := make(map[string][]string, len(ignoredUnits))
+	for key, values := range ignoredUnits {
+		fmt.Printf("%v", values)
```
aalexand: Is this a debugging leftover?

profile/profile_test.go
```diff
@@ -505,6 +570,63 @@ func TestMergeAll(t *testing.T) {
...
+				},
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			prof, err := Merge(tc.profs)
+			if err != nil {
+				t.Errorf("merge error: %v", err)
+			}
+			fmt.Printf("%v", prof)
```
aalexand: Is this a debugging leftover?

profile/profile_test.go
```diff
@@ -688,6 +810,123 @@ func locationHash(s *Sample) string {
...
+func TestInferUnits(t *testing.T) {
+	var tagFilterTests = []struct {
+		name             string
+		tagVals          []map[string][]int64
+		tagUnits         []map[string][]string
+		wantUnits        map[string]string
+		wantIgnoredUnits map[string][]string
+	}{
+		{
+			"test1",
```
aalexand: Name the tests by the actual thing they test? "test1", "test2" and so on is not going to be useful in a couple months.

profile/profile_test.go
```diff
@@ -688,6 +810,123 @@ func locationHash(s *Sample) string {
...
+		}
+		for key, ignored := range ignoredUnits {
+			wantIgnored := test.wantIgnoredUnits[key]
+			if len(wantIgnored) != len(ignored) {
+				t.Errorf("%s: for key %s, want ignored units %v, got ignored units %v", test.name, key, wantIgnored, ignored)
+				continue
+			}
+			for i, want := range wantIgnored {
+				got := ignored[i]
+				if want != got {
```
aalexand: Nit: merge the lines as "if got := ignored[i]; want != got {"

profile/profile_test.go
```diff
@@ -505,6 +570,63 @@ func TestMergeAll(t *testing.T) {
...
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			prof, err := Merge(tc.profs)
+			if err != nil {
+				t.Errorf("merge error: %v", err)
+			}
+			fmt.Printf("%v", prof)
+			if want, got := len(tc.wantNumLabels), len(prof.Sample); want != got {
+				t.Fatalf("want %d samples, got %d samples", want, got)
```
aalexand: "got" should go first.

profile/profile_test.go
```diff
@@ -505,6 +570,63 @@ func TestMergeAll(t *testing.T) {
...
+			}
+			for i, wantLabels := range tc.wantNumLabels {
+				numLabels := prof.Sample[i].NumLabel
+				if !reflect.DeepEqual(wantLabels, numLabels) {
+					t.Errorf("want numeric labels %v, got %v", wantLabels, numLabels)
```
aalexand: "got" should go first.

profile/profile_test.go
```diff
@@ -505,6 +570,63 @@ func TestMergeAll(t *testing.T) {
...
+				}
+
+				wantUnits := tc.wantNumUnits[i]
+				numUnits := prof.Sample[i].NumUnit
+				if !reflect.DeepEqual(wantUnits, numUnits) {
+					t.Errorf("want numeric labels %v, got %v", wantUnits, numUnits)
```
aalexand: "got" should go first.

aalexand:
Please also rebase onto the current master.

nolanmar511:
PTAL

profile/profile_test.go
```diff
@@ -689,6 +811,122 @@ func locationHash(s *Sample) string {
...
+		},
+		{
+			"One sample, one key with one value, unit specified",
+			[]map[string][]int64{{"key1": {8}}},
+			[]map[string][]string{{"key1": {"bytes"}}},
+			map[string]string{"key1": "bytes"},
+			map[string][]string{},
+		},
+		{
+			"bytes key, unit not specified",
```
aalexand: Nit: you start the test name with lowercase here, but with uppercase for the two other cases above, let's be consistent and use the uppercase here and in others.

profile/profile_test.go
```diff
@@ -689,6 +811,122 @@ func locationHash(s *Sample) string {
...
+				NumLabel: numLabel,
+				NumUnit:  test.tagUnits[i],
+			}
+			p.Sample[i] = &s
+		}
+		units, ignoredUnits := p.NumLabelUnits()
+		for key, wantUnit := range test.wantUnits {
+			unit := units[key]
+			if wantUnit != unit {
+				t.Errorf("%s: for key %s, want unit %s, got unit %s", test.name, key, wantUnit, unit)
```
aalexand: The "got" part should go first here. There are two other places like this to fix in the new code below.

profile/encode.go
```diff
@@ -304,6 +320,12 @@ func (p *Profile) postDecode() error {
 		}
 		if len(numLabels) > 0 {
 			s.NumLabel = numLabels
+			for key, units := range numUnits {
+				if lenUnits, lenLabels := len(units), len(numLabels[key]); lenUnits > 0 && lenUnits < lenLabels {
```
aalexand: You don't need the "&& lenUnits < lenLabels" part I think since the condition is checked in padStringArray anyway.

internal/report/report.go
```diff
@@ -241,15 +242,27 @@ func (rpt *Report) newGraph(nodes graph.NodeSet) *graph.Graph {
 	for _, f := range prof.Function {
 		f.Filename = trimPath(f.Filename)
 	}
-	// Remove numeric tags not recognized by pprof.
+	// Removes all numeric tags except for the bytes tag prior
+	// to making graph.
+	// TODO: modify to select first numeric tag if no bytes tag
```
aalexand: I'd suggest filing an issue for this.

nolanmar511:
PTAL

internal/report/report.go
aalexand: Ping.

profile/encode.go
```diff
@@ -304,6 +320,12 @@ func (p *Profile) postDecode() error {
 		}
 		if len(numLabels) > 0 {
 			s.NumLabel = numLabels
+			for key, units := range numUnits {
+				if lenUnits, lenLabels := len(units), len(numLabels[key]); lenUnits > 0 {
```
aalexand: You don't need the temporary variables because you only use the value once. It should just be
```
if len(units) > 0 {
  numUnits[key] = padStringArray(units, len(numLabels[key]))
}
```

profile/profile.go
```diff
@@ -435,6 +438,74 @@ func (p *Profile) Aggregate(inlineFrame, function, filename, linenumber, address
...
+						ignoredUnits[k] = map[string]bool{unit: true}
+					}
+				}
+			}
+		}
+	}
+
+	// Infer units for keys without any units associated with
+	// numeric tag values.
+	for key := range numLabelUnits {
```
aalexand: This can be "for key, unit := range numLabelUnits {" I think and you don't need the lookup by key in the next condition line.

internal/graph/graph.go
```diff
@@ -633,9 +633,15 @@ func (n *Node) addSample(dw, w int64, labels string, numLabel map[string][]int64
...
 		format = defaultLabelFormat
 	}
-	for key, nvals := range numLabel {
-		for _, v := range nvals {
-			t := numericTags.findOrAddTag(format(v, key), key, v)
+	for k, nvals := range numLabel {
+		units := numUnit[k]
+		for i, v := range nvals {
+			var t *Tag
+			if len(units) == len(nvals) {
```
aalexand: It's a bit weird that to detect the presence of units we still check the equality of the len of the arrays here while in other places we check for zero like we discussed earlier. Should we be consistent?

profile/merge.go
```diff
@@ -200,7 +205,8 @@ func (sample *Sample) key() sampleKey {
 
 	numlabels := make([]string, 0, len(sample.NumLabel))
 	for k, v := range sample.NumLabel {
-		numlabels = append(numlabels, fmt.Sprintf("%q%x", k, v))
+		u := sample.NumUnit[k]
```
aalexand: Nit: I'd get rid of this local. It's used only once so meh.

aalexand:
Why is there "cla: no" label on the PR now?

nolanmar511:
PTAL

Issue #193, opened 2017-08-18
Ability to specify a comment to record in the saved profile


aalexand:
It is sometimes desirable to be able to associate a custom comment with a profile. There is "comment" field in the profile proto, so it would be good to have a `-comment` flag that can be used like:

```
pprof.go -comment "My comment" localhost:8080
```
or
```
pprof.go -comment "My comment" -proto -output new-profile.pb.gz profile.pb.gz
```

In the former case of remote collection, pprof stores the remote profile in a local file printing its path, the comment should become a part of that profile. Storing the address of the remote source as a comment for the remote collections would also be good.

PR #215, opened 2017-09-06
Add a flag to append a comment to a profile

lvng:
Here's a diff for discussion. The change is simple enough, but I'm not sure how to properly implement the test for it.

The current test fails, but is a starting point for discussion.

### DIFF
--- a/internal/driver/driver_test.go
+++ b/internal/driver/driver_test.go
@@ -523,7 +523,7 @@ func testFetchSymbols(source, post string) ([]byte, error) {
 type testSymbolzSymbolizer struct{}
 
 func (testSymbolzSymbolizer) Symbolize(variables string, sources plugin.MappingSources, p *profile.Profile) error {
-	return symbolz.Symbolize(sources, testFetchSymbols, p, nil)
+	return symbolz.Symbolize(p, false, sources, testFetchSymbols, nil)
 }
 
 func fakeDemangler(name string) string {
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -42,21 +42,26 @@ type Symbolizer struct {
 // test taps for dependency injection
 var symbolzSymbolize = symbolz.Symbolize
 var localSymbolize = doLocalSymbolize
+var demangleFunction = Demangle
 
 // Symbolize attempts to symbolize profile p. First uses binutils on
 // local binaries; if the source is a URL it attempts to get any
 // missed entries using symbolz.
 func (s *Symbolizer) Symbolize(mode string, sources plugin.MappingSources, p *profile.Profile) error {
-	remote, local, force, demanglerMode := true, true, false, ""
+	remote, local, fast, force, demanglerMode := true, true, false, false, ""
 	for _, o := range strings.Split(strings.ToLower(mode), ":") {
 		switch o {
+		case "":
+			continue
 		case "none", "no":
 			return nil
-		case "local", "fastlocal":
+		case "local":
 			remote, local = false, true
+		case "fastlocal":
+			remote, local, fast = false, true, true
 		case "remote":
 			remote, local = true, false
-		case "", "force":
+		case "force":
 			force = true
 		default:
 			switch d := strings.TrimPrefix(o, "demangle="); d {
@@ -75,17 +80,17 @@ func (s *Symbolizer) Symbolize(mode string, sources plugin.MappingSources, p *pr
 	var err error
 	if local {
 		// Symbolize locally using binutils.
-		if err = localSymbolize(mode, p, s.Obj, s.UI); err != nil {
+		if err = localSymbolize(p, fast, force, s.Obj, s.UI); err != nil {
 			s.UI.PrintErr("local symbolization: " + err.Error())
 		}
 	}
 	if remote {
-		if err = symbolzSymbolize(sources, postURL, p, s.UI); err != nil {
+		if err = symbolzSymbolize(p, force, sources, postURL, s.UI); err != nil {
 			return err // Ran out of options.
 		}
 	}
 
-	Demangle(p, force, demanglerMode)
+	demangleFunction(p, force, demanglerMode)
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
+		t.Errorf("cannot create tempfile: %v", err)
+	}
+	defer os.Remove(protoTempFile.Name())
+	defer protoTempFile.Close()
+
+	testFlagSet := baseFlags()
+	testFlagSet.bools["proto"] = true
+	testFlagSet.strings["comment"] = "test comment please ignore"
+	testFlagSet.strings["output"] = protoTempFile.Name()
+	testFlagSet.args = []string{"cpu"}
```
nolanmar511: Consider using "cpusmall" instead of "cpu" since profile content isn't super important here, and it might be easier to debug this test case with a smaller profile?

internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
+	if string(expectedBuf.Bytes()) != string(actualBytes) {
+		actualProfile, err := profile.ParseData(actualBytes)
+		if err != nil {
+
+		}
+
+		ajs := proftest.EncodeJSON(&actualProfile)
+		ejs := proftest.EncodeJSON(&expectedProfile)
+
+		t.Errorf("output doesnt match \n%v\n%v", string(ejs), string(ajs))
```
nolanmar511: To make this easier to debug, maybe use` proftest.diff(ajs,ejs)` like in profile/proto_test.go

aalexand:
* Please name the pull request in a clearer way. Capitalize the first word.
* Reference #193 in the pull request description.
* Make the submission checks pass.

internal/driver/commands.go
```diff
@@ -131,6 +131,8 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
+	"comment": &variable{stringKind, "", "", helpText("Append comment to output")},
```
aalexand: The help text is somewhat unclear.

internal/driver/driver_test.go
aalexand: typo

internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
+	expectedProfile.Comments = append(expectedProfile.Comments, "test comment please ignore")
+	expectedBuf := bytes.NewBuffer(nil)
+	err = expectedProfile.Write(expectedBuf)
+	if err != nil {
+		t.Errorf("error in Profile.Write %v", err)
+	}
+
+	if string(expectedBuf.Bytes()) != string(actualBytes) {
+		actualProfile, err := profile.ParseData(actualBytes)
+		if err != nil {
```
aalexand: 'then' clause is empty

internal/driver/fetch.go
```diff
@@ -83,6 +83,10 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
 	p.RemoveUninteresting()
 	unsourceMappings(p)
 
+	if len(s.Comment) > 0 {
```
aalexand: `if s.Comment != ""`

internal/driver/cli.go
```diff
@@ -34,6 +34,9 @@ type source struct {
 	Timeout      int
 	Symbolize    string
 	HTTPHostport string
+
+	// Comment will be appended to any output format that supports comments.
```
aalexand: Can you list these output formats?

internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
+	options := setDefaults(nil)
+	options.Flagset = testFlagSet
+	options.Fetch = testFetcher{}
+	options.Sym = testSymbolizer{}
+
+	if err := PProf(options); err != nil {
+		t.Errorf("error in PProf %v", err)
+	}
+
+	actualBytes, err := ioutil.ReadFile(protoTempFile.Name())
```
aalexand: actual -> got

internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("error in ReadFile %v", err)
+	}
+
+	expectedProfile := cpuProfile()
```
aalexand: expected -> want

internal/driver/cli.go
```diff
@@ -62,6 +65,8 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flagTimeout := flag.Int("timeout", -1, "Timeout in seconds for fetching a profile")
 	flagHTTP := flag.String("http", "", "Present interactive web based UI at the specified http host:port")
 
+	comment := flag.String("comment", "", "Append comment to output")
```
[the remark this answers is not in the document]
lvng: done

internal/driver/driver_test.go
lvng: done

internal/driver/cli.go
lvng: good question. i'm not sure I can, and i dont know if thats worthwhile. the comment gets appended to the profile before it gets processed by any of the other commands, and its up to them whether they do anything with comments. i assume most of them ignore comments, except for proto, raw and comments.

perhaps it would be better to say something "append comment to profile before processing to output format?"

internal/driver/commands.go
lvng: same response as above.

internal/driver/driver_test.go
lvng: any reason to prefer want/got vs expected/actual? most testing frameworks i'm familiar with use expected/actual.

internal/driver/driver_test.go
lvng: this was temporary. it is now removed.

internal/driver/driver_test.go
lvng: done.

lvng: done

internal/driver/fetch.go
lvng: is there a reason to prefer equality over length?

aalexand: It makes it clear that s.Comment is a string, not a slice.

PR #217, opened 2017-09-11
Add flag to append a comment to a profile

### LINKED ISSUE #193: Ability to specify a comment to record in the saved profile

lvng:
resolve #193

### DIFF
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -131,6 +131,8 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
+	"comment": &variable{stringKind, "", "", helpText("Append comment to profile before processing to output format")},
+
 	// Comparisons.
 	"drop_negative": &variable{boolKind, "f", "", helpText(
 		"Ignore negative differences",
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -131,7 +131,11 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
-	"comment": &variable{stringKind, "", "", helpText("Append comment to profile before processing to output format")},
+	"comment": &variable{stringKind, "", "", helpText(
+		"Append comment to profile",
+		"The comment is appended to the profile on import, and will be ",
+		"included in any output format that includes comments, for example, ",
+		"'proto'.")},
 
 	// Comparisons.
 	"drop_negative": &variable{boolKind, "f", "", helpText(
--- a/internal/driver/driver_test.go
+++ b/internal/driver/driver_test.go
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
 	}
 }
 
+// TestAppendComment tests that the flag '-comment foo' will append a comment to the profile.
+func TestAppendComment(t *testing.T) {
+
+	savedVariables := pprofVariables.makeCopy()
+	defer func() { pprofVariables = savedVariables }()
+
+	tempFile, err := ioutil.TempFile("", "profile_proto")
+	if err != nil {
+		t.Fatalf("Error in TempFile: %v", err)
+	}
+	defer os.Remove(tempFile.Name())
+	defer tempFile.Close()
+
+	testFlagSet := baseFlags()
+	testFlagSet.bools["proto"] = true
+	testFlagSet.strings["comment"] = "test comment please ignore"
+	testFlagSet.strings["output"] = tempFile.Name()
+	testFlagSet.args = []string{"cpusmall"}
+
+	options := setDefaults(nil)
+	options.Flagset = testFlagSet
+	options.Fetch = testFetcher{}
+	options.Sym = testSymbolizer{}
+
+	if err := PProf(options); err != nil {
+		t.Fatalf("error in PProf %v", err)
+	}
+
+	gotBytes, err := ioutil.ReadFile(tempFile.Name())
+	if err != nil {
+		t.Fatalf("error in ReadFile %v", err)
+	}
+
+	wantProfile := cpuProfileSmall()
+	wantProfile.Comments = append(wantProfile.Comments, "test comment please ignore")
+	wantBuf := bytes.NewBuffer(nil)
+	err = wantProfile.Write(wantBuf)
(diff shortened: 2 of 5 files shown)


nolanmar511:
Could you reference issue #193 in the description, not the closed PR?

lvng:
oops! didnt realize they were different. fixed.

internal/driver/driver_test.go
```diff
@@ -633,27 +684,27 @@ func cpuProfileSmall() *profile.Profile {
 
 	var cpuL = []*profile.Location{
 		{
-			ID:      1000,
+			ID:      1,
```
aalexand: Why did the ID changed?

internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
 
+// TestAppendComment tests that the flag '-comment foo' will append a comment to the profile.
+func TestAppendComment(t *testing.T) {
+
+	savedVariables := pprofVariables.makeCopy()
+	defer func() { pprofVariables = savedVariables }()
+
+	tempFile, err := ioutil.TempFile("", "profile_proto")
+	if err != nil {
+		t.Fatalf("Error in TempFile: %v", err)
```
aalexand: Nit: the most used wording for test failures is `t.Fatalf("failed to do something: %v", err)` or `t.Fatalf("ioutil.TempFile(): got error %v, want no error", err)` -- consider using one of these.

internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
+	err = wantProfile.Write(wantBuf)
+	if err != nil {
+		t.Fatalf("error in Profile.Write %v", err)
+	}
+	wantBytes := wantBuf.Bytes()
+
+	if string(wantBytes) != string(gotBytes) {
+		diff, err := proftest.Diff(wantBytes, gotBytes)
+		if err != nil {
+			t.Fatalf("Error in Diff: %v", err)
```
aalexand: Nit: "Error" here but "error" in other messages above, please be consistent. Better use one of the wordings I mentioned above.

internal/driver/driver_test.go
```diff
@@ -203,6 +203,57 @@ func TestParse(t *testing.T) {
...
+		}
+		t.Errorf("output not expected\n%s\n%s\n%s", wantBytes, gotBytes, diff)
```
aalexand: Does this print binary proto bytes to the console? If so, it is probably not very useful.

internal/driver/commands.go
```diff
@@ -131,6 +131,8 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
+	"comment": &variable{stringKind, "", "", helpText("Append comment to profile before processing to output format")},
```
aalexand: The only output format we mean here is I think profile proto... At least that's the only thing we test. So maybe something like `String to record as a comment for profile proto output`?

internal/driver/driver_test.go
lvng: when the profile gets serialized, it generates a fresh set of ID's starting from 1. So when it compares the saved profile to the profile created in this function, the id's would be different and the test would fail.

aalexand: I think we should update the IDs before we compare then. We should keep testing that the profiles with sparse IDs are read well imo.

lvng: I did some investigating and found that the ID's get regenerated in the fetching step, where it merges profiles from multiple sources. I changed the code to fetch the expected profile as well, so that both profiles have normalized IDs and we dont have to change the IDs of the test profiles.

internal/driver/driver_test.go
lvng: change these to "function(): got error %v"

internal/driver/driver_test.go
lvng: fixed by above change.

internal/driver/driver_test.go
lvng: changed this to check if the encoded profiles are equal, and if not, decode the profiles, serialize them to json, and log a diff. Now the errors are much more useful.

internal/driver/driver_test.go
aalexand: These are not changed yet? Let me know with a comment like "Made all changes, the PR is ready to be reviewed again" when you make all the changes. Holding off the review for now.

internal/driver/commands.go
```diff
@@ -131,6 +131,12 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
+	"comment": &variable{stringKind, "", "", helpText(
+		"Append comment to profile",
+		"The comment is appended to the profile on import, and will be ",
+		"included in any output format that includes comments, for example, ",
```
aalexand: You don't need the trailing space, see the source of helpText().

internal/driver/driver_test.go
```diff
@@ -203,6 +203,92 @@ func TestParse(t *testing.T) {
...
+// If the test fails, the profiles are decoded, serialized to JSON, and the
+// diff is logged.
+func assertProfilesEqual(t *testing.T, wantBytes []byte, gotBytes []byte) {
+	if string(wantBytes) != string(gotBytes) {
+
+		wantProfile, err := profile.Parse(bytes.NewBuffer(wantBytes))
+		if err != nil {
+			t.Fatalf("Profile.Parse(): got error %v", err)
+		}
+		wantJSON := proftest.EncodeJSON(&wantProfile)
```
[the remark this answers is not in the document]
aalexand: In fact, we can just use existing TestParse(), we don't need much new code. I'll prepare a PR to demo.

aalexand:
@lvng, see https://github.com/google/pprof/pull/221, it does the testing in a simpler way.

PR #219, opened 2017-09-12
Change menu buttons to links.

ghemawat:
This allows right-click/middle-click etc. to work properly.

This one is only html/css/javascript changes, so presumably
@josef-jelinek is the best reviewer.

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -479,7 +479,8 @@ function viewer(baseUrl, nodes) {
 
   function handleKey(e) {
     if (e.keyCode != 13) return
-    handleFocus()
+    window.location.href =
+        updateUrl(new URL({{.BaseURL}}, window.location.href), "f")
     e.preventDefault()
   }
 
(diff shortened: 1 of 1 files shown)


internal/driver/webhtml.go
```diff
@@ -607,36 +602,58 @@ function viewer(baseUrl, nodes) {
...
       }
-      params.set(param, re)
+
+      elem.href = url.toString()
     }
 
-    window.location.href = url.toString()
+    // We update on mouseenter so middle-click/right-click work properly.
+    for (const e of ["mouseenter", "touchstart"]) {
+      elem.addEventListener(e, updater)
```
josef-jelinek: It would be easier to follow if this loop is above the updater() definition (JS defines all function statements before executing the surrounding code) so the code is together.
Also this loop unrolled to two lines would be both shorter and easier to read.

internal/driver/webhtml.go
```diff
@@ -607,36 +602,58 @@ function viewer(baseUrl, nodes) {
...
+    // updater does the actual work.
+    function updater() {
+      const url = new URL(elem.href)
+      url.hash = ""
+
+      // The selection can be in one of two modes: regexp-based or
+      // list-based.  Construct regular expression depending on mode.
+      let re = regexpActive ? search.value : ""
+      if (!regexpActive) {
+        selected.forEach(function(v, key) {
```
josef-jelinek: a more functional/compact way to write this section would be:
```
let re = regexpActive
    ? search.value
    : selected.keys().map(key => quotemeta(nodes[key]).join("|")
```

internal/driver/webhtml.go
```diff
@@ -690,6 +707,11 @@ function viewer(baseUrl, nodes) {
     toptable.addEventListener("touchstart", handleTopClick)
   }
 
+  for (const id of ["topbtn", "graphbtn", "peek", "list", "disasm",
```
josef-jelinek: at this point it seems that these arrays (also e.g. line 690) would benefit from being extracted to a constant as they may be tedious to update consistently.
There is a nicer way to make this loop as
`ids.forEach(makeLinkDynamic)`.

There is a potential JS trap here if written as `["focus", "ignore", "hide", "show"].forEach(makeLinkDynamic)` it may fail if the code before is an expression, since it will parse it as indexing of the result value instead of an array literal.
In that case for the semicolon-free JS a semicolon would be needed after all.

internal/driver/webhtml.go
ghemawat: Doesn't work:
"caught TypeError: selected.keys(...).map is not a function"

I think map() does not work Iterator, which is what selected.keys() returns:
https://stackoverflow.com/questions/43885365/using-map-on-an-iterator

I am inclined to stick with forEach; seems clear enough, even if it is not as compact.

internal/driver/webhtml.go
ghemawat: Done.

I also noticed a stale call to handleFocus=>navigate (when user hits Enter), which I changed to use the new updater code instead.

internal/driver/webhtml.go
ghemawat: Done.

internal/driver/webhtml.go
josef-jelinek: Oh, right, I meant Array.from(selected.keys()), but mentally filtered that out... It is a suggestion that does not need to deal with two conditions and special treatment of the first item...

ghemawat: Yes "Array.from" works. Done.

PR #220, opened 2017-09-12
Make top table sortable by column

ghemawat:
Just Javascript changes, so again for @josef-jelinek

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -147,6 +147,8 @@ button {
   border-bottom: 1px solid black;
   text-align: right;
   padding-left: 1em;
+  padding-top: 0.2em;
+  padding-bottom: 0.2em;
 }
 #toptable tr th:nth-child(6) { text-align: left; }
 #toptable tr th:nth-child(7) { text-align: left; }
@@ -741,6 +743,11 @@ function viewer(baseUrl, nodes) {
 <meta charset="utf-8">
 <title>{{.Title}}</title>
 {{template "css" .}}
+<style type="text/css">
+#flathdr1, #flathdr2, #cumhdr1, #cumhdr2, #namehdr {
+  cursor: ns-resize;
+}
+</style>
 </head>
 <body>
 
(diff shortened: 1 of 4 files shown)


internal/driver/webhtml.go
```diff
@@ -741,22 +744,147 @@ function viewer(baseUrl, nodes) {
...
-<script>viewer({{.BaseURL}}, {{.Nodes}})</script>
+<script>
+function makeSortable() {
+  // Regular expression to extract leading number and a suffix.
+  // E.g., "1.3s" => 1.3, "s"
+  let numParseRe = new RegExp("^([\\-+]?[0-9]+([.][0-9]+)?)(.*)$")
+
+  // Unit conversions (for toUnit values produced by measurement.go)
+  const units = new Map([
+    ["B", 1.0],
```
josef-jelinek: there is no difference when ".0" is added to a number, 1 and 1.0 is exactly the same number and type in JS.

internal/driver/webhtml.go
```diff
@@ -741,22 +744,147 @@ function viewer(baseUrl, nodes) {
...
+    ["kB", 1024.0],
+    ["MB", 1024.0 * 1024.0],
+    ["GB", 1024.0 * 1024.0 * 1024.0],
+    ["TB", 1024.0 * 1024.0 * 1024.0 * 1024.0],
+    ["PB", 1024.0 * 1024.0 * 1024.0 * 1024.0 * 1024.0],
+    ["ns", 1e-9],
+    ["us", 1e-6],
```
josef-jelinek: is there a reason for not using "µs"? ("\u00B5s" alternatively)

internal/driver/webhtml.go
```diff
@@ -741,22 +744,147 @@ function viewer(baseUrl, nodes) {
...
+  let numParseRe = new RegExp("^([\\-+]?[0-9]+([.][0-9]+)?)(.*)$")
+
+  // Unit conversions (for toUnit values produced by measurement.go)
+  const units = new Map([
+    ["B", 1.0],
+    ["kB", 1024.0],
+    ["MB", 1024.0 * 1024.0],
+    ["GB", 1024.0 * 1024.0 * 1024.0],
+    ["TB", 1024.0 * 1024.0 * 1024.0 * 1024.0],
+    ["PB", 1024.0 * 1024.0 * 1024.0 * 1024.0 * 1024.0],
```
josef-jelinek: In ES6, you can use power operator like `2**10`, `2**20`, `2**30`, ... (even `2**0` for 1)...

internal/driver/webhtml.go
```diff
@@ -741,22 +744,147 @@ function viewer(baseUrl, nodes) {
...
+      if (r.children.length <= column) continue
+      const c = r.children[column]
+      const val = c.innerText
+
+      // Extract numeric and string portions from val.
+      let num = 0
+      let suffix = val
+      if (numeric) {
+        const match = val.match(numParseRe)
+	if (match != null) {
```
josef-jelinek: nit. inconsistent indenting

internal/driver/webhtml.go
```diff
@@ -741,22 +744,147 @@ function viewer(baseUrl, nodes) {
...
+      if (row.children.length < 3) continue
+      const flatPercent = parseFloat(row.children[1].innerText)
+      if (!isNaN(flatPercent)) sumPercent += flatPercent
+      row.children[2].innerText = sumPercent.toFixed(2) + "%"
+    }
+
+    for (const r of list) {
+      rows.appendChild(r[2])
+    }
+  }
```
josef-jelinek: It looks like a lot of work can be saved if Go generated just a table data for JS and JS always sorted those and then rendered new rows (for efficiency into a html fragment that can be then added into the tbody.rows. Something like (have not tried, just winging it here):

```
var tableData = [
  // generated by {{range ...}} as before only producing JS instead of HTML
  [{label: "2ms", value: 0.002}, {label: "30%", value: 30}, {label: "ny-name", value: "my-name"}],
  ...
];
...
function sortBy(column) {
  // select the cmp to be asc or desc, but will be simpler only doing a<b, a>b,
  tableData.sort((a, b) => cmp(a[column].value, b[column].value))
  const fragment = document.createDocumentFragment()
  for (const row of tableData) {
    const tr = document.createElement('tr')
    for (const cell of row) {
      const td = document.createElement('td')
      td.textContent = cell.label
      tr.appendChild(td)
    }
    fragment.appendChild(tr)
  }
  rows.textContent = "" // empty the table body first
  rows.appendChild(fragment)
}
```

internal/driver/webhtml.go
ghemawat: Done.

internal/driver/webhtml.go
ghemawat: Done.

internal/driver/webhtml.go
ghemawat: I am parsing text produced by measurement.go, which uses "us".

internal/driver/webhtml.go
ghemawat: Done.

internal/driver/webhtml.go
ghemawat: This is an interesting point.  It will require a fair bit of surgery, both here as well as in the Go code, but we might end up in an overall better place.  I'll try that out and get back to you.

ghemawat: Done.

internal/driver/webhtml.go
```diff
@@ -767,125 +767,94 @@ function viewer(baseUrl, nodes) {
...
-      const row = r[2]
-      if (row.children.length < 3) continue
-      const flatPercent = parseFloat(row.children[1].innerText)
-      if (!isNaN(flatPercent)) sumPercent += flatPercent
-      row.children[2].innerText = sumPercent.toFixed(2) + "%"
+    entries.sort(cmp)
+    if (descending) entries.reverse()
+
+    function addCell(tr, val) {
+      const td = document.createElement('td')
```
aalexand: Should all JS code here have semicolons at the end of statements? Seems like it is a [recommended practice](https://stackoverflow.com/questions/444080/do-you-recommend-using-semicolons-after-every-statement-in-javascript). This isn't related to this PR, just curious.

ghemawat: Google Javascript style guide, which we might as well follow, says to use semicolons.

https://google.github.io/styleguide/jsguide.html#formatting-semicolons-are-required

However we should not extend this PR to add semicolons and should make a distinct change.  And I am worried about churning the code if there are substantial other changes in flight out there.  The only potential one is the flamegraph one, but that seems to not be progressing?

aalexand: Agree that adding semicolons (or any kind of wide formatting changes) would be a separate CL. The flamegraph PR #188 seems to have stalled for some reason, yes. I'll ping that. We can defer adding semicolons until that PR is done. But it doesn't seem that PR makes a lot of changes to the existing JS code?

josef-jelinek: I agree, I think I would capture potential problems with automatic semicolon insertion (ASI) in the code here, but probably it would be worth considering for a future PR extracting the static JS into a separate lintable file (that usually demands using semicolons anyway).

PR #221, opened 2017-09-14
Add flag to append a comment to a profile.

### LINKED ISSUE #193: Ability to specify a comment to record in the saved profile

aalexand:
Fixes #193.

### DIFF
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -131,12 +131,6 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
-	"comment": &variable{stringKind, "", "", helpText(
-		"Comment string to add to the profile",
-		"The string is appended to any output format that includes comments,",
-		"for example, 'proto'. The common use is with remote sources to record",
-		"a note within the saved profile.")},
-
 	// Comparisons.
 	"drop_negative": &variable{boolKind, "f", "", helpText(
 		"Ignore negative differences",
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -131,6 +131,12 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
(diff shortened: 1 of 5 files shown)


internal/driver/commands.go
```diff
@@ -131,6 +131,12 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
+	"comment": &variable{stringKind, "", "", helpText(
+		"Comment string to add to the profile",
+		"The string is appended to any output format that includes comments,",
```
rauls5382: The long help message is a bit confusing. How about something like:

"This adds free-form annotations to the profile, which are generated on some reports or can be explicitly queries with pprof -comments"

internal/driver/commands.go
```diff
@@ -131,6 +131,12 @@ var pprofVariables = variables{
 	// Filename for file-based output formats, stdout by default.
 	"output": &variable{stringKind, "", "", helpText("Output filename for file-based outputs")},
 
+	"comment": &variable{stringKind, "", "", helpText(
```
rauls5382: Not sure this should be a variable.

If somebody were to set comment="my comment" on the pprof shell it would have no effect, right?

Seems to me more like -buildid, in that it only affects the profile during fetch. Consider parsing it explicitly on parseFlags instead.

aalexand: Made it a flag similar to what `-buildid` is.

internal/driver/commands.go
aalexand: Done.

aalexand:
Addressed the comments, PTAL.

aalexand:
@rsilvera, could you look at the updated PR?

PR #224, opened 2017-09-15
Fix id assignment in dot+svg output.

ghemawat:
Previously, web interface was assuming that the ith node in
the dot graph was being assigned the id "node<i+1>".  However,
this assumption was invalid when extra nodes were inserted
into the dot output (e.g., nodelets added for allocation
profiles).  Because of this bug, node selection by clicking
in the webui graph output would sometimes pick a random
string to match against.

The problem is now fixed by manually assigning an id to
every node in the generated dot graph.  dot will propagate
this id reliably to the svg output.

Updated the golden tests to reflect the new dot output.

Extended dotgraph_test.go with an -update flag that can be
used to update golden files.  Also removed a bunch of code
duplication in the test.


aalexand:
I'll merge once the Travis build comes out green. Not sure why it takes so long to finish.

PR #229, opened 2017-09-22
Replaced http wrapper field in plugin Options with HTTPServer.

ghemawat:
HTTPServer allows somebody embedding pprof to control how the
hosting HTTP server is constructed which makes pprof work in
more environments.


aalexand:
I'd ideally like to see the test coverage going up. Is this feasible here?

ghemawat:
Increased coverage for webui.go from 50% to 66%.

PR #231, opened 2017-09-24
Make binutils thread safe.

ghemawat:
binutils is now used from multiple goroutines when the
web interface is enabled (since http server spins up a new
goroutine for each request). Added appropriate synchronization
to various objects in binutils to provide the appropriate
thread-safety.

In addition, added a bunch of tests:

. Test parsing of nm output.
. Tests for Binutils methods.
. Test llvm-symbolizer interface (using an llvm-symbolizer emulator).


internal/binutils/binutils.go
```diff
@@ -47,19 +54,50 @@ type Binutils struct {
 	fast bool
 }
 
+// get returns the current representation for b, initializing it if necessary.
```
aalexand: nit: "for bu"

internal/binutils/binutils.go
```diff
@@ -47,19 +54,50 @@ type Binutils struct {
...
+	if r == nil {
+		r = &binrep{}
+		initTools(r, "")
+		bu.rep = r
+	}
+	bu.Mutex.Unlock()
+	return r
+}
+
+// update modifies the rep for b via the supplied function.
```
aalexand: nit: "for bu"

internal/binutils/binutils.go
ghemawat: Done.

internal/binutils/binutils.go
ghemawat: Done.

ghemawat: Yes. profile.Copy should also be fixed.  I think a separate PR since it is a completely separate code path?

internal/binutils/addr2liner.go
```diff
@@ -36,6 +37,7 @@ const (
 // addr2Liner is a connection to an addr2line command for obtaining
 // address and line number information from a binary.
 type addr2Liner struct {
+	sync.Mutex
```
aalexand: Maybe make mutex a separate field like "mu sync.Mutex" like other fields? Does it need to be special?

ghemawat: Done

ghemawat:
Can this one be merged now?  If so, I can merge it back in and resubmit #232 (which seems to have failed Travis builds due to environmental flakiness).

aalexand:
There is a Travis failure below but it seems unrelated, so merging.

```
Error: SHA256 mismatch
Expected: e5435f02e02d2b87bb8e4efdcaa14b1f78c9cf3ab1ed80f94b6382fb6acc7d78
Actual: 7a07d3f7cca5c0b38ca811984ef8da536da32932d68c1a6cce33ec2462b930bf
Archive: /Users/travis/Library/Caches/Homebrew/freetype-2.8.1.tar.bz2
To retry an incomplete download, remove the file above.
The command "if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install graphviz; fi" failed and exited with 1 during .
Your build has been stopped.
```

PR #235, opened 2017-09-28
weblist output disassembly now contains inlined source.

ghemawat:
In addition, fixed the line number assigned to instructions from inlined calls.

As an illustration, see the before and after output for a "pprof -weblist" command for a simple function:

[before](https://ghemawat.github.io/scratch/tmp/list1.html)
[after](https://ghemawat.github.io/scratch/tmp/list2.html)

Note that in the "before" page, the 7.26 second line cannot be expanded, but in the "after" page clicking on it reveals a wealth of information.


ghemawat:
I am confused by these code coverage reports.

E.g., clicking through to https://codecov.io/gh/google/pprof/pull/235/src/internal/report/report.go, it says that coverage has disappeared for printCallgrind, and showed up for getDisambiguatedNames.  Neither of those two functions should have been affected by this PR.

Is the code coverage report actually reliable?

aalexand:
@ghemawat It does look confusing sometimes. E.g. on https://github.com/google/pprof/pull/236 where I just fixed the test.sh script to strip some redundant tags from the coverage output the coverage changed which I wouldn't expect. I will look more into this.

That said, looking at your [report.go](https://codecov.io/gh/google/pprof/pull/235/src/internal/report/report.go) link, where does it say the coverage for printCallgrind disappeared? I can see it all read which I think means it's just not covered at all.

aalexand:
Similarly for getDisambiguatedNames, I don't think it shows that the function is now covered, it just shows that it used to be covered and covered now. Codecov has a Chrome extension which is useful to overlay the current coverage data over the source in Github:

selection_108

internal/report/source.go
```diff
@@ -255,9 +257,33 @@ func assemblyPerSourceLine(objSyms []*objSymbol, rs graph.Nodes, src string, obj
...
+		//
+		// E.g., suppose we are printing source code for F and this
+		// instruction is from H where F called G called H and both
+		// of those calls were inlined.  We want to use the line
+		// number from F, not from H (which is what Disasm gives us).
+		//
+		// So find the outer-most linenumber in the source file.
+		frames, err := o.file.SourceLine(an.address + o.base)
+		found := false
+		if err == nil {
```
aalexand: Nit: Go style is to limit the scope of variables when possible by putting the assignment within if:

```
found := false
if frames, err := o.file.SourceLine(an.address + o.base); err == nil {
   ...
```

Also, should something be done with the error?

internal/report/source.go
```diff
@@ -483,6 +514,55 @@ func getMissingFunctionSource(filename string, asm map[int][]assemblyInstruction
...
+}
+
+func (reader *sourceReader) line(path string, lineno int) (string, bool) {
+	lines, ok := reader.files[path]
+	if !ok {
+		// Read and cache file contents.
+		lines = []string{""} // Skip 0th line
+		f, err := openSourceFile(path, reader.searchPath)
+		if err != nil {
+			reader.errors[path] = err
```
aalexand: Cover the error path?

internal/report/source.go
```diff
@@ -529,3 +609,20 @@ func trimPath(path string) string {
 	}
 	return path
 }
+
+func indentation(line string) int {
```
aalexand: Add a test.

internal/report/source.go
```diff
@@ -255,9 +257,33 @@ func assemblyPerSourceLine(objSyms []*objSymbol, rs graph.Nodes, src string, obj
...
+		// Do not rely solely on the line number produced by Disasm
+		// since it is not what we want in the presence of inlining.
+		//
+		// E.g., suppose we are printing source code for F and this
+		// instruction is from H where F called G called H and both
+		// of those calls were inlined.  We want to use the line
+		// number from F, not from H (which is what Disasm gives us).
+		//
+		// So find the outer-most linenumber in the source file.
+		frames, err := o.file.SourceLine(an.address + o.base)
```
aalexand: From the [diff view in the coverage UI](https://codecov.io/gh/google/pprof/pull/235/diff) looks like this new code is not covered, apparently because the enclosing function is not covered. Consider adding a test, though feel free to decline because the lack of test seems pre-existing.

aalexand:
The change is pretty cool, I forgot to say.

internal/report/testdata/sample/sample.go
```diff
@@ -0,0 +1,41 @@
...
+		log.Fatal("could not create CPU profile: ", err)
+	}
+	if err := pprof.StartCPUProfile(f); err != nil {
+		log.Fatal("could not start CPU profile: ", err)
+	}
+	defer pprof.StopCPUProfile()
+	busyLoop()
+}
+
+func busyLoop() {
```
aalexand: FWIW, I am not sure Golang includes the DWARF inline debug information, at least I don't see the inline expansion working with this test case - https://aalexand.github.io/scratch/busyloop.html. We can improve it later I guess.

internal/report/source.go
aalexand: I was not sure what caused "reader.errors[path] = err" line under the "err != nil" condition to be covered, which is what the coverage says now. The test case provides the source files so my feeling was it shouldn't get into the error handler here. Maybe the coverage info is off.

internal/report/testdata/sample/sample.go
aalexand: Agree, it's better to avoid fakes as much as possible.

PR #240, opened 2017-10-08
Fix unit warnings

### LINKED ISSUE #239: "For tag bytes used unit bytes, also encountered unit(s)" error message when opening a heap profile
Steps to reproduce - see below. Note the "For tag bytes used unit bytes, also encountered unit(s)" message that shouldn't be printed.

```
$ go run pprof.go -top ./profile/testdata/cppbench.heap
Local symbolization failed for cppbench_server_main: stat /home/cppbench_server_main: no such file or directory
Local symbolization failed for libpthread-2.15.so: stat /lib/libpthread-2.15.so: no such file or directory
Local symbolization failed for libc-2.15.so: stat /lib/libc-2.15.so: no such file or directory
Some binary filenames not available. Symbolization may be incomplete.
Try setting PPROF_BINARY_PATH to the search path for local binaries.
For tag bytes used unit bytes, also encountered unit(s) 
File: cppbench_server_main
Type: space
Showing nodes accounting for 77.19MB, 100% of 77.19MB total
   77.19MB   100%   100%    77.19MB   100%  [cppbench_server_main]
         0     0%   100%        2MB  2.59%  [libc-2.15.so]
         0     0%   100%    73.18MB 94.81%  [libpthread-2.15.so]
```

nolanmar511:
Now, warnings will only be printed if multiple units are associated with one key, but not if units are inferred based on the keys, units are inferred, or some values are missing units.

Fixes #239

### DIFF
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -450,20 +450,16 @@ func (p *Profile) Aggregate(inlineFrame, function, filename, linenumber, address
 func (p *Profile) NumLabelUnits() (map[string]string, map[string][]string) {
 	numLabelUnits := map[string]string{}
 	ignoredUnits := map[string]map[string]bool{}
+	encounteredKeys := map[string]bool{}
 
 	// Determine units based on numeric tags for each sample.
 	for _, s := range p.Sample {
-		for k, vs := range s.NumLabel {
-			units := s.NumUnit[k]
-			if len(units) != len(vs) {
-				if _, ok := numLabelUnits[k]; !ok {
-					numLabelUnits[k] = ""
-				} else {
-					ignoredUnits[k] = map[string]bool{"": true}
+		for k := range s.NumLabel {
+			encounteredKeys[k] = true
+			for _, unit := range s.NumUnit[k] {
+				if unit == "" {
+					continue
 				}
-				continue
-			}
-			for _, unit := range units {
 				if wantUnit, ok := numLabelUnits[k]; !ok {
 					numLabelUnits[k] = unit
 				} else if wantUnit != unit {
@@ -476,10 +472,10 @@ func (p *Profile) NumLabelUnits() (map[string]string, map[string][]string) {
 			}
 		}
 	}
-
 	// Infer units for keys without any units associated with
 	// numeric tag values.
-	for key, unit := range numLabelUnits {
+	for key := range encounteredKeys {
+		unit := numLabelUnits[key]
 		if unit == "" {
 			switch key {
 			case "alignment", "request":
--- a/profile/profile_test.go
+++ b/profile/profile_test.go
@@ -811,9 +811,9 @@ func locationHash(s *Sample) string {
 	return tb
 }
 
-func TestInferUnits(t *testing.T) {
+func TestNumLabelUnits(t *testing.T) {
 	var tagFilterTests = []struct {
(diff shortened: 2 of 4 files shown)


profile/profile_test.go
```diff
@@ -811,7 +811,7 @@ func locationHash(s *Sample) string {
 	return tb
 }
 
-func TestInferUnits(t *testing.T) {
+func TestNumLabelUnits(t *testing.T) {
```
aalexand: Do the modified tests fail with the original code in profile.go for the reason same as in the issue? I am not sure they do. I actually just tried it, here is what I see:

```
profile_test.go:925: One sample, one key with multiple values and three different units: for key key1, got unit key1, want unit bytes
```

But this failure is unrelated - this is from the second case you've added which tests that if there are more units than there unit values, we still process the units for the values rather than pretending there are no units at all. This change is unrelated to the issue I believe?

I believe the test case that is missing is something like modifying this existing one:

```
                {
                        "Key alignment, unit not specified",
-                       []map[string][]int64{{"alignment": {8}}},
-                       []map[string][]string{nil},
+                       []map[string][]int64{{"alignment": {8}}, {"alignment": {8}}},
+                       []map[string][]string{nil, nil},
                        map[string]string{"alignment": "bytes"},
                        map[string][]string{},
                },
```

Plus, (and probably more importantly), the comparison of want and got maps for units and ignoredUnits does not quite test the equality and needs to be updated. These are now done as `for key, wantUnit := range test.wantUnits {` and `for key, ignored := range ignoredUnits` respectively and they don't seem to compare the size of want and got maps before the loop, so these only seem to check that test.wantUnits is a subset of gotUnits and that gotIgnoredUnits is a subset of test.wantIgnoredUnits. Besides the size checks, make what you iterate by consistent perhaps - there isn't a good reason to have the loop over want things in the first loop but by got things in the second loop, is there?

profile/profile.go
```diff
@@ -450,20 +450,16 @@ func (p *Profile) Aggregate(inlineFrame, function, filename, linenumber, address
 func (p *Profile) NumLabelUnits() (map[string]string, map[string][]string) {
 	numLabelUnits := map[string]string{}
 	ignoredUnits := map[string]map[string]bool{}
+	encounteredKeys := map[string]bool{}
```
aalexand: I am somewhat curious about the performance impact this additional map might have as the loop in this function is really tight being called for every sample in a profile, but I don't have any data. We should probably add a [benchmark test](https://golang.org/pkg/testing/#hdr-Benchmarks) for profile encoding / decoding.

aalexand: Adding the benchmark shouldn't be part of this PR.

profile/profile_test.go
nolanmar511: Modified tests did failed, but I added the test you described. And I modified comparison of want and got for this test.

nolanmar511:
PTAL

profile/profile_test.go
aalexand: FWIW, yes, the tests failed with the error I mentioned, but as I said I think they failed for reason unrelated to the one that caused the warning messages I think.

nolanmar511:
Just changed empty lists to nil in a couple spots.

PR #243, opened 2017-10-17
Show menu on click, not hover.

ghemawat:
Previously, users had to be careful to not move the mouse
outside the menu (otherwise it would disappear unintentionally).

Also fixed a minor bug where we were not overwriting URL
parameters correctly when no selection was made.

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -85,9 +85,7 @@ button {
   position: relative;
   display: inline-block;
   padding: 2px 2px;
-  cursor: default;
   font-size: 14pt;
-  user-select: none;
 }
 .menu {
   display: none;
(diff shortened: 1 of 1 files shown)


ghemawat:
Thanks.  I looked at caniuse: most features have been available for at least several months on Edge/FF/Chrome/Safari.  IE is a laggard, but I think we should be fine with that (pprof is targeting developers).

The "passive" event listener support is just a few days old (Edge 16).  Should I drop it?  It might cause a potential performance issue for touch events, but probably not a big deal.

josef-jelinek:
It should be fine, just something to be aware of... The user-select is implemented now everywhere, but it does not work outside Chrome unless you use prefixed versions like -ms-user-select, -moz-user-select, -webkit-user-select.

ghemawat:
Added vendor prefixes for user-select.

aalexand:
@josef-jelinek approve again and merge if LGTY?

PR #249, opened 2017-10-29
Locate symbol endpoints with deeper paths

anacrolix:
https://github.com/golang/go/issues/22483


anacrolix:
Oops, forgot to commit those.

PR #250, opened 2017-10-29
Improve POST URL status error

anacrolix:
Errors due to unwanted status codes in POST responses give no URL context. This adds it to match the style used for connection issues. This is useful for example when symbol look fails and generates a 404. Previously there was no mention of why a 404 error occurred, and the action that triggered it.

### DIFF
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -121,7 +121,7 @@ func postURL(source, post string) ([]byte, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return nil, statusCodeError(resp)
+		return nil, fmt.Errorf("%s: %v", source, statusCodeError(resp))
 	}
 	return ioutil.ReadAll(resp.Body)
 }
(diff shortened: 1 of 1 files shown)


internal/symbolizer/symbolizer.go
```diff
@@ -121,7 +121,7 @@ func postURL(source, post string) ([]byte, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return nil, statusCodeError(resp)
+		return nil, fmt.Errorf("%s: %v", source, statusCodeError(resp))
```
aalexand: Nit: "http post %s: %v" to be consistent with the other error above.

internal/symbolizer/symbolizer.go
```diff
@@ -121,7 +121,7 @@ func postURL(source, post string) ([]byte, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
```
aalexand: Optional: this error code path doesn't seem to be tested. It's not quite fair to ask you to add missing tests but if you will like it, it would be welcome.

PR #254, opened 2017-11-03
Fix TestHttpsInsecure on Plan 9

### LINKED ISSUE #253: TestHttpsInsecure failing on Plan 9
Changeset 696c91bc6dcd6c603a1685c5f623f5ca0ddafc06 added TestHttpsInsecure, which is failing on Plan 9:

```
cpu% go test -v -run TestHttpsInsecure
=== RUN   TestHttpsInsecure
--- FAIL: TestHttpsInsecure (11.28s)
	fetch_test.go:420: fetchProfiles(https+insecure://127.0.0.1:47334/debug/pprof/profile) got non-symbolized profile: len(p.Function)==0
	fetch_test.go:376: Serve got error accept tcp 127.0.0.1:47334: open /net/tcp/11/listen: '/net/tcp/11/listen' listen hungup, want "use of closed"
FAIL
exit status: 'driver.test 1773648: 1'
FAIL	cmd/vendor/github.com/google/pprof/internal/driver	11.370s
```

There are two issues:

- CPU profiling is not implemented on Plan 9, so the profile is empty.
- Closing a closed listener on Plan 9 returns `listen hungup` instead of `use of closed` (see [/sys/src/9/ip/devip.c:476](https://github.com/0intro/plan9/blob/4225c3e/sys/src/9/ip/devip.c#L476)).

0intro:
These two changes fix TestHttpsInsecure on Plan 9.

Fixes #253.

### DIFF
--- a/internal/driver/fetch_test.go
+++ b/internal/driver/fetch_test.go
@@ -351,6 +351,13 @@ func stubHTTPGet(source string, _ time.Duration) (*http.Response, error) {
 	return c.Get("file:///" + file)
 }
 
+func closedError() string {
+	if runtime.GOOS == "plan9" {
+		return "listen hungup"
+	}
+	return "use of closed"
+}
+
 func TestHttpsInsecure(t *testing.T) {
 	if runtime.GOOS == "nacl" {
 		t.Skip("test assumes tcp available")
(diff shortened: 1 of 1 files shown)


internal/driver/fetch_test.go
```diff
@@ -416,6 +423,10 @@ func TestHttpsInsecure(t *testing.T) {
 	if len(p.SampleType) == 0 {
 		t.Fatalf("fetchProfiles(%s) got empty profile: len(p.SampleType)==0", address)
 	}
+	if runtime.GOOS == "plan9" {
```
aalexand: There is badSigprofOS map below which lists plan9. That map is for environments where the CPU profiling is supported but produces incorrect results. I suggest we remove plan9 from there since it's now handled here.

aalexand:
The description should say "Fixes" or is there work left?

internal/driver/fetch_test.go
0intro: Done.

0intro:
I've updated the commits.

PR #256, opened 2017-11-07
internal/driver: skip TestHttpsInsecure test on iOS

eliasnaur:
Also, replace a PrintErr with Print to avoid a warning message
being registered as an error.

This will in turn fix golang.org/issues/22612 for the Go iOS
builders.

### DIFF
--- a/internal/driver/fetch_test.go
+++ b/internal/driver/fetch_test.go
@@ -411,9 +411,14 @@ func TestHttpsInsecure(t *testing.T) {
 		Timeout:   10,
 		Symbolize: "remote",
 	}
+	rx := "Saved profile in"
+	if runtime.GOOS == "darwin" && (runtime.GOARCH == "arm" || runtime.GOARCH == "arm64") {
+		// On iOS, $HOME points to the app root directory and is not writable.
+		rx += "|Could not use temp dir"
+	}
 	o := &plugin.Options{
 		Obj: &binutils.Binutils{},
-		UI:  &proftest.TestUI{T: t, AllowRx: "Saved profile in"},
+		UI:  &proftest.TestUI{T: t, AllowRx: rx},
 	}
 	o.Sym = &symbolizer.Symbolizer{Obj: o.Obj, UI: o.UI}
 	p, err := fetchProfiles(s, o)
@@ -423,9 +428,15 @@ func TestHttpsInsecure(t *testing.T) {
 	if len(p.SampleType) == 0 {
 		t.Fatalf("fetchProfiles(%s) got empty profile: len(p.SampleType)==0", address)
 	}
-	if runtime.GOOS == "plan9" {
+	switch runtime.GOOS {
+	case "plan9":
 		// CPU profiling is not supported on Plan9; see golang.org/issues/22564.
 		return
+	case "darwin":
+		if runtime.GOARCH == "arm" || runtime.GOARCH == "arm64" {
+			// CPU profiling on iOS os not symbolized; see golang.org/issues/22612.
+			return
+		}
 	}
 	if len(p.Function) == 0 {
 		t.Fatalf("fetchProfiles(%s) got non-symbolized profile: len(p.Function)==0", address)


eliasnaur:
I haven't contributed to this repository before, but fwiw the Travis failures doesn't seem related to this PR.

internal/driver/fetch_test.go
```diff
@@ -423,9 +423,15 @@ func TestHttpsInsecure(t *testing.T) {
...
 		t.Fatalf("fetchProfiles(%s) got empty profile: len(p.SampleType)==0", address)
 	}
-	if runtime.GOOS == "plan9" {
+	switch runtime.GOOS {
+	case "plan9":
 		// CPU profiling is not supported on Plan9; see golang.org/issues/22564.
 		return
+	case "darwin":
+		if runtime.GOARCH == "arm" || runtime.GOARCH == "arm64" {
+			// CPU profiling is not supported on OS; see golang.org/issues/22612.
```
aalexand: Is CPU profiling not supported on iOS in the same way as on plan9, or the quality of profiles is low and there are no symbolized function names in the profile? Looking at https://github.com/golang/go/blob/83a1a2ba63cd79ed65a97bead2c31ef0753f4cd2/src/runtime/pprof/pprof_test.go#L148, I don't see the CPU profile test being skipped in the Go tests?

internal/driver/fetch.go
```diff
@@ -294,7 +294,7 @@ func setTmpDir(ui plugin.UI) (string, error) {
 	dirs = append(dirs, os.TempDir())
 	for _, tmpDir := range dirs {
 		if err := os.MkdirAll(tmpDir, 0755); err != nil {
-			ui.PrintErr("Could not use temp dir ", tmpDir, ": ", err.Error())
+			ui.Print("Could not use temp dir ", tmpDir, ": ", err.Error())
```
aalexand: I don't think it's the right way to fix the single test on a single platform, as it changes the behavior globally.

You can conditionally (for the problematic environment only) add a pattern to ignore in the configuration of the TestUI in the test, https://github.com/google/pprof/blob/79c4198ef7bd1069f8f56501fc05f0f1d2c33d8a/internal/driver/fetch_test.go#L416 which should be sufficient.

I also wonder if the
```
proftest.go:114: Could not use temp dir /private/var/mobile/Containers/Data/Application/2B578AA2-4677-47C6-8665-DDB9509D5C92/pprof: mkdir /private/var/mobile/Containers/Data/Application/2B578AA2-4677-47C6-8665-DDB9509D5C92/pprof: operation not permitted
```
message from the test represents a real issue with temporary directory location on iOS. That can be handled separately though.

internal/driver/fetch_test.go
aalexand: Is there a more useful bug to link, the one which actually tracks the profiling issue on iOS rather than points to the pprof's test breakage? Like we list golang.org/issues/22564 for Plan9.

internal/driver/fetch.go
eliasnaur: setTmpDir tries $HOME/pprof first before os.TempDir. On iOS $HOME is the base directory of the installed app and isn't writable so I assumed the "Could not use temp dir" should be a warning, not an error.

However, I changed the PrintErr back and made your suggested change to the ignore pattern.

internal/driver/fetch_test.go
eliasnaur: I think CPU profiling works on iOS, so my guess would be missing symbols, judging from the failure log:

````
Local symbolization failed for cppbench_server_main: stat cppbench_server_main: no such file or directory
Local symbolization failed for libpthread-2.15.so: stat /libpthread-2.15.so: no such file or directory
Some binary filenames not available. Symbolization may be incomplete.
Try setting PPROF_BINARY_PATH to the search path for local binaries.
Local symbolization failed for cppbench_server_main: stat cppbench_server_main: no such file or directory
Local symbolization failed for libpthread-2.15.so: stat /libpthread-2.15.so: no such file or directory
Some binary filenames not available. Symbolization may be incomplete.
Try setting PPROF_BINARY_PATH to the search path for local binaries.
Local symbolization failed for cppbench_server_main: stat cppbench_server_main: no such file or directory
Local symbolization failed for libpthread-2.15.so: stat /libpthread-2.15.so: no such file or directory
Some binary filenames not available. Symbolization may be incomplete.
Try setting PPROF_BINARY_PATH to the search path for local binaries.
````

I updated the comment to reflect that.

eliasnaur: As per my comment above, I don't there is an issue with the basic CPU profiling machinery on iOS.

eliasnaur:
I've amended the PR to match your comments and suggestions. Sorry for the drive-by nature of this; I'm trying to get the Go iOS builder green again after refreshing the vendored copy of this package but I know very little about this package (nor the runtime/pprof package).

aalexand:
The https://travis-ci.org/google/pprof/jobs/299036158 failure is unrelated, it's #255

Issue #244, opened 2017-10-19
Fetch need a https fetch


xianlubird:
I think we should provide a https fetch that accept ca cert file to query https server's info

aalexand:
@xianlubird This should be implemented now, so if you could verify the support it would be good.

PR #261, opened 2017-11-14
Support HTTPS certs, keys and CAs

sjug:
As subject says, allows users to pass `-cert` `-key` and `-CA` so that pprof can be used against authenticated TLS endpoints.

Closes https://github.com/golang/go/issues/20939 and https://github.com/google/pprof/issues/244

### DIFF
--- a/internal/driver/cli.go
+++ b/internal/driver/cli.go
@@ -40,7 +40,7 @@ type source struct {
 // Parse parses the command lines through the specified flags package
 // and returns the source of the profile and optionally the command
 // for the kind of report to generate (nil for interactive use).
-func parseFlags(o *plugin.Options) (*source, []string, error) {
+func parseFlags(o *plugin.Options) (*source, *plugin.TLSParams, []string, error) {
 	flag := o.Flagset
 	// Comparisons.
 	flagBase := flag.StringList("base", "", "Source for base profile for comparison")
@@ -61,6 +61,10 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flagContentions := flag.Bool("contentions", false, "Display number of delays at each region")
 	flagMeanDelay := flag.Bool("mean_delay", false, "Display mean delay at each region")
 	flagTools := flag.String("tools", os.Getenv("PPROF_TOOLS"), "Path for object tool pathnames")
+	// HTTPS TLS parameters for profile and symbol fetch
+	flagCertFile := flag.String("cert", "", "TLS client certificate file for fetching profile and symbols")
+	flagKeyFile := flag.String("key", "", "TLS private key file for fetching profile and symbols")
+	flagCAFile := flag.String("ca", "", "TLS CA certs file for fetching profile and symbols")
 
 	flagHTTP := flag.String("http", "", "Present interactive web based UI at the specified http host:port")
 
@@ -85,7 +89,7 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 			usageMsgVars)
 	})
 	if len(args) == 0 {
-		return nil, nil, fmt.Errorf("no profile source specified")
+		return nil, nil, nil, fmt.Errorf("no profile source specified")
 	}
 
 	var execName string
@@ -104,15 +108,15 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 
 	// Report conflicting options
 	if err := updateFlags(installedFlags); err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	cmd, err := outputFormat(flagCommands, flagParamCommands)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 	if cmd != nil && *flagHTTP != "" {
-		return nil, nil, fmt.Errorf("-http is not compatible with an output format on the command line")
+		return nil, nil, nil, fmt.Errorf("-http is not compatible with an output format on the command line")
 	}
 
 	si := pprofVariables["sample_index"].value
@@ -140,6 +144,12 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 		Comment:      *flagAddComment,
 	}
 
+	tlsParam := &plugin.TLSParams{
+		HTTPSCert: *flagCertFile,
+		HTTPSKey:  *flagKeyFile,
+		HTTPSCA:   *flagCAFile,
+	}
+
 	for _, s := range *flagBase {
 		if *s != "" {
 			source.Base = append(source.Base, *s)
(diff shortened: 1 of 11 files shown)


sjug:
Will fix the tests shortly.

internal/driver/cli.go
```diff
@@ -61,6 +64,10 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flagContentions := flag.Bool("contentions", false, "Display number of delays at each region")
 	flagMeanDelay := flag.Bool("mean_delay", false, "Display mean delay at each region")
 	flagTools := flag.String("tools", os.Getenv("PPROF_TOOLS"), "Path for object tool pathnames")
+	// HTTPS options
+	flagCertFile := flag.String("cert", "", "TLS client certificate file")
+	flagKeyFile := flag.String("key", "", "TLS private key file")
+	flagCAFile := flag.String("CA", "", "TLS CA certs file")
```
aalexand: It's a bit strange to have command line flag in capital letters. Is there a precedent for this?

vbatts:
seems fine to me

internal/driver/fetch.go
```diff
@@ -580,17 +581,35 @@ func adjustURL(source string, duration, timeout time.Duration) (string, time.Dur
...
+		caCertPool := x509.NewCertPool()
+		caCert, err := ioutil.ReadFile(s.HTTPSCA)
+		if err != nil {
+			return nil, err
 		}
+		caCertPool.AppendCertsFromPEM(caCert)
+		tlsConfig.RootCAs = caCertPool
+	}
+
+	if url.Scheme == "https+insecure" {
```
aalexand: Should `internal/symbolizer/symbolizer.go` be updated as well?

internal/driver/cli.go
sjug: Agreed.

internal/driver/fetch.go
sjug: Do you mean the `postURL` in `symbolizer.go`?

aalexand: Yes.

aalexand:
@sjug Did you also see my

> Can you please add tests that verify the new functionality end-to-end?

request? Just want to make sure it's not missed since it was a whole-PR comment rather than on a specific line and so looks subtle in the GH UI.

sjug:
@aalexand Yep, didn't miss it. Will do.

internal/driver/fetch.go
sjug: The problem is the `source` struct which has the command line flags is in `driver` pkg. That's why `Symbolizer()` takes `mode string` rather than `s *source`. In order to keep it clean I would need to create a new package for the `source` struct, do you have any preference where it should go?

internal/driver/cli.go
```diff
@@ -34,6 +34,9 @@ type source struct {
 	Timeout      int
 	Symbolize    string
 	HTTPHostport string
+	HTTPSCert    string
```
aalexand: Perhaps move these up to follow Sources field to emphasize that this is about setting up the TLS for fetching the profiles not for the HTTP server mode of pprof itself.

internal/driver/cli.go
```diff
@@ -61,6 +64,10 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flagContentions := flag.Bool("contentions", false, "Display number of delays at each region")
 	flagMeanDelay := flag.Bool("mean_delay", false, "Display mean delay at each region")
 	flagTools := flag.String("tools", os.Getenv("PPROF_TOOLS"), "Path for object tool pathnames")
+	// HTTPS options
+	flagCertFile := flag.String("cert", "", "TLS client certificate file")
```
aalexand: I like the brevity of the docs but we should here also clarify that this about the TLS params for profile and symbol fetch.

internal/driver/cli.go
```diff
@@ -280,6 +290,9 @@ var usageMsgSrc = "\n\n" +
 	"  Source options:\n" +
 	"    -seconds              Duration for time-based profile collection\n" +
 	"    -timeout              Timeout in seconds for profile collection\n" +
+	"    -key                  Specify TLS private key\n" +
```
aalexand: Ditto (I like the brevity of the docs but we should here also clarify that this about the TLS params for profile and symbol fetch).

internal/driver/fetch.go
```diff
@@ -580,17 +581,35 @@ func adjustURL(source string, duration, timeout time.Duration) (string, time.Dur
 
 // httpGet is a wrapper around http.Get; it is defined as a variable
 // so it can be redefined during for testing.
-var httpGet = func(source string, timeout time.Duration) (*http.Response, error) {
+var httpGet = func(s *source, source string, timeout time.Duration) (*http.Response, error) {
```
aalexand: We shouldn't need to pass all of "s *source" here, we only need the TLS params?

internal/driver/fetch.go
aalexand: I don't think we should pass all of `source` struct to the symbolizer, it doesn't need most of that. Maybe something like this in `internal/plugin/plugin.go`:

```
type TLSParams struct {
  HTTPSCert, HTTPSKey, HTTPSCA string
}

and then pass it as to the symbolizer and to the fetcher as a new param.

internal/driver/fetch.go
sjug: Just seemed cleaner to not create anything extra, just reuse the existing struct. If we make the `TLSParams` in `plugin.go` that should work for `Symbolizer()` as well. Isn't having the same `HTTPS*` fields in the `source` struct redundant?

aalexand: I would TLSParams in the source structure as well to avoid the duplicated fields, but (as I said) would pass the source / URL string and the TLS params alone instead of passing the whole input structure. Passing only what's needed is better than being able to avoid another little struct imo.

internal/driver/fetch.go
sjug: Changed the implementation to use `TLSParams`, thoughts?

aalexand:
The only thing that concerns me is that the TLS parameters are visible in the help and are exposed in the method signatures now regardless of whether the underlying symbolizer and fetcher need them. For example, at Google internally we use a custom fetch implementation which wouldn't need these options. I wonder if there is a better way to abstract these into the implementation of the fetcher and symbolizer. @rauls5382 do you have some ideas?

sjug:
@aalexand Has a decision been made as to how this should be implemented so I can move forward with writing tests for this?

aalexand:
Not yet. I was curious to hear @rauls5382's thoughts. Raul, any thoughts here?

dgryski:
Ping  @rauls5382.

nolanmar511:
This functionality was added with https://github.com/google/pprof/pull/423, so I'm closing this PR.

Thanks for all the work here!

PR #266, opened 2017-11-21
fix: viewer function receiving null node list

spiermar:
Further testing revealed an issue with the flame graph search function. Since a node list was not being exported anymore after the `call_tree` change, the `selectMatching` function was throwing an exception because the the list was `null`. This is a simple fix, following the same approach as the _dot_ view and exporting the node name list again. The other option would be to move a few functions out of the `{{define "script"}}` template, load them separately in the _dot_ view, and not call the `viewer` function in the flame graph view.


spiermar:
@aalexand heads up, since I've found it this morning.

PR #267, opened 2017-11-24
Improve indenting on webhtml.go

tdewolff:
This commit is merely WS changes in `webhtml.go`, append `?w=1` to the GitHub URL to ignore WS changes.

PS: I prefer if we can go to indenting using four spaces as this is my default in Vim, but not a biggie


tdewolff:
Takes more time to work with Git to rebase than to redo the indentation work. Let me get back on this later....

PR #269, opened 2017-11-27
Add semicolons to all of JS

tdewolff:


aalexand:
I think we should follow Google's JS style in this repo and that uses single quotes: https://google.github.io/styleguide/jsguide.html#features-string-literals

tdewolff:
Personally I'm not a fan of the single-line if-statements, but I guess it makes the code shorter.

aalexand:
@tdewolff Using the one line if statement for the shorter then blocks is optional - the guide says "may". I personally don't like it as well, so you didn't have to make this change. Let me know if you want to revert it, or leave as is, I don't have a preference.

tdewolff:
I haven't changed the if-statements yet, I'll leave them as they are.

The build failure appears to be from something else...

aalexand:
@tdewolff LGTM, shall I merge?

tdewolff:
Yes please

Issue #276, opened 2017-12-06
Compatibility with original pprof for remote profiling is broken (symbolization)


ErwanLegrand:
### What version of pprof are you using?

Tested with Git revision 4b386c29ec7eb5d6227bae1e3e3c0c826d681389

### What operating system and processor architecture are you using?

Linux on AMD64

### What did you do?

I am trying to profile a remote server which implements the remote profiling protocol defined by gperftools (formerly Google Perftools): https://gperftools.github.io/gperftools/pprof_remote_servers.html

I used the following commands:
```
$ pprof -symbolize=remote https+insecure://XXX/pprof/profile
$ pprof -symbolize=remote https+insecure://XXX/pprof/heap
```
### What did you expect to see?

I expected profiling to succeed, as it does with the original (Perl) version of pprof.

### What did you see instead?
```
$ pprof -symbolize=remote https+insecure://XXX/pprof/heap
Fetching profile over HTTP from https+insecure://XXX/pprof/heap
pprof: http post https://XXX/symbolz: server response: 404 Not Found
```
Thus, the new pprof sends a request to /symbolz instead of /pprof/symbol as the original pprof did.

aalexand:
I think the fix would be around https://github.com/google/pprof/blob/4b386c29ec7eb5d6227bae1e3e3c0c826d681389/internal/symbolz/symbolz.go#L67

Can you try changing that to check if the path ends with /pprof/profile or /pprof/heap and use /pprof/symbol in that case and see if that works?

ErwanLegrand:
I thank you for the fast reply!

Indeed, I managed to make pprof request the expected /pprof/symbol URL with a simple change to this code.

But now I have another issue.
```
$ ~/go/src/github.com/google/pprof/pprof --symbolize=remote https+insecure://XXX/pprof/heap
Fetching profile over HTTP from https+insecure://XXX/pprof/heap
pprof: unexpected negative adjusted address, mapping &{4 0 18446744073709551615 0 https+insecure://root:admin@localhost:8080/pprof/heap  false false false false 0 0} source 18446744073709551615, offset 0
``` 

I do not know at this point whether the problem lies in the code of pprof or in the code of the server.

aalexand:
/cc @rauls5382 who might have some thoughts.

The way it works is that

https://github.com/google/pprof/blob/a613757740614d4865db70ca9c80bc55cab3ab8f/internal/elfexec/elfexec.go#L174

[...]

PR #278, opened 2017-12-07
Change URL matching code to restore compatibility with old pprof (#276)

ErwanLegrand:
Here are the changes I have made to restore compatibility with the old pprof. I have left `strings.Contains(url.Path, "/debug/pprof/")` in the condition, together with the two `HasSuffix()` tests I have added. Leaving it seemed like the safe thing to do.

### DIFF
--- a/internal/symbolz/symbolz.go
+++ b/internal/symbolz/symbolz.go
@@ -65,7 +65,7 @@ func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, sy
 
 // Check whether path ends with one of the suffixes listed in
 // pprof_remote_servers.html from the gperftools distribution
-func hasLegacySuffix(path string) bool {
+func hasGperftoolsSuffix(path string) bool {
 	var suffixes = [...]string{
 		"/pprof/heap",
 		"/pprof/growth",
@@ -84,8 +84,8 @@ func hasLegacySuffix(path string) bool {
 // symbolz returns the corresponding symbolz source for a profile URL.
 func symbolz(source string) string {
 	if url, err := url.Parse(source); err == nil && url.Host != "" {
-		// All paths in the net/http/pprof Go package start with /debug/pprof/
-		if strings.Contains(url.Path, "/debug/pprof/") || hasLegacySuffix(url.Path) {
+		// All paths in the net/http/pprof Go package contain /debug/pprof/
+		if strings.Contains(url.Path, "/debug/pprof/") || hasGperftoolsSuffix(url.Path) {
 			url.Path = path.Clean(url.Path + "/../symbol")
 		} else {
 			url.Path = "/symbolz"
--- a/internal/symbolz/symbolz.go
+++ b/internal/symbolz/symbolz.go
@@ -63,10 +63,29 @@ func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, sy
 	return nil
 }
 
+// Check whether path ends with one of the suffixes listed in
+// pprof_remote_servers.html from the gperftools distribution
+func hasLegacySuffix(path string) bool {
+	var suffixes = [...]string {
+		"/pprof/heap",
+		"/pprof/growth",
+		"/pprof/profile",
+		"/pprof/pmuprofile",
+		"/pprof/contention",
+	}
+	for _, s := range suffixes {
+		if strings.HasSuffix(path, s) {
+			return true
+		}
+	}
+	return false
+}
+
 // symbolz returns the corresponding symbolz source for a profile URL.
 func symbolz(source string) string {
 	if url, err := url.Parse(source); err == nil && url.Host != "" {
-		if strings.Contains(url.Path, "/debug/pprof/") || strings.HasSuffix(url.Path, "/pprof/profile") || strings.HasSuffix(url.Path, "/pprof/heap") {
+		// All paths in the net/http/pprof Go package start with /debug/pprof/
+		if strings.Contains(url.Path, "/debug/pprof/") || hasLegacySuffix(url.Path) {
 			url.Path = path.Clean(url.Path + "/../symbol")
 		} else {
 			url.Path = "/symbolz"
--- a/internal/symbolz/symbolz.go
+++ b/internal/symbolz/symbolz.go
@@ -66,7 +66,7 @@ func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, sy
 // Check whether path ends with one of the suffixes listed in
 // pprof_remote_servers.html from the gperftools distribution
 func hasGperftoolsSuffix(path string) bool {
-	var suffixes = [...]string{
+	suffixes := []string{
 		"/pprof/heap",
 		"/pprof/growth",
 		"/pprof/profile",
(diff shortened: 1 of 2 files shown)


ErwanLegrand:
I have signed it!

(I would have thought I had signed this already. I'm pretty sure I made some minor contribution to Go in the past.)

aalexand:
Thank you! There is symbolz_test.go with tests, could you update it with some test cases that fail before your change and pass after?

ErwanLegrand:
I did not even notice the test file! I'll update it right away.

ErwanLegrand:
I can't run the tests locally. I get the following errors:

> + go test -v github.com/ErwanLegrand/pprof/fuzz github.com/ErwanLegrand/pprof/internal/binutils github.com/ErwanLegrand/pprof/internal/driver github.com/ErwanLegrand/pprof/internal/elfexec github.com/ErwanLegrand/pprof/internal/graph github.com/ErwanLegrand/pprof/internal/measurement github.com/ErwanLegrand/pprof/internal/report github.com/ErwanLegrand/pprof/internal/symbolizer github.com/ErwanLegrand/pprof/internal/symbolz github.com/ErwanLegrand/pprof/profile
> internal/binutils/binutils.go:29:2: use of internal package not allowed
> internal/binutils/addr2liner.go:26:2: use of internal package not allowed
> internal/driver/cli.go:22:2: use of internal package not allowed
> internal/driver/flamegraph.go:23:2: use of internal package not allowed
> internal/driver/driver_focus.go:23:2: use of internal package not allowed
> internal/driver/cli.go:23:2: use of internal package not allowed
> internal/driver/commands.go:30:2: use of internal package not allowed
> internal/driver/options.go:27:2: use of internal package not allowed
> internal/graph/dotgraph.go:24:2: use of internal package not allowed
> internal/report/report.go:30:2: use of internal package not allowed
> internal/report/report.go:31:2: use of internal package not allowed
> internal/report/report.go:32:2: use of internal package not allowed
> internal/symbolizer/symbolizer.go:29:2: use of internal package not allowed
> internal/symbolizer/symbolizer.go:30:2: use of internal package not allowed
> internal/symbolizer/symbolizer.go:31:2: use of internal package not allowed
> internal/symbolz/symbolz.go:29:2: use of internal package not allowed

I'll update the pull request with the new tests anyway.

aalexand:
@ErwanLegrand see #279

ErwanLegrand:
@aalexand now it works. Thanks!

ErwanLegrand:
Looking at the tests in `symbolz_test.go`, this looks like I could have taken away the `strings.Contains(url.Path, "/debug/pprof/")` test, after all. What do you think, @aalexand ?

aalexand:
With Go, there are also paths like `/debug/pprof/mutex`, see https://golang.org/pkg/net/http/pprof/, so no. Adding a test for that, or more generically for `/debug/pprof/foo` would be good to be explicit about why that check is still required. Can you add that please?

ErwanLegrand:
Good catch! I have added tests for these and for the URIs listed in https://gperftools.github.io/gperftools/pprof_remote_servers.html

I have also added support for the three listed cases which were not handled. I believe that all the cases are now covered.

internal/symbolz/symbolz.go
```diff
@@ -63,10 +63,29 @@ func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, sy
 	return nil
 }
 
+// Check whether path ends with one of the suffixes listed in
+// pprof_remote_servers.html from the gperftools distribution
+func hasLegacySuffix(path string) bool {
+	var suffixes = [...]string{
```
aalexand: You shouldn't need the `...`.

aalexand: Can be const:
```
const suffixes := []string{
  ...
}
```

internal/symbolz/symbolz.go
```diff
@@ -63,10 +63,29 @@ func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, sy
 	return nil
 }
 
+// Check whether path ends with one of the suffixes listed in
+// pprof_remote_servers.html from the gperftools distribution
+func hasLegacySuffix(path string) bool {
```
aalexand: I am not sure "legacy" is the best name. The remote serving is a feature supported by gperftools, right? Is that really "legacy"? Maybe it should be called "hasGperftoolsSuffix"?

internal/symbolz/symbolz.go
```diff
@@ -63,10 +63,29 @@ func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, sy
...
+		}
+	}
+	return false
+}
+
 // symbolz returns the corresponding symbolz source for a profile URL.
 func symbolz(source string) string {
 	if url, err := url.Parse(source); err == nil && url.Host != "" {
-		if strings.Contains(url.Path, "/debug/pprof/") {
+		// All paths in the net/http/pprof Go package start with /debug/pprof/
```
aalexand: Not necessarily "start" - may rather contain "/debug/pprof" because the code and #249.

internal/symbolz/symbolz.go
ErwanLegrand: The gperftools devs themselves call their version of pprof legacy and recommend switching to this one. Yet, `hasGperftoolsSuffix` would certainly be clearer.

internal/symbolz/symbolz.go
ErwanLegrand: It seems arrays and slices cannot be made const:
```
internal/symbolz/symbolz.go:69:8: const initializer [5]string literal is not a constant
```

ErwanLegrand: Const arrays (and slices) are discussed here: https://github.com/golang/go/issues/6386

ErwanLegrand: It seemed logical to me to make this a fixed size array rather than a slice. I guess this does not matter much in practice, though.

internal/symbolz/symbolz.go
```diff
@@ -65,7 +65,7 @@ func Symbolize(p *profile.Profile, force bool, sources plugin.MappingSources, sy
 
 // Check whether path ends with one of the suffixes listed in
 // pprof_remote_servers.html from the gperftools distribution
-func hasLegacySuffix(path string) bool {
+func hasGperftoolsSuffix(path string) bool {
 	var suffixes = [...]string{
```
aalexand: Using `var suffixes = [...]string{` is not quite idiomatic. `suffixes := []string{` is more so.

CONTRIBUTING.md
```diff
@@ -11,6 +11,45 @@ through the issue tracker with your idea so that we can help out and possibly
 guide you. Coordinating up front makes it much easier to avoid frustration later
 on.
 
+# Development
```
aalexand: This shouldn't be in this PR?

ErwanLegrand: Argh! :-(

Probably I have rebased against the wrong branch at some point.

aalexand: Git can be hard!

ErwanLegrand: Even more so when you have to switch between different git workflows!

PR #284, opened 2017-12-20
Update svgpan, add license, separate pprof-specific code.

aalexand:
The third_party svg directory has been a mix of pprof-specific code and
svgpan itself. This change moves the pprof-specific code out and adds a
license file. It also updates svgpan to 1.2.2 from
https://github.com/aleofreddi/svgpan while we are here. It also renames
the directory to svgpan to have all directories under third_party be
named precisely over the third party package they hold.


aalexand:
FWIW, I don't know why codecov says the coverage is decreased now, should
be the same.

PR #286, opened 2017-12-20
Keep refine menu enabled for views other than graph.

### LINKED ISSUE #285: In the web UI the Refine menu does not enable after a search string is typed

Master at 9cd381157e8842177d38d881690003a328e9c3ba.

Linux

### What did you do?
I run `go run pprof.go -http :8080 profile.pb.gz`. I then enter something like "foo" in the search box.

I expected the Refine menu to enable. It used to work.

### What did you see instead?
The refine menu does not enable.

aalexand:
Pull request #263 disabled the refine menu for any views other than
Graph, but that's not expected - it's perfectly fine to use it for Flame
and Top, so just keep it enabled.

Fixes #285.

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -117,20 +117,11 @@ a {
   font-family: 'Roboto Medium', -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
   position: relative;
 }
-.menu-item.disabled {
-  opacity: 0.5;
-}
 .menu-item .menu-name:hover {
   opacity: 0.75;
 }
 .menu-item .menu-name:hover .downArrow {
   border-top-color: #666;
-}
-.menu-item.disabled .menu-name:hover {
-  opacity: 1;
-}
-.menu-item.disabled .menu-name:hover .downArrow {
-  border-top-color: #ccc;
 }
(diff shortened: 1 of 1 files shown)


tdewolff:
I would add the exceptions to the if statement, instead of removing it entirely. It doesn't work on disasm page for example, so that create confusion.

BTW, regex search doesn't work for me on the flame graph, is that expected?

tdewolff:
Unrelated, but all of a sudden the font for VIEW and REFINE seem very thin for me =/

aalexand:
@tdewolff RE It doesn't work on disasm page for example, so that create confusion.

I think it works. Run something like `PPROF_BINARY_PATH=internal/report/testdata go run pprof.go -http :8080 internal/report/testdata/sample.cpu` then focusing on a `busy` substring as an example displays the assembly limited to a subset of the data (the URL being http://localhost:8080/disasm?f=busy).

@tdewolff RE BTW, regex search doesn't work for me on the flame graph, is that expected?

Focusing on a function like http://localhost:8080/flamegraph?f=busyLoop does work for me in the flame graph displaying only stacks that contain that substring. But interactive search of a function doesn't work for some reason (e.g. typing "busy" and expecting functions with that name to be highlighted), I think it worked but broke. @spiermar Could you take a look?

spiermar:
@aalexand I can have a look. I'm using an event listener in the search field for the highlight function and that might have changed.

aalexand:
@spiermar Thanks - I filed #287.

PR #289, opened 2018-01-08
Cast javascript and css variables to string when using them

nolanmar511:
When third_party files are overridden in other locations, the type of these variables will be []byte.  Casting to string allows these to be used.


nolanmar511:
Not actually necessary.

PR #294, opened 2018-01-17
Clarify error messages in interactive for groups

### LINKED ISSUE #293: Misleading error messages when using bad values with 'granularity' and other options

I'm at e9167955d76d4625c2309f3e02881f4424173440

Linux/amd64

Certain options print a reasonable error message when they're used incorrectly. For example, `sample_index`

```
(pprof) sample_index = 0      # works

(pprof) sample_index = 7      # bad value
sample_index 7 is outside the range [0..3]
```

And with the verbose option name:

```
(pprof) sample_index = alloc_space  # works

(pprof) sample_index = wrong        # bad value
sample_index "wrong" must be one of: [alloc_objects alloc_space inuse_objects inuse_space]
```

Other options, like `granularity`, print a misleading error message when used incorrectly. Example:

```
(pprof) granularity = lines      # works

(pprof) granularity = wrong      # bad value
Unrecognized command: "granularity"
```

The error message says `Unrecognized command: "granularity"` which is wrong and misleading.

nolanmar511:
Fixes #293

### DIFF
--- a/internal/driver/interactive.go
+++ b/internal/driver/interactive.go
@@ -42,16 +42,6 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 	interactiveMode = true
 	shortcuts := profileShortcuts(p)
 
-	// Get all groups in pprofVariables to allow for clearer error messages.
-	groups := make(map[string][]string)
-	for name, option := range pprofVariables {
-		group := option.group
-		if group != "" {
-			groupValues := groups[group]
-			groups[group] = append(groupValues, name)
-		}
-	}
-
 	greetings(p, o.UI)
 	for {
 		input, err := o.UI.ReadLine("(pprof) ")
@@ -63,6 +53,9 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 				return nil
 			}
 		}
+
+		// Get all groups in pprofVariables to allow for clearer error messages.
+		groups := groupOptions()
 
 		for _, input := range shortcuts.expand(input) {
 			// Process assignments of the form variable=value
@@ -99,9 +92,7 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 					continue
 				} else {
 					if okValues := groups[name]; okValues != nil {
-						sort.Strings(okValues)
-						okValuesStr := strings.Join(okValues, ", ")
-						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q. Use one of %s", name, value, okValuesStr))
+						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q. Use one of %s", name, value, strings.Join(okValues, ", ")))
 						continue
 					}
 				}
@@ -133,6 +124,23 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 			}
 		}
 	}
+}
+
+// groupOptions returns an map containing all non-empty groups
+// mapped to an array of the option names in that group in
+// sorted order.
+func groupOptions() map[string][]string {
+	groups := make(map[string][]string)
+	for name, option := range pprofVariables {
+		group := option.group
+		if group != "" {
+			groups[group] = append(groups[group], name)
+		}
+	}
+	for _, names := range groups {
+		sort.Strings(names)
+	}
+	return groups
 }
 
 var generateReportWrapper = generateReport // For testing purposes.
--- a/internal/driver/interactive.go
+++ b/internal/driver/interactive.go
@@ -43,9 +43,16 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 	shortcuts := profileShortcuts(p)
 
 	// Get all groups in pprofVariables to allow for clearer error messages.
-	groups := make(map[string]bool)
-	for _, option := range pprofVariables {
-		groups[option.group] = true
+	groups := make(map[string][]string)
+	for name, option := range pprofVariables {
+		group := option.group
+		if group != "" {
+			if groupValues := groups[group]; groupValues != nil {
+				groups[group] = append(groupValues, name)
+			} else {
+				groups[group] = []string{name}
+			}
+		}
 	}
 
 	greetings(p, o.UI)
@@ -88,13 +95,15 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 					continue
 				}
 				// Allow group=variable syntax by converting into variable="".
-				if groups[name] {
+				if okValues := groups[name]; okValues != nil {
 					if v := pprofVariables[value]; v != nil && v.group == name {
 						if err := pprofVariables.set(value, ""); err != nil {
 							o.UI.PrintErr(err)
 						}
 					} else {
-						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q", name, value))
+						sort.Strings(okValues)
+						recognizedValues := strings.Join(okValues, ", ")
+						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q. Recognized values are %s", name, value, recognizedValues))
 					}
 					continue
 				}
--- a/internal/driver/interactive.go
+++ b/internal/driver/interactive.go
@@ -90,11 +90,9 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 						o.UI.PrintErr(err)
 					}
 					continue
-				} else {
-					if okValues := groups[name]; okValues != nil {
-						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q. Use one of %s", name, value, strings.Join(okValues, ", ")))
-						continue
-					}
+				} else if okValues := groups[name]; okValues != nil {
+					o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q. Use one of %s", name, value, strings.Join(okValues, ", ")))
+					continue
 				}
 			}
 
--- a/internal/driver/interactive.go
+++ b/internal/driver/interactive.go
@@ -42,6 +42,9 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 	interactiveMode = true
 	shortcuts := profileShortcuts(p)
 
+	// Get all groups in pprofVariables to allow for clearer error messages.
+	groups := groupOptions(pprofVariables)
+
 	greetings(p, o.UI)
 	for {
 		input, err := o.UI.ReadLine("(pprof) ")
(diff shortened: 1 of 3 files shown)


internal/driver/interactive_test.go
```diff
@@ -151,9 +162,10 @@ func newUI(t *testing.T, input []string) plugin.UI {
 }
 
 type testUI struct {
-	t     *testing.T
-	input []string
-	index int
+	t       *testing.T
+	input   []string
+	index   int
+	wantErr string
```
aalexand: A similar functionality exists in TestUI in proftest.go, but that lacks the ability to simulate the console input. Would it be better to extend that with the capability to simulate the input and reuse it here?

internal/driver/interactive.go
```diff
@@ -82,9 +88,13 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
...
-				if v := pprofVariables[value]; v != nil && v.group == name {
-					if err := pprofVariables.set(value, ""); err != nil {
-						o.UI.PrintErr(err)
+				if groups[name] {
+					if v := pprofVariables[value]; v != nil && v.group == name {
+						if err := pprofVariables.set(value, ""); err != nil {
+							o.UI.PrintErr(err)
+						}
+					} else {
+						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q", name, value))
```
aalexand: It might be good to print out the list of values that are recognized?

internal/driver/interactive_test.go
nolanmar511: Done.

internal/driver/interactive.go
nolanmar511: Added that.

nolanmar511:
PTAL

internal/driver/interactive.go
```diff
@@ -82,9 +95,15 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 					continue
 				}
 				// Allow group=variable syntax by converting into variable="".
-				if v := pprofVariables[value]; v != nil && v.group == name {
-					if err := pprofVariables.set(value, ""); err != nil {
-						o.UI.PrintErr(err)
+				if okValues := groups[name]; okValues != nil {
```
aalexand: Why is this check required?

internal/driver/interactive.go
```diff
@@ -82,9 +95,15 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
...
+					if v := pprofVariables[value]; v != nil && v.group == name {
+						if err := pprofVariables.set(value, ""); err != nil {
+							o.UI.PrintErr(err)
+						}
+					} else {
+						sort.Strings(okValues)
+						recognizedValues := strings.Join(okValues, ", ")
+						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q. Recognized values are %s", name, value, recognizedValues))
```
aalexand: ". Recognized values are" -> ", use one of", to avoid repeating longish "recognized" stem in the sentence.

internal/driver/interactive.go
aalexand: Or why it's not inside the else branch below?

internal/driver/interactive.go
```diff
@@ -82,9 +95,15 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
...
-					if err := pprofVariables.set(value, ""); err != nil {
-						o.UI.PrintErr(err)
+				if okValues := groups[name]; okValues != nil {
+					if v := pprofVariables[value]; v != nil && v.group == name {
+						if err := pprofVariables.set(value, ""); err != nil {
+							o.UI.PrintErr(err)
+						}
+					} else {
+						sort.Strings(okValues)
+						recognizedValues := strings.Join(okValues, ", ")
```
aalexand: Nit: the change of terminology from "ok" to "recognized" stands out a bit, I'd get it consistent.

internal/driver/interactive_test.go
```diff
@@ -70,6 +70,21 @@ func TestShell(t *testing.T) {
...
+		Input:   []string{"cumulative=this"},
+		AllowRx: `Unrecognized value for cumulative: "this". Recognized values are cum, flat`,
+	}
+	o.UI = ui
+	if err := interactive(p, o); err != nil {
+		t.Error("invalid group value:", err)
+	}
+	// Confirm error message written out once.
+	if ui.NumAllowRxMatches != 1 {
+		t.Errorf("want error message to be longed 1 time, got %v", ui.NumAllowRxMatches)
```
aalexand: "longed" -> "logged", or "printed".

internal/proftest/proftest.go
```diff
@@ -80,11 +81,21 @@ type TestUI struct {
...
 func (ui *TestUI) ReadLine(_ string) (string, error) {
-	return "", fmt.Errorf("no input")
+	if ui.index >= len(ui.Input) {
+		return "", io.EOF
+	}
+	input := ui.Input[ui.index]
+	if input == "**error**" {
+		return "", fmt.Errorf("Error: %s", input)
+	}
+	ui.index++
```
aalexand: This is probably a no-op but it's a bit strange that the index is not increased in case of the error simulation input.

internal/driver/interactive.go
```diff
@@ -42,6 +42,19 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 	interactiveMode = true
 	shortcuts := profileShortcuts(p)
 
+	// Get all groups in pprofVariables to allow for clearer error messages.
+	groups := make(map[string][]string)
+	for name, option := range pprofVariables {
+		group := option.group
+		if group != "" {
+			if groupValues := groups[group]; groupValues != nil {
```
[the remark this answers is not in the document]
nolanmar511: Done.

internal/driver/interactive.go
nolanmar511: Done.

internal/driver/interactive.go
nolanmar511: Done -- using recognized.

internal/driver/interactive.go
nolanmar511: Done.

internal/driver/interactive_test.go
nolanmar511: Thanks! -- done.

internal/proftest/proftest.go
nolanmar511: I agree -- changing that.

nolanmar511:
PTAL

internal/driver/interactive.go
```diff
@@ -42,6 +42,16 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 	interactiveMode = true
 	shortcuts := profileShortcuts(p)
 
+	// Get all groups in pprofVariables to allow for clearer error messages.
+	groups := make(map[string][]string)
+	for name, option := range pprofVariables {
+		group := option.group
+		if group != "" {
+			groupValues := groups[group]
```
aalexand: I think you can get rid of this local variable, it's only used once below.

internal/driver/interactive.go
```diff
@@ -42,6 +42,16 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 	interactiveMode = true
 	shortcuts := profileShortcuts(p)
 
+	// Get all groups in pprofVariables to allow for clearer error messages.
+	groups := make(map[string][]string)
```
aalexand: It might be better to make this a separate function like
```
func (v *variable) groupOptions() map[string][]string {
```
in commands.go, as this `interactive()` function got a bit big. Probably also push the sorting down as well so that this function returns sorted list of option names in each named group.

internal/driver/interactive.go
```diff
@@ -87,6 +97,13 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 						o.UI.PrintErr(err)
 					}
 					continue
+				} else {
+					if okValues := groups[name]; okValues != nil {
+						sort.Strings(okValues)
+						okValuesStr := strings.Join(okValues, ", ")
+						o.UI.PrintErr(fmt.Errorf("Unrecognized value for %s: %q. Use one of %s", name, value, okValuesStr))
```
aalexand: Nit: maybe just use `strings.Join(okValues, ", ")` instead of okValuesStr here, it's not much longer.

internal/driver/interactive.go
nolanmar511: Done.

internal/driver/interactive.go
nolanmar511: Done.

internal/driver/interactive.go
nolanmar511: Done.

internal/driver/interactive.go
```diff
@@ -64,6 +54,9 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 			}
 		}
 
+		// Get all groups in pprofVariables to allow for clearer error messages.
+		groups := groupOptions()
```
aalexand: Move this out of the for loop to avoid calling it on every input line?

internal/driver/interactive.go
```diff
@@ -135,6 +126,23 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 	}
 }
 
+// groupOptions returns an map containing all non-empty groups
```
aalexand: Nit: "a map"

internal/driver/interactive.go
```diff
@@ -135,6 +126,23 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
...
+// mapped to an array of the option names in that group in
+// sorted order.
+func groupOptions() map[string][]string {
```
aalexand: I'd make this a function on the variables type like I mentioned in the original comment about this, but this is not too important.

internal/driver/interactive.go
nolanmar511: Done.

internal/driver/interactive.go
nolanmar511: Done.

internal/driver/interactive.go
nolanmar511: Done.

nolanmar511:
PTAL

internal/driver/interactive.go
```diff
@@ -87,6 +90,11 @@ func interactive(p *profile.Profile, o *plugin.Options) error {
 						o.UI.PrintErr(err)
 					}
 					continue
+				} else {
+					if okValues := groups[name]; okValues != nil {
```
aalexand: Merge to the previous line as "else if"?

PR #296, opened 2018-01-23
Support parsing of is_folded bit

nolanmar511:
Modified pprof to keep is_folded bit during encoding and decoding, factor in is_folded bit during merges, and print is_folded bit when converting profile to string.

### DIFF
--- a/profile/encode.go
+++ b/profile/encode.go
@@ -494,7 +494,7 @@ var locationDecoder = []decoder{
 		pp.Line = append(pp.Line, Line{})
 		return decodeMessage(b, &pp.Line[n])
 	},
-	func(b *buffer, m message) error { return decodeBool(b, &m.(*Location).IsFolded) },
+	func(b *buffer, m message) error { return decodeBool(b, &m.(*Location).IsFolded) }, // optional bool is_folded = 5;
 }
 
 func (p *Line) decoder() []decoder {
--- a/profile/encode.go
+++ b/profile/encode.go
@@ -480,6 +480,7 @@ func (p *Location) encode(b *buffer) {
 	for i := range p.Line {
 		encodeMessage(b, 4, &p.Line[i])
 	}
+	encodeBoolOpt(b, 5, p.IsFolded)
 }
 
 var locationDecoder = []decoder{
@@ -493,6 +494,7 @@ var locationDecoder = []decoder{
 		pp.Line = append(pp.Line, Line{})
 		return decodeMessage(b, &pp.Line[n])
 	},
+	func(b *buffer, m message) error { return decodeBool(b, &m.(*Location).IsFolded) },
 }
 
 func (p *Line) decoder() []decoder {
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -592,6 +592,9 @@ func (l *Location) string() string {
 	if m := l.Mapping; m != nil {
 		locStr = locStr + fmt.Sprintf("M=%d ", m.ID)
 	}
+	if l.IsFolded {
+		locStr = locStr + "(folded) "
+	}
 	if len(l.Line) == 0 {
 		ss = append(ss, locStr)
 	}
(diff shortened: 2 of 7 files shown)


nolanmar511:
Realized I did not yet account for the is_folded bit in merges, which needs to be done. Will re-open once I do that.

proto/profile.proto
```diff
@@ -113,7 +118,7 @@ message Label {
   // Should only be present when num is present.
   // Specifies the units of num.
   // Use arbitrary string (for example, "requests") as a custom count unit.
-  // If no unit is specified, consumer may apply heuristic to deduce the unit.
+  // If no unit is specificed, consumer may apply heuristic to deduce the unit.
```
aalexand: The change looks like it introduces a typo.

profile/profile.go
```diff
@@ -591,6 +592,9 @@ func (l *Location) string() string {
 	if m := l.Mapping; m != nil {
 		locStr = locStr + fmt.Sprintf("M=%d ", m.ID)
 	}
+	if l.IsFolded {
+		locStr = locStr + "(folded) "
```
aalexand: What does the full output look like with the update? One thing I wondered about is whether we should use something like "[F]" instead of "(folded)" to sort of match how the function-level bits are printed out, see code above. But it's hard to say for sure without looking at the output, hence asking.

profile/encode.go
```diff
@@ -493,6 +494,7 @@ var locationDecoder = []decoder{
 		pp.Line = append(pp.Line, Line{})
 		return decodeMessage(b, &pp.Line[n])
 	},
+	func(b *buffer, m message) error { return decodeBool(b, &m.(*Location).IsFolded) },
```
aalexand: Add a comment, similar to other fields above?

profile/profile.go
```diff
@@ -109,10 +109,11 @@ type Mapping struct {
...
 type Location struct {
-	ID      uint64
-	Mapping *Mapping
-	Address uint64
-	Line    []Line
+	ID       uint64
+	Mapping  *Mapping
+	Address  uint64
+	Line     []Line
+	IsFolded bool
```
aalexand: One thing to check is that this flag should be reset when Line array is reset in the symbolization scenarios.

profile/profile.go
nolanmar511: Currently looks like this
```
  2000: 0x2000 M=1 (folded) foo foo.c:2 s=0
             foo_caller foo.c:1 s=0
```

I think [F] would look better, so changing.

proto/profile.proto
nolanmar511: Fixing here and separately in profile.proto

profile/encode.go
nolanmar511: Done.

profile/profile.go
nolanmar511: I think I took care of this in internal/symbolizer/symbolizer.go, but let me know if my understanding here is wrong.

profile/profile_test.go
```diff
@@ -275,6 +275,53 @@ var cpuL = []*Location{
 	},
 }
 
+var cpuLFolded = []*Location{
```
aalexand: This seems to mostly duplicate `cpuL` profile from above setting IsFolded flag? Perhaps we could generate one profile from the other? For example, set IsFolded on some locations in `cpuL` in the test?

Or actually even produce testProfile1Folded from testProfile1 by massaging some of its locations to have the folded flag? This would save a lot of duplication unless the profiles you got are not actually exact copies of each other modulo the folded flag.

nolanmar511: The only tricky thing is setting the Location field in samples so samples reference the folded locations.

aalexand: The profile copying is deep, so I think it should work fine.

nolanmar511: Yep -- this works. Done!

PR #304, opened 2018-01-29
Detect concatenated profile and give clearer error message

### LINKED ISSUE #273: profile: detect and reject concatenated profiles
golang/go#22036 reported a simple problem with a mysterious result. 437 concatenated pprof profiles were passed to pprof as one file and produced the error:

	malformed profile: mismatch: sample has: 4 values vs. 1748 types

Concatenated profiles are certainly not supported today, and I can't really see that changing given the details of the profile format.

A change like the below seems like it would produce a better error.

Will leave the actual change to you.

```
$ git diff
diff --git a/src/cmd/vendor/github.com/google/pprof/profile/encode.go b/src/cmd/vendor/github.com/google/pprof/profile/encode.go
index 622319484a..9453fa19cd 100644
--- a/src/cmd/vendor/github.com/google/pprof/profile/encode.go
+++ b/src/cmd/vendor/github.com/google/pprof/profile/encode.go
@@ -214,7 +214,13 @@ var profileDecoder = []decoder{
 	// int64 keep_frames = 8
 	func(b *buffer, m message) error { return decodeInt64(b, &m.(*Profile).keepFramesX) },
 	// int64 time_nanos = 9
-	func(b *buffer, m message) error { return decodeInt64(b, &m.(*Profile).TimeNanos) },
+	func(b *buffer, m message) error {
+		p := m.(*Profile)
+		if p.TimeNanos != 0 {
+			return errors.New("concatenated pprof profiles detected")
+		}
+		return decodeInt64(b, &m.(*Profile).TimeNanos)
+	},
 	// int64 duration_nanos = 10
 	func(b *buffer, m message) error { return decodeInt64(b, &m.(*Profile).DurationNanos) },
 	// ValueType period_type = 11
$ 
```

nolanmar511:
This change detects concatenated profiles by checking if time_nanos appears more than once in the profile being parsed. 

Fixes #273

### DIFF
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -164,7 +164,7 @@ func ParseData(data []byte) (*Profile, error) {
 			return nil, fmt.Errorf("decompressing profile: %v", err)
 		}
 	}
-	if p, err = ParseUncompressed(data); err != nil && err != errNoData && err != ErrConcatProfile {
+	if p, err = ParseUncompressed(data); err != nil && err != errNoData && err != errConcatProfile {
 		p, err = parseLegacy(data)
 	}
 
@@ -181,9 +181,7 @@ func ParseData(data []byte) (*Profile, error) {
 var errUnrecognized = fmt.Errorf("unrecognized profile format")
 var errMalformed = fmt.Errorf("malformed profile format")
 var errNoData = fmt.Errorf("empty input file")
-
-// ErrConcatProfile is the error returned when a concatenated profile is detected.
-var ErrConcatProfile = fmt.Errorf("concatenated profiles detected")
+var errConcatProfile = fmt.Errorf("concatenated profiles detected")
 
 func parseLegacy(data []byte) (*Profile, error) {
 	parsers := []func([]byte) (*Profile, error){
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -164,7 +164,7 @@ func ParseData(data []byte) (*Profile, error) {
 			return nil, fmt.Errorf("decompressing profile: %v", err)
 		}
 	}
-	if p, err = ParseUncompressed(data); err != nil && err != errNoData {
+	if p, err = ParseUncompressed(data); err != nil && err != errNoData && err != ErrConcatProfile {
 		p, err = parseLegacy(data)
 	}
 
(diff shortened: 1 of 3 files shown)


profile/profile.go
```diff
@@ -182,6 +182,9 @@ var errUnrecognized = fmt.Errorf("unrecognized profile format")
 var errMalformed = fmt.Errorf("malformed profile format")
 var errNoData = fmt.Errorf("empty input file")
 
+// ErrConcatProfile is the error returned when a concatenated profile is detected.
+var ErrConcatProfile = fmt.Errorf("concatenated profiles detected")
```
aalexand: I think this can be package-private.

profile/profile_test.go
```diff
@@ -108,6 +108,22 @@ func TestParseError(t *testing.T) {
 	}
 }
 
+func TestParseConcatentated(t *testing.T) {
+	prof := testProfile1.Copy()
+	prof.TimeNanos = 10000
```
aalexand: Perhaps just add this to the testProfile1 initialization? In general every profile should have TimeNanos set, so it's a bit strange to just have it initialized here.

profile/profile.go
nolanmar511: Yep -- fixed.

profile/profile_test.go
nolanmar511: Done.

profile/profile_test.go
```diff
@@ -108,6 +108,21 @@ func TestParseError(t *testing.T) {
 	}
 }
 
+func TestParseConcatentated(t *testing.T) {
+	prof := testProfile1.Copy()
+	// Write the profile twice to buffer to create concatented profile.
+	buf := bytes.NewBuffer(nil)
```
aalexand: Nit: just "var buf bytes.Buffer"

nolanmar511: Tests failed after I'd made this change. I saw the following error:
```
panic: runtime error: invalid memory address or nil pointer dereference
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x50a819]

```

I think I might need to keep this as is.

aalexand: @nolanmar511 #312 seems to pass tests for me so not sure what was the issue you saw. Unless you did something like `var buf *bytes.Buffer` which would of course would crash because it does not really create a buffer var.

nolanmar511: Yeah -- that was probably my problem.

PR #305, opened 2018-01-30
Set HOME to a temporary directory in TestHttpsInsecure

mwhudson:
Debian builds packages with HOME set to a bogus value and this causes tests to
fail when packaging Go 1.10...

### DIFF
--- a/internal/driver/fetch_test.go
+++ b/internal/driver/fetch_test.go
@@ -362,6 +362,14 @@ func TestHttpsInsecure(t *testing.T) {
 	if runtime.GOOS == "nacl" {
 		t.Skip("test assumes tcp available")
 	}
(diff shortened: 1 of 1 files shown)


internal/driver/fetch_test.go
```diff
@@ -362,6 +362,14 @@ func TestHttpsInsecure(t *testing.T) {
 	if runtime.GOOS == "nacl" {
 		t.Skip("test assumes tcp available")
 	}
+	saveHome := os.Getenv(homeEnv())
```
aalexand: Please add a comment on why this is done here.

mwhudson: Done, but as this has been merged already this requires another PR I guess...

aalexand:
@eliasnaur Yes, sent #308.

Issue #311, opened 2018-02-05
Fails with shared libraries because mapping is ignored on OSX


Timmmm:
See [this issue](https://github.com/gperftools/gperftools/issues/956) for background and test case.

Basically on OSX pprof does not subtract the mapped base address of shared libraries from code addresses before feeding them to `llvm-symbolizer`.

I don't know exactly how it is supposed to work, but replacing [this line](https://github.com/google/pprof/blob/master/internal/symbolizer/symbolizer.go#L164) with this seems to make it work:

```
		addr := l.Address
		if l.Mapping != nil {
			addr -= l.Mapping.Start
		}

		stack, err := segment.SourceLine(addr)
```

Also, for some very strange reason, with this patch, `-web`, `-top` and `-list '.*'` work but `-weblist '.*'` does not. However `-weblist 'factorial'` *does* work. It seems like the web interface trips up on the first `??` output:

```
$ pprof -list ".*" out.prof
Local symbolization failed for libdyld.dylib: unrecognized binary: /usr/lib/system/libdyld.dylib
Some binary filenames not available. Symbolization may be incomplete.
Try setting PPROF_BINARY_PATH to the search path for local binaries.
Total: 690ms
ROUTINE ========================  in ??
         0      690ms (flat, cum)   100% of Total
 Error: Could not find file ?? on path /Users/me/test
ROUTINE ======================== factorial in /Users/me/test/lib.cpp
     690ms      690ms (flat, cum)   100% of Total
         .          .      3:    long long int result=m_nValue;
         .          .      4:    long long int result_next;
         .          .      5:    long long int pc = m_nValue;
         .          .      6:    do
         .          .      7:    {
     310ms      310ms      8:        result_next = result*(pc-1);
     250ms      250ms      9:        result = result_next;
      60ms       60ms     10:        pc--;
      70ms       70ms     11:    } while(pc>2);
         .          .     12:    m_nValue = result;
         .          .     13:    return m_nValue;
         .          .     14:}
         .          .     15:
         .          .     16:int foo() {
```

aalexand:
Huh, thanks a lot for tracking this down, this makes a lot of sense. I think your fix will also help a number of cases people had in #130 - some of symbolization issues mentioned there are with dynamic libraries and some could have been because of executable's ASLR.

PR #313, opened 2018-02-06
Set base address on OSX

Timmmm:
This fixes issue #311. The mapped address of libraries was never considered, nor was the load address of the segments.

### DIFF
--- a/internal/binutils/binutils.go
+++ b/internal/binutils/binutils.go
@@ -172,6 +172,21 @@ func (bu *Binutils) Open(name string, start, limit, offset uint64) (plugin.ObjFi
 	return nil, fmt.Errorf("unrecognized binary: %s", name)
 }
 
+// Base address of the __TEXT section. Usually 0 for shared libraries
+// of 0x100000000 for executables. You can check this value by
+// running `objdump -private-headers <file>`.
+func getTextAddress(of *macho.File) uint64 {
+	for _, ld := range of.Loads {
+		switch s := ld.(type) {
+		case *macho.Segment:
+			if s.Name == "__TEXT" {
+				return s.Addr
+			}
+		}
+	}
+	return 0
+}
+
 func (b *binrep) openMachO(name string, start, limit, offset uint64) (plugin.ObjFile, error) {
 	of, err := macho.Open(name)
 	if err != nil {
@@ -179,10 +194,12 @@ func (b *binrep) openMachO(name string, start, limit, offset uint64) (plugin.Obj
 	}
 	defer of.Close()
 
+	base := start - getTextAddress(of)
+
 	if b.fast || (!b.addr2lineFound && !b.llvmSymbolizerFound) {
-		return &fileNM{file: file{b: b, name: name}}, nil
+		return &fileNM{file: file{b: b, name: name, base: base}}, nil
 	}
-	return &fileAddr2Line{file: file{b: b, name: name}}, nil
+	return &fileAddr2Line{file: file{b: b, name: name, base: base}}, nil
 }
 
 func (b *binrep) openELF(name string, start, limit, offset uint64) (plugin.ObjFile, error) {


aalexand:
I can accept the fix as is because it's an improvement we need, but how about a test?

Timmmm:
Sure. What should it test exactly though? I could throw a small mach-o library and binary in there to check `getTextAddress()` works?

aalexand:
Yes, a test similar to the ones in https://github.com/google/pprof/blob/master/internal/binutils/binutils_test.go against a canned binary from testdata. The test will need to be skipped on anything non-Mac.

Timmmm:
I added a test. The test actually found some bugs elsewhere in the code. Also I'm not 100% sure about the intent of some of the test I copied it from. Anyway the test passes and I guess it was useful as it contains bug fixes.

internal/binutils/binutils.go
```diff
@@ -179,10 +179,20 @@ func (b *binrep) openMachO(name string, start, limit, offset uint64) (plugin.Obj
 	}
 	defer of.Close()
 
+	base := start
```
aalexand: I am a bit hesitant whether a better value for the base in the absence of __TEXT is zero or `start`. But I guess it's not important since files without __TEXT should be impractical case.

internal/binutils/binutils_test.go
```diff
@@ -255,6 +261,73 @@ func TestObjFile(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Open: unexpected error %v", err)
+			}
+			defer f.Close()
+			syms, err := f.Symbols(nil, 0)
+			if err != nil {
+				t.Fatalf("Symbols: unexpected error %v", err)
+			}
+
+			find := func(name string) *plugin.Sym {
```
aalexand: Nit: could move this piece out to an outer function to reuse with the other test.

internal/binutils/disasm.go
```diff
@@ -62,7 +69,7 @@ func matchSymbol(names []string, start, end uint64, r *regexp.Regexp, address ui
 		return names
 	}
 	for _, name := range names {
-		if r.MatchString(name) {
+		if r == nil || r.MatchString(name) {
```
aalexand: Do we pass r as nil from anywhere? Ah, you do know in `f.Symbols(nil, 0)` in the test. I think you can just do `f.Symbols(regexp.MustCompile(tc.sym), 0)`?.

internal/binutils/disasm.go
```diff
@@ -37,6 +37,7 @@ func findSymbols(syms []byte, file string, r *regexp.Regexp, address uint64) ([]
 	var symbols []*plugin.Sym
 	names, start := []string{}, uint64(0)
 	buf := bytes.NewBuffer(syms)
+	symAddr := uint64(0)
```
aalexand: I think this symAddr will get shadowed by the assignment in the loop header, so it's probably not what the code after the loop expects.

aalexand:
@Timmmm Huh, Mac tests fail on Travis with

```
--- FAIL: TestMachoFiles (0.62s)
    --- FAIL: TestMachoFiles/normal_mapping (0.57s)
    	binutils_test.go:325: SourceLine for main: got [{_main  0}]; want [{main /tmp/hello.c 3}]
    --- FAIL: TestMachoFiles/other_mapping (0.02s)
    	binutils_test.go:325: SourceLine for main: got [{_main  0}]; want [{main /tmp/hello.c 3}]
    --- FAIL: TestMachoFiles/lib_normal_mapping (0.03s)
    	binutils_test.go:325: SourceLine for main: got [{_bar  0}]; want [{bar /tmp/lib.c 6}]
```

internal/binutils/binutils.go
Timmmm: Yeah I'll add a warning if it isn't found.

internal/binutils/binutils_test.go
Timmmm: Will do.

internal/binutils/disasm.go
Timmmm: Well spotted!

internal/binutils/disasm.go
Timmmm: No but [the comments for the `Symbols()` interface](https://github.com/google/pprof/blob/43bdac25ec8dcf95b98bba82e7a162ff406a20b3/internal/plugin/plugin.go#L149) say it can be nil.

aalexand:
Hoisting out this question from the last round of comments because those top-level Github comments are easy to miss so not sure you saw it:

Do files internal/binutils/debug.test, internal/binutils/testdata/exe_mac_64.dSYM/Contents/Info.plist, internal/binutils/testdata/exe_mac_64.dSYM/Contents/Resources/DWARF/exe_mac_64, internal/binutils/testdata/lib_mac_64.dSYM/Contents/Info.plist, internal/binutils/testdata/lib_mac_64.dSYM/Contents/Resources/DWARF/lib_mac_64 need to be part of the PR?

Timmmm:
Ah yeah I missed that, and yes they do, except `debug.test`. Not sure where that came from. I can't work out why the travis test fails. I just downloaded it on a totally separate mac and the test passes for me. I wonder if Travis uses `addr2line` instead of `llvm-symbolizer` and there is some difference there?

Timmmm:
Yeah that seems to be it. Removing `llvm-symbolizer` and installing binutils (for `gaddr2line`) gives the exact same failure as above.

However `gaddr2line` does seem to work when I run it manually, so I guess there is *another* bug in `pprof` around that. I'll look into it tomorrow. I guess you guys don't use Macs!

aalexand:
@Timmmm A lot of development here happens on Linux. I carry a Mac laptop but a lot of times I'm just in an SSH window to my Linux machine. And when I do use it for actual coding, that's Go and I'm not hitting the issues you've found because the Golang profiles are pre-symbolized by the runtime.

Thanks for figuring many things out here!

Timmmm:
I've updated it as requested. The test will still fail because `pprof` because of a number of issues:

1. `pprof` only looks for `addr2line` but on my system at least it is called `gaddr2line`. No idea why. Probably a stupid reason. It should search for both anyway. I wasn't sure how you would want to handle that so I'll leave it to you.
2. When using `addr2line` there is [some code that attempts to "improve" the symbol names](https://github.com/google/pprof/blob/master/internal/binutils/addr2liner.go#L218) using `nm`. Unfortunately the names `nm` returns are mangled (it is executed as `nm -n` to prevent demangling), so it finds `_main` instead of `main` and the test fails.
3. If there is no `addr2line` or `llvm-symbolizer` the test will clearly fail, but I guess that case is ok because the test environment should have one of those.

I'd suggest the easiest thing for now is to install `llvm-symbolizer` on the Mac test environment so the test passes, and open issues for the other things. I can't really spend any more time on this sorry!

Timmmm:
Oh yeah I also forgot to ask - is the logic behind `matchSymbol()` actually correct? If the symbol is matched by address, all of the names are returned, but for regex matching only the first one is. Seems suspicious to me.

internal/binutils/binutils.go
```diff
@@ -179,10 +179,25 @@ func (b *binrep) openMachO(name string, start, limit, offset uint64) (plugin.Obj
 	}
 	defer of.Close()
 
+	// Subtract the load address of the __TEXT section. Usually 0 for shared
+	// libraries or 0x100000000 for executables. You can check this value by
+	// running `objdump -private-headers <file>`.
+
+	textSegment := of.Segment("__TEXT")
+	if textSegment == nil {
+		return nil, fmt.Errorf("Parsing %s: no __TEXT segment", name)
```
aalexand: Error messages should start with lowercase: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Suggesting `fmt.Errorf("not a mach executable: no __TEXT segment: %s", name)`, to be somewhat consistent with `fmt.Errorf("unrecognized binary: %s", name)` above in this file.

internal/binutils/binutils.go
```diff
@@ -179,10 +179,25 @@ func (b *binrep) openMachO(name string, start, limit, offset uint64) (plugin.Obj
...
+	}
+	if textSegment.Addr > start {
+		return nil, fmt.Errorf("Parsing %s: __TEXT segment address (0x%x) > mapping start address (0x%x)",
```
aalexand: Error messages should start with lowercase: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Suggesting `fmt.Errorf("not a mach executable: __TEXT segment address (0x%x) > mapping start address (0x%x): %s", testSegment.Addr, start, name)`.

By the way, could `base` be negative in case of some exotic executable ASLR? Probably not, I think in reality for PIE executables textSegment address is zero, at least that's what I saw on Linux.

internal/binutils/disasm.go
```diff
@@ -34,24 +34,48 @@ var (
...
+	// The current group of symbol names, and the address they are all at.
 	names, start := []string{}, uint64(0)
+
 	buf := bytes.NewBuffer(syms)
-	for symAddr, name, err := nextSymbol(buf); err == nil; symAddr, name, err = nextSymbol(buf) {
+
+	for {
+		symAddr, name, err := nextSymbol(buf)
+		if err == io.EOF {
+			// Done! If there was an unfinished group, append it.
```
aalexand: Nit / opiniated: remove the exclamation mark, those are overly emotional for comments and log messages to my taste...

internal/binutils/testdata/lib_mac_64.dSYM/Contents/Info.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
aalexand: Reading https://stackoverflow.com/questions/8058497/can-clang-build-executables-with-dsym-debug-information, it appears to me that building the test binaries from the command line it should be possible to have the debug information as part of those, and these additional dSYM files and directory would not be needed. I can look into doing that myself, can you include the source code of the executable and library test binaries that you used?

aalexand:
@Timm I left some comments, those are minor, please address.

I apologize it took so much longer than I thought it would. I'd probably suggest that you leave the test code in as is, but make the mac test skipped permanently, and I'll look into enabling that.

For `matchSymbol` behavior - that code pre-dates me on the project, I found the behavior a bit unexpected as well, it should be at least documented better, will look at it later.

internal/binutils/binutils.go
Timmmm: I just copied the one above but ok.

internal/binutils/binutils.go
Timmmm: Yeah I wasn't sure about that but even if it is you'd probably want to know about it - i.e. get a bug report here, rather than just having it underflow and fail silently (or get lucky with the unsigned arithmetic).

Timmmm:
Ok I fixed some of the nits (getting a little extreme!) and lowercased other error messages. I didn't use your exact error message suggestions because they still *are* mach files - and there was already an error message for failing to find the base address for ELF so I copied that.

The source I used was:

`/tmp/hello.c`

```
#include <stdio.h>

int main() {
	printf("Hello, world!\n");
	return 0;
}
```

`/tmp/lib.c`

```
int foo() {

	return 1;	
}

int bar() {

	return 2;
}
```

Compiled with `clang -g /tmp/hello.c -o exe_mac_64` and `clang -g /tmp/lib.c -dynamiclib -o lib_mac_64` (which should generate the dSYM's in both cases).

Issue #300, opened 2018-01-29
internal/driver: test timed out after 3m0s on freebsd, openbsd, windows


mvdan:
Encountered in https://go-review.googlesource.com/c/go/+/89715, after trying to update the vendored copy to pull https://github.com/google/pprof/pull/298.

https://storage.googleapis.com/go-build-log/b9cbc869/freebsd-amd64-11_1_dbe018c1.log

Relevant bit of output:

```
Generating report in /tmp/profile_proto401663026
Generating report in /tmp/profile_output286659561
Generating report in /tmp/profile_proto375996468
Generating report in /tmp/profile_output864397315
Generating report in /tmp/profile_proto039962502
Generating report in /tmp/profile_output187100461
Generating report in /tmp/profile_proto409183912
Ignoring local file /path/to/buildid: build-id mismatch (buildid != abcdef)
Generating report in /tmp/profile_output785726951
Generating report in /tmp/profile_proto884247066
Ignoring local file /path/to/buildid: build-id mismatch (buildid != abcdef)
Generating report in /tmp/profile_output637371569
Generating report in /tmp/profile_proto247173724
Ignoring local file /path/to/buildid: build-id mismatch (buildid != abcdef)
Generating report in /tmp/profile_output097694731
Generating report in /tmp/profile_proto492795374
Ignoring local file /path/to/buildid: build-id mismatch (buildid != abcdef)
Generating report in /tmp/profile_output034252917
Generating report in /tmp/profile_proto368253776
Generating report in /tmp/profile_output322350703
Generating report in /tmp/profile_proto977961218
Generating report in /tmp/profile_output894972025
Generating report in /tmp/profile_proto420701572
Generating report in /tmp/profile_output650228499
Generating report in /tmp/profile_proto732512086
Hide expression matched no samples
Generating report in /tmp/profile_output521628349
Generating report in /tmp/profile_proto277399288
Ignoring local file /path/to/buildid: build-id mismatch (buildid != abcdef)
Generating report in /tmp/profile_output057706487
Generating report in /tmp/profile_proto464104682
Ignoring local file /path/to/buildid: build-id mismatch (buildid != abcdef)
Generating report in /tmp/profile_output775549761
Generating report in /tmp/profile_proto674043820
Generating report in /tmp/profile_output590092059
Generating report in /tmp/profile_proto546088894
Generating report in /tmp/profile_output341598725
Generating report in /tmp/profile_proto335752608
[...]

PR #315, opened 2018-02-07
Read response from http.Get() in TestWebInterface, so http.Get() does not hang

nolanmar511:
This fixes test failure described in #300.

The response from http.Get() was not previously read, which caused http.Get() to hang forever.

### DIFF
--- a/internal/driver/webui_test.go
+++ b/internal/driver/webui_test.go
@@ -119,7 +119,10 @@ func TestWebInterface(t *testing.T) {
 		for count := 0; count < 2; count++ {
 			wg.Add(1)
 			go func() {
-				http.Get(path)
+				res, err := http.Get(path)
+				if err == nil {
+					ioutil.ReadAll(res.Body)
+				}
 				wg.Done()
 			}()
 		}
--- a/internal/driver/webui_test.go
+++ b/internal/driver/webui_test.go
@@ -120,8 +120,13 @@ func TestWebInterface(t *testing.T) {
 			wg.Add(1)
 			go func() {
 				res, err := http.Get(path)
-				if err == nil {
-					ioutil.ReadAll(res.Body)
+				if err != nil {
+					t.Error("could not fetch", c.path, err)
+					wg.Done()
+					return
+				}
+				if _, err = ioutil.ReadAll(res.Body); err != nil {
+					t.Error("could not read response", c.path, err)
 				}
 				wg.Done()
 			}()
--- a/internal/driver/webui_test.go
+++ b/internal/driver/webui_test.go
@@ -119,16 +119,15 @@ func TestWebInterface(t *testing.T) {
 		for count := 0; count < 2; count++ {
 			wg.Add(1)
 			go func() {
+				defer wg.Done()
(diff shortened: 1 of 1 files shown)


aalexand:
Please rename the PR to be a short summary of what exactly is being fixed and why.

internal/driver/webui_test.go
```diff
@@ -119,7 +119,10 @@ func TestWebInterface(t *testing.T) {
 		for count := 0; count < 2; count++ {
 			wg.Add(1)
 			go func() {
-				http.Get(path)
+				res, err := http.Get(path)
```
aalexand: Why not fail the test if the call fails? Do you expect it to ever fail?

internal/driver/webui_test.go
```diff
@@ -119,7 +119,10 @@ func TestWebInterface(t *testing.T) {
...
+				if err == nil {
+					ioutil.ReadAll(res.Body)
```
aalexand: Check the error returned?

internal/driver/webui_test.go
nolanmar511: Done.

internal/driver/webui_test.go
nolanmar511: Done.

internal/driver/webui_test.go
```diff
@@ -119,7 +119,15 @@ func TestWebInterface(t *testing.T) {
 		for count := 0; count < 2; count++ {
 			wg.Add(1)
 			go func() {
-				http.Get(path)
+				res, err := http.Get(path)
+				if err != nil {
+					t.Error("could not fetch", c.path, err)
+				} else {
+					_, err = ioutil.ReadAll(res.Body)
+					if err != nil {
```
aalexand: Merge with the assignment.

internal/driver/webui_test.go
```diff
@@ -119,7 +119,15 @@ func TestWebInterface(t *testing.T) {
 		for count := 0; count < 2; count++ {
 			wg.Add(1)
 			go func() {
-				http.Get(path)
+				res, err := http.Get(path)
+				if err != nil {
+					t.Error("could not fetch", c.path, err)
```
aalexand: Add continue, drop else. https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

nolanmar511: Added a returned and called wg.Done() before. 
Done.

internal/driver/webui_test.go
nolanmar511: Done.

nolanmar511:
PTAL

internal/driver/webui_test.go
```diff
@@ -119,7 +119,15 @@ func TestWebInterface(t *testing.T) {
...
+					wg.Done()
```
aalexand: Consider moving these wg.Done() calls to be a single "defer wg.Done()" at the very beginning of the function. That will make sure that it is always called, regardless of how the control flow changes in the future.

PR #316, opened 2018-02-07
Use random port when no port specified with -http

### LINKED ISSUE #307: Random port doesn't work
The readme says if you use `-http=":"` it will run on a random port, but actually you just get a parse error due to [this code](https://github.com/google/pprof/blob/80fc05d9036b19d9903eb19b0fb8aca62af01f8a/internal/driver/webui.go#L90).

    pprof: invalid port number: strconv.Atoi: parsing "": invalid syntax

nolanmar511:
This fixes #307.
When the port is not specified with the `-http` option, a random port is selected using `net.Listen()`.

### DIFF
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -142,7 +142,6 @@ func getHostAndPort(hostport string) (string, int, error) {
 		if err != nil {
 			return "", 0, fmt.Errorf("could not generate random port: %v", err)
 		}
-		fmt.Printf("\n%v\n", port)
 	} else {
 		port, err = strconv.Atoi(portStr)
 		if err != nil {
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -83,18 +83,10 @@ type webArgs struct {
 }
 
 func serveWebInterface(hostport string, p *profile.Profile, o *plugin.Options, wantBrowser bool) error {
-	host, portStr, err := net.SplitHostPort(hostport)
+	host, port, err := getHostAndPort(hostport)
 	if err != nil {
-		return fmt.Errorf("could not split http address: %v", err)
+		return err
 	}
-	port, err := strconv.Atoi(portStr)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/driver/webui.go
```diff
@@ -131,6 +123,34 @@ func serveWebInterface(hostport string, p *profile.Profile, o *plugin.Options, w
...
+		ln, err := net.Listen("tcp", net.JoinHostPort(host, "0"))
+		if err != nil {
+			return "", 0, fmt.Errorf("could not generate random port: %v", err)
+		}
+		port = ln.Addr().(*net.TCPAddr).Port
+		err = ln.Close()
+		if err != nil {
+			return "", 0, fmt.Errorf("could not generate random port: %v", err)
+		}
+		fmt.Printf("\n%v\n", port)
```
aalexand: Leftover debug output? (certainly doesn't look like meaningful user-visible output, so assuming this is from debug)

internal/driver/webui.go
```diff
@@ -131,6 +123,34 @@ func serveWebInterface(hostport string, p *profile.Profile, o *plugin.Options, w
...
+		host = "localhost"
+	}
+	var port int
+	if portStr == "" {
+		ln, err := net.Listen("tcp", net.JoinHostPort(host, "0"))
+		if err != nil {
+			return "", 0, fmt.Errorf("could not generate random port: %v", err)
+		}
+		port = ln.Addr().(*net.TCPAddr).Port
+		err = ln.Close()
```
aalexand: I think this is racy in the sense that another program may grab the port after the socket is closed here and opened again by the HTTP server. Did you consider this possibility? Is this too unlikely to worry about?

nolanmar511: It's definitely possible -- I believe it will occur rarely and I could not find another way to generate an unused port. 

To reduce this possibility, I could shift the functionality of getHostAndPort() into serveWebInterface(), and close the connection later. But that would make it harder to test random port selection and still not completely eliminate the race.

Issue #277, opened 2017-12-06
Cannot open perf.data generated by Linux perf (regression)


navytux:
Please answer these questions before submitting your issue. Thanks!

### What version of pprof are you using?

I'm on latest pprof from `github.com/google/pprof` (4b386c29ec) and on latest `github.com/google/perf_data_converter` (b86dd5276e)

### What operating system and processor architecture are you using?

`Linux deco 4.13.0-1-amd64 #1 SMP Debian 4.13.13-1 (2017-11-16) x86_64 GNU/Linux`

### What did you do?

I've tried to analyze `perf.data` generated by perf via pprof and it failed even though I have latest `perf_to_profile` on my path.

### What did you expect to see?

pprof loads profiling data from perf.data just ok, as readme says.

### What did you see instead?

```
$ perf record sleep 1
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.021 MB perf.data (7 samples) ]

$ pprof perf.data
Converting perf.data to a profile.proto... (May take a few minutes)
perf.data: failed to convert perf.data file. Try github.com/google/perf_data_converter: signal: aborted
pprof: failed to fetch any source profiles
```

--------

I've traced this down to the following:

- when seeing perf.data, pprof, before invoking perf_to_profile, creates temporary file for destination: https://github.com/google/pprof/blob/41ac4e1c/internal/driver/fetch.go#L532
- however perf_to_profile aborts when destination already exists: https://github.com/google/perf_data_converter/blob/b86dd5276e/perf_to_profile.cc#L35
https://github.com/google/perf_data_converter/blob/b86dd5276e/perf_to_profile.cc#L64
- it used not to be the case before
https://github.com/google/perf_data_converter/commit/48a540dd9376e37bb5edcc89ddd0715dcd867145#diff-183b816d357be307fcb59fc0a2fd2e03L71.

and indeed:

```
$ perf_to_profile perf.data out.pprof
[INFO:quipper/perf_reader.cc:847] Number of events stored: 185
[ERROR:quipper/perf_parser.cc:217] Unknown event type: 79
[INFO:perf_data_handler.cc:414] guessing main for pid: 13792 /bin/sleep
$ echo $?
0
$ perf_to_profile perf.data out.pprof
[INFO:quipper/perf_reader.cc:847] Number of events stored: 185
[ERROR:quipper/perf_parser.cc:217] Unknown event type: 79
[...]

PR #320, opened 2018-02-13
Skip creating temp file before calling perf_to_profile. (#277)

lannadorai:
Issue: #277
perf_to_profile aborts if the provided output profile *.pb.gz is already created so skip creating a temp profile file in pprof before calling perf_to_profile.


internal/driver/tempfile.go
```diff
@@ -21,6 +21,19 @@ import (
 	"sync"
 )
 
+// getNextTempFile returns the next available temp file in dir with the provided prefix and suffix.
+func getNextTempFile(dir, prefix, suffix string) (string, error) {
```
aalexand: There is quite a bit of duplicated code in here...  Can't you just use existing `newTempFile`, grab the name of the returned file using Name() and then Close() it before use?

lannadorai: `newTempFile` creates the temp file. But this change is to skip it. I will modify `newTempFile` to use `getNextTempFile`.

aalexand:
Did you mean to close the PR?

lannadorai:
I wanted to send out a new patch. It didn’t appear in the PR. So, I created
a new one.

PR #321, opened 2018-02-14
Skip creating temp file before calling perf_to_profile. (#277)

lannadorai:
Issue: #277
perf_to_profile aborts if the provided output profile *.pb.gz is already created so skip creating a temp profile file in pprof before calling perf_to_profile.

### DIFF
--- a/third_party/svgpan/svgpan.go
+++ b/third_party/svgpan/svgpan.go
@@ -12,7 +12,7 @@ const JSSource = `
  * ======================
  *
  * Given an unique existing element with id "viewport" (or when missing, the
- * first g-element), including the the library into any SVG adds the following
+ * first g-element), including the library into any SVG adds the following
  * capabilities:
  *
  *  - Mouse panning


internal/driver/fetch.go
```diff
@@ -529,16 +529,23 @@ func convertPerfData(perfPath string, ui plugin.UI) (*os.File, error) {
 	ui.Print(fmt.Sprintf(
 		"Converting %s to a profile.proto... (May take a few minutes)",
 		perfPath))
-	profile, err := newTempFile(os.TempDir(), "pprof_", ".pb.gz")
+	// Get a tempFile
```
aalexand: Remove this comment, it repeats the code.

internal/driver/fetch.go
```diff
@@ -529,16 +529,23 @@ func convertPerfData(perfPath string, ui plugin.UI) (*os.File, error) {
...
 	}
-	deferDeleteTempFile(profile.Name())
-	cmd := exec.Command("perf_to_profile", perfPath, profile.Name())
+	cmd := exec.Command("perf_to_profile", perfPath, tempFilename)
 	if err := cmd.Run(); err != nil {
-		profile.Close()
 		return nil, fmt.Errorf("failed to convert perf.data file. Try github.com/google/perf_data_converter: %v", err)
 	}
+
+	// Read the profile generated by perf_to_profile
```
aalexand: I'd remove this comment as well. And it's not really reading the profile here, it only opens the file.

internal/driver/fetch.go
```diff
@@ -529,16 +529,23 @@ func convertPerfData(perfPath string, ui plugin.UI) (*os.File, error) {
...
+	profile, err := os.Open(tempFilename)
+	if err != nil {
+		return nil, fmt.Errorf("open profile for reading: %v", err)
```
aalexand: "failed to open profile file %q: %v", tempFilename, err

internal/driver/tempfile.go
```diff
@@ -21,16 +21,28 @@ import (
 	"sync"
 )
 
-// newTempFile returns a new output file in dir with the provided prefix and suffix.
-func newTempFile(dir, prefix, suffix string) (*os.File, error) {
+// getNextTempFile returns the next available temp file in dir with the provided prefix and suffix.
```
aalexand: Nit: please fit the doc text into 80 cols.

aalexand: ... returns the path of the next available temp file ...

internal/driver/tempfile.go
```diff
@@ -21,16 +21,28 @@ import (
...
-	return nil, fmt.Errorf("could not create file of the form %s%03d%s", prefix, 1, suffix)
+	return "", fmt.Errorf("could not fetch next available file of the form %s%03d%s till %s%03d%s", prefix, 1, suffix, prefix, 9999, suffix)
+}
+
+// newTempFile returns a new output file in dir with the provided prefix and suffix.
+func newTempFile(dir, prefix, suffix string) (*os.File, error) {
+	var filePath string
+	var err error
+
+	if filePath, err = getNextTempFile(dir, prefix, suffix); err != nil {
```
aalexand: To write this a bit shorter, assign the variables:
```
func newTempFile(...) (...) {
  filePath, err := getNextTempFile(...)
  if err != nil {
    return nil, err
  }
  return os.Create(filePath)
}
```

internal/driver/tempfile.go
```diff
@@ -21,16 +21,28 @@ import (
 	"sync"
 )
 
-// newTempFile returns a new output file in dir with the provided prefix and suffix.
-func newTempFile(dir, prefix, suffix string) (*os.File, error) {
+// getNextTempFile returns the next available temp file in dir with the provided prefix and suffix.
+func getNextTempFile(dir, prefix, suffix string) (string, error) {
```
aalexand: Using `get` is not idiomatic in Go. Better: `nextTempPath`.

internal/driver/fetch.go
```diff
@@ -529,16 +529,23 @@ func convertPerfData(perfPath string, ui plugin.UI) (*os.File, error) {
...
 		perfPath))
-	profile, err := newTempFile(os.TempDir(), "pprof_", ".pb.gz")
+	// Get a tempFile
+	tempFilename, err := getNextTempFile(os.TempDir(), "pprof_", ".pb.gz")
 	if err != nil {
 		return nil, err
 	}
-	deferDeleteTempFile(profile.Name())
-	cmd := exec.Command("perf_to_profile", perfPath, profile.Name())
+	cmd := exec.Command("perf_to_profile", perfPath, tempFilename)
```
aalexand: By the way, I think the perf_data_converter change that broke its usage by pprof is this commit: https://github.com/google/perf_data_converter/pull/8/files#diff-183b816d357be307fcb59fc0a2fd2e03. Before that commit, perf_to_profile would happily overwrite the destination file. With that change, it started to refuse the conversion if the destination file exists. I know wonder what was the reason for that change. Without that change no pprof changes would be needed.

aalexand:
@lannadorai We discussed that instead of making this code change you'll add a flag to perf_to_profile which allows overwriting the output file. So I'm closing this pull request. Please send a pull request for the converter instead. (Let's do this soon please, this fix has been taking longer than it should.)

PR #325, opened 2018-02-19
Add a separate function to binutils to check if a file is an executable

rauls5382:
Currently bu.Open() is used to detect if a file is an executable,
but it depends on the mapping, which is not available before opening
the input profiles.

Create a separate ExecutableBuildID function on the ObjTool interface that only
does a basic check to see if a binary is an executable, and extracts the build
id if available.

This produces better diagnostics if there are issues with the mappings, as we
will still recognize the file as a binary.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -123,6 +123,12 @@ type MappingSources map[string][]struct {
 
 // An ObjTool inspects shared libraries and executable files.
 type ObjTool interface {
+	// ExecutableBuildID returns the build id associated to an executable.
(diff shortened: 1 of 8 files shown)


rauls5382:
Actually I'm going to withdraw this. Probably not worth changing the plugin interface, which could break users.

aalexand:
To improve the diagnostics I was thinking of something simple like return the last error between attempting to open a mach file and an ELF file and do the mach open on Mac last but on Linux do the ELF open last. The code would be a bit gross but the interface doesn't have to be changed.

Issue #329, opened 2018-02-23
The function internal/driver.generateReport does not use o.Writer to write output


sudachen:
Pprof always creates output file and does not use o.Writer field even it's defined. 
https://github.com/google/pprof/blob/master/internal/driver/driver.go#L139
```go
    o.UI.PrintErr("Generating report in ", output)
    out, err := os.Create(output)
    if err != nil {
    	return err
    }
```

PR #330, opened 2018-02-23
Use o.Writer if not nil to write output

sudachen:
#329 (The function internal/driver.generateReport does not use o.Writer to write output)

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -20,6 +20,7 @@ package driver
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


internal/driver/driver.go
```diff
@@ -138,7 +139,15 @@ func generateReport(p *profile.Profile, cmd []string, vars variables, o *plugin.
 
 	// Output to specified file.
 	o.UI.PrintErr("Generating report in ", output)
-	out, err := os.Create(output)
+
+	openOutput := func(output string) (io.WriteCloser, error) {
+		if o.Writer != nil {
```
aalexand: Rather than have a condition here, should the code always use the writer since it's initialized to oswriter{} in https://github.com/google/pprof/blob/db0be723d40dfbb90c702d493a71d398173358e7/internal/driver/options.go#L38?

aalexand:
You ned to sign the CLA before you can contribute.

internal/driver/driver.go
sudachen: Yep, sure! I missed It.

PR #331, opened 2018-02-26
Update perf_to_profile cli arguments to use flags. (#277)

lannadorai:
Issue: #277
As perf_to_profile uses flags to parse its arguments, modify the call to perf_to_profile to use flags. Also pass the '-f' flag to overwrite the existing output profile file.
Redirect perf_to_profile's stdout/stderr to pprof's stdout/stderr.

Tested by executing pprof with a perf.data:
Eg: 
perf record -e cycles -o perf.data -- sleep 1
./pprof perf.data


aalexand:
@lannadorai Could you describe the testing you did in the PR description?

PR #341, opened 2018-03-23
Allow passing in a custom HTTPServer

tschottdorf:
This allows embedding the pprof ui into an existing application. See
https://github.com/cockroachdb/cockroach/pull/24145 for an example.

The second change replaces e82ee9a. It may look cosmetic, but note that
`plugin.UI` can be passed in from the outside. This is important for embedding
pprof into an application for easy access to the web ui: with this change, such
an application can pass a non-Terminal `profile.UI` and avoids the browser
code which doesn't make sense in that context.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -41,24 +41,36 @@ func (o *Options) internalOptions() *plugin.Options {
 	if o.Sym != nil {
 		sym = &internalSymbolizer{o.Sym}
 	}
+	var httpServer func(args *plugin.HTTPServerArgs) error
+	if o.HTTPServer != nil {
+		httpServer = func(args *plugin.HTTPServerArgs) error {
+			return o.HTTPServer(((*HTTPServerArgs)(args)))
+		}
+	}
 	return &plugin.Options{
-		Writer:  o.Writer,
-		Flagset: o.Flagset,
-		Fetch:   o.Fetch,
-		Sym:     sym,
-		Obj:     obj,
-		UI:      o.UI,
+		Writer:     o.Writer,
+		Flagset:    o.Flagset,
+		Fetch:      o.Fetch,
+		Sym:        sym,
+		Obj:        obj,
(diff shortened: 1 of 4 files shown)


driver/driver.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/google/pprof/profile"
 )
 
+type HTTPServerArgs plugin.HTTPServerArgs
```
nolanmar511: The golint check in out test suite isn't passing right now because this exported type doesn't have a comment. 

Should `plugin.HTTPServerArgs` be defined as a separate type? I ask because I see both `plugin.HTTPServerArgs` and `HTTPServerArgs` used below.

internal/driver/webui.go
```diff
@@ -117,7 +117,7 @@ func serveWebInterface(hostport string, p *profile.Profile, o *plugin.Options, w
 		},
 	}
 
-	if wantBrowser {
+	if o.UI.IsTerminal() {
```
aalexand: We may need to extend the UI interface with a separate function like CanOpenBrowser() or even with OpenBrowser(url) - I am thinking about environments like a terminal with remote SSH session where it is terminal, but opening a browser GUI against a port of a process user spawns in that terminal is not usually feasible.

aalexand: This doesn't need to be in this PR, I'm just raising a discussion point.

driver/driver.go
tschottdorf: Thanks, added the comment.
Even if I define this as a separate type, I'll still have to convert to the `internal` variant below, so it wouldn't change anything there. I might be missing your point.

internal/driver/webui.go
tschottdorf: Ack, I'll hold off on that for now.

aalexand:
@tschottdorf Great, I'll merge once Travis CI is completed and green.

Issue #208, opened 2017-08-24
Automated continuous testing of the web UI


aalexand:
#207 discusses future enhancements to the web UI. It would be good to improve the automated testing story for the feature so that contributors feel safe adding new functionality.

Using https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-the-Chrome-addon-in-the-headless-mode is one option to consider.

sourabh1031:
Hello I am new here . I am beginner in golang but want to learn further by solving this issue. Can you please guide me the necessary prerequisites that i must have to solve the issue

kalyanac:
Thank you for your interest and offering to help. 

This particular issue is specifically about adding a full end-to-end test for the web UI to run as a continuous Travis test. As such, it will focus more on using a Chrome remoting interface, and less on Go. 

If this interests you, please see below:
Chrome DevTools Protocol: https://github.com/chromedp/chromedp
https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-the-Chrome-addon-in-the-headless-mode

PR #348, opened 2018-03-29
Allow serving web ui at arbitrary root

tschottdorf:
Previously, the web UI was hard-coded to be served at `/` but
applications which are *embedding* the ui are unlikely to host it there.

This change makes the links between the various ui pages relative,
addressing this problem.

To make sure this doesn't rot, the internal http server now serves the
ui at /ui/ (redirecting there from the root).

I think we can all agree that the ui code and JavaScript is kind of a
haphazard affair. I performed a great deal of clicking around to make
sure the semantics didn't change and interestingly spent some time
trying to fix semantics I later realized weren't actually there: I
thought that when navigating to another page from a focused search (say
`flamegraph?f=x`) the focus would be carried over. This doesn't seem to
be the case, however.

The point being, there's no test coverage and setting it up is way
beyond what I can do here. I'm happy to manually verify whatever
sequence of clicks is suggested, though.

Touches https://github.com/cockroachdb/cockroach/pull/24145.

### DIFF
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -69,7 +69,6 @@ func (ec *errorCatcher) PrintErr(args ...interface{}) {
 
 // webArgs contains arguments passed to templates in webhtml.go.
 type webArgs struct {
-	BaseURL    string
 	Title      string
 	Errors     []string
 	Total      int64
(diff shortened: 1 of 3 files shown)


internal/driver/webui.go
```diff
@@ -172,7 +171,11 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
 		}
 		h.ServeHTTP(w, req)
 	})
-	s := &http.Server{Handler: handler}
+
+	mux := http.NewServeMux()
+	mux.Handle("/ui/", http.StripPrefix("/ui", handler))
+	mux.Handle("/", http.RedirectHandler("/ui/", http.StatusTemporaryRedirect))
```
aalexand: I'd add a comment on why the redirect is done.

aalexand:
@tschottdorf Added a small comment. I agree #208 is overdue.

internal/driver/webui.go
tschottdorf: Done.

tschottdorf:
Thanks for the review, added the comment.

PR #354, opened 2018-04-05
Improve profile filter unit tests, fix show filtering of mappings

lvng:
The first step to adding new filters is to improve the existing unit tests. It appears that the existing unit tests for filtering only check the return values, not the filtered profile, so i started working on a new test case that covers that more thoroughly. Eventually I'll add the new filters, and add test cases for that there.

The test cases I'm adding are based on the tests for continuous pprof frontend, which will help us ensure compatibility between the two filter implementations.

This is still a work in progress but I would appreciate some early feedback to make sure I'm on the right track.


profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
```
aalexand: Fix the year.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+		if gh != tc.hm {
+			t.Errorf("Filter #%d, got hm=%v, want %v", tx, gh, tc.hm)
+		}
+		if gnh != tc.hnm {
+			t.Errorf("Filter #%d, got hnm=%v, want %v", tx, gnh, tc.hnm)
+		}
+	}
+}
+
+func TestFilter2(t *testing.T) {
```
aalexand: Try to come up with a better name for the test based on what it's doing and how it's different from the existing test. Ideally just merge these tests into one - it's not clear why both need to exist, they are rather similar.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+			{Value: []int64{3}, Location: []*Location{locations[7], locations[8]}},
+			{Value: []int64{4}, Location: []*Location{locations[9], locations[4], locations[10], locations[7]}},
+		},
+	}
+
+	for _, testCase := range []struct{
+		name string
+		focus, ignore, hide, show *regexp.Regexp
+		// todo: fm, im, hm, hnm           bool
+		wantPaths [][]int // expected paths. wantPaths[sampleIndex][frameIndex] = locationIndex
```
aalexand: Nit: proper sentences in comments please.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+		wantValues []int64 // expected values. wantValues[sampleIndex] = sampleValue
+	}{
+		{
+			name: "empty filters keep all frames",
+			// todo: fm: true, // nil focus matches every sample
```
aalexand: Commented code should be removed.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+			newLocation := *locations[i]
+			newLocation.Line = nil
+			filteredLocations = append(filteredLocations, &newLocation)
+		}
+	}
+
+	return samples, filteredLocations
+}
+
+func TestTagFilter(t *testing.T) {
```
aalexand: It would be better to move things around in a separate PR so that it's clear whether anything was changed when things were moved.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+					t.Fatalf("failed to get diff: %v", err)
+				}
+				t.Errorf("filtered profile got diff(want->got): \n%s", diff)
+			}
+		})
+	}
+}
+
+// makeSamples constructs profile samples from basic inputs.
+// paths: location indices, indexed by sample and frame.
```
aalexand: Nit: I don't remember seeing such per-argument format of documentation in Go code or pprof, could you look at existing documentation for functions in pprof and document similarly?

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+		{ID: 4, Mapping: mappings[0], Address: 0x4000, Line: []Line{{Function: functions[3], Line: 1}}},
+		{ID: 5, Mapping: mappings[0], Address: 0x5000, Line: []Line{{Function: functions[4], Line: 1}}},
+		{ID: 6, Mapping: mappings[0], Address: 0x6000, Line: []Line{{Function: functions[5], Line: 1}}},
+		{ID: 7, Mapping: mappings[0], Address: 0x7000, Line: []Line{{Function: functions[6], Line: 1}}},
+		{ID: 8, Mapping: mappings[0], Address: 0x8000, Line: []Line{{Function: functions[7], Line: 1}}},
+		{ID: 9, Mapping: mappings[0], Address: 0x9000, Line: []Line{{Function: functions[8], Line: 1}}},
+		{ID: 10, Mapping: mappings[0], Address: 0x10000, Line: []Line{{Function: functions[9], Line: 1}}},
+		{ID: 11, Mapping: mappings[1], Address: 0x11000, Line: []Line{{Function: functions[10], Line: 1}}},
+	}
+	profile := &Profile{
```
nolanmar511: In test_profile.go, we declare profiles (and functions/locations, etc) outside of the test cases so they can be copied and used in multiple tests. This may be something you want to consider doing.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+package profile
+
+import (
+	"regexp"
+	"testing"
+
+	"github.com/google/pprof/internal/proftest"
+)
+
+func TestFilter(t *testing.T) {
```
nolanmar511: I think you could (and should for clarity) combine TestFilter and TestFilter2 by checking fm, im, hm, and hmn in TestFilter2.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+			hide: regexp.MustCompile("fun10"),
+			wantPaths: [][]int{
+				{1},
+				{1},
+			},
+			wantValues: []int64{1, 2},
+		},
+	} {
+		t.Run(testCase.name, func(t *testing.T) {
+			gotProfile := profile.Copy()
```
nolanmar511: It may make sense to have the profile be a field in the test structure.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+}
+
+func TestFilter2(t *testing.T) {
+	mappings := []*Mapping{
+		{ID: 1, Start: 0x10000, Limit: 0x40000, File: "map0", HasFunctions: true, HasFilenames: true, HasLineNumbers: true, HasInlineFrames: true},
+		{ID: 2, Start: 0x50000, Limit: 0x70000, File: "map1", HasFunctions: true, HasFilenames: true, HasLineNumbers: true, HasInlineFrames: true},
+	}
+	functions := []*Function{
+		{ID: 1, Name: "fun0", SystemName: "fun0", Filename: "file0"},
```
aalexand: Up to you, but I feel you can achieve equivalent code coverage with less functions in the test input.

profile/filter_test.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+}
+
+// makeSamples constructs profile samples from basic inputs.
+// paths: location indices, indexed by sample and frame.
+// 	i.e. paths[sampleIndex][frameIndex] = locationIndex
+// values: values for each sample. i.e. values[sampleIndex] = value
+// locations: locations to copy for new samples. unused locations are not
+// 	removed but their lines are set to nil.
+func makeFilteredSamples(paths [][]int, values []int64, locations []*Location) ([]*Sample, []*Location) {
```
aalexand: This helper feels over-complex, I am still reading through its code. I would consider a clearer (even if a bit less accurate) validation of the expected result. For example, walk over profile and compute a simple flat report (funcName, flat, cum) similar to how printTopProto in internal/report/report.go does. And express the expected result as a list of rows of that type, like [("fun1", 200, 500), "fun2, 300, 600)] or like that. With that and a bit smaller number of functions I think the test would end up being more readable and easier extensible over time, if needed.

profile/filter_test.go
lvng: I've added checks for the return value, and removed the original test function since its now redundant.

profile/filter_test.go
lvng: I put it in the function to avoid polluting the global namespace while it's only used by one function. we can always pull it out if we decide to reuse it.

profile/filter_test.go
lvng: we can add that if we ever need to test against more than one profile. for now, i think its simpler to just have one profile which can serve all the test cases.

profile/filter_test.go
lvng: done

profile/filter_test.go
lvng: added return value checks to the new function, making the original obsolete so it is removed.

profile/filter_test.go
lvng: done

profile/filter_test.go
lvng: done

profile/filter_test.go
lvng: i looked and couldnt find anything, but i think this is right thing to do here.

profile/filter_test.go
lvng: true, though i used this profile because it's identical to the one used in the frontend tests, which makes it easier to compare test cases.

profile/filter_test.go
lvng: That doesn't seem worth it, imo. It sounds like its just as much work to create and maintain a less accurate test.

profile/filter_test.go
aalexand: Try harder :)

aalexand: Go doc style is that the text should be readable English, not some kind of machine-friendly special formatting. One example is https://github.com/golang/go/blob/e13a213c7f9d78cfb79f91bcf9b75e3d22362a92/src/runtime/map.go#L691, there are more around.

aalexand:
@lvng Travis CI fails FYI, please fix that before PTAL.

profile/filter_test.go
lvng: done: https://github.com/google/pprof/pull/356

profile/filter_test.go
aalexand: We could simplify the test here and take the changes back. It would also be good to have in the test data locations that have multiple lines to simulate the inline data and verify its filtering.

profile/filter_test.go
```diff
@@ -0,0 +1,404 @@
...
+
+	allPaths := [][]int{
+		{0, 1, 2, 3},
+		{4, 5, 1, 6},
+		{7, 8},
+		{9, 4, 10, 7},
+	}
+	allValues := []int64{1, 2, 3, 4}
+
+	for _, testCase := range []struct {
```
aalexand: testCase -> tc, like in TestTagFilter. It's an idiomatic abbreviation.

profile/filter_test.go
```diff
@@ -0,0 +1,404 @@
...
+					t.Fatalf("failed to get diff: %v", err)
+				} else {
+					t.Errorf("filtered profile got diff(want->got):\n%s", diff)
+				}
+			}
+		})
+	}
+}
+
+// makeFilteredSamples constructs  filtered profile samples and locations from
```
aalexand: Double space.

profile/filter_test.go
```diff
@@ -0,0 +1,404 @@
...
+// basic inputs to simplify the specification of expected values in filter test
+// cases. paths contains the location indices, indexed by sample and frame, of
```
aalexand: The text is still mostly list of arguments. Could you make it more descriptive in terms of what the function is doing rather than in terms of what arguments are?

profile/filter_test.go
```diff
@@ -0,0 +1,404 @@
...
+// the filtered samples. values contains the values, indexed by sample, of the
+// filtered samples. filterLocations is whether to remove function info from
+// unused locations. locations is the slice of locations from the original,
+// unfiltered profile, used to derive the new filtered locations.
+func makeFilteredSamples(paths [][]int, values []int64, filterLocations bool, locations []*Location) ([]*Sample, []*Location) {
+	samples := make([]*Sample, len(paths))
+	usedLocationIndices := map[int]bool{}
```
aalexand: Local variable names should be shorter to reduce the visual noise. "usedLocationIndices" -> "usedLocs" etc.

profile/filter.go
```diff
@@ -41,8 +41,8 @@ func (p *Profile) FilterSamplesByName(focus, ignore, hide, show *regexp.Regexp)
 			}
 		}
 		if show != nil {
-			l.Line = l.matchedLines(show)
-			if len(l.Line) == 0 {
+			if !l.matchesName(show) {
```
aalexand: It feels the new code won't handle correctly show filter when it matches some lines within the location. l.matchesName() will return true in that case which means line filtering won't be done.

profile/filter_test.go
aalexand: After looking at the fix in filter.go and leaving a comment there, I do think we need to add test cases that simulate inlines.

profile/filter_test.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+// makeFilteredSamples constructs  filtered profile samples and locations from
+// basic inputs to simplify the specification of expected values in filter test
+// cases. paths contains the location indices, indexed by sample and frame, of
+// the filtered samples. values contains the values, indexed by sample, of the
+// filtered samples. filterLocations is whether to remove function info from
+// unused locations. locations is the slice of locations from the original,
+// unfiltered profile, used to derive the new filtered locations.
+func makeFilteredSamples(paths [][]int, values []int64, filterLocations bool, locations []*Location) ([]*Sample, []*Location) {
```
aalexand: I still think there is a possibility to get by with a simpler helper and a more readable way to express the expected result for the tests. Here is one approach: https://github.com/google/pprof/compare/master...aalexand:filter-helper

profile/filter_test.go
lvng: added test profile and test cases for inline functions

profile/filter_test.go
lvng: done

profile/filter.go
lvng: good catch. i fixed the issue and added test cases to cover it.

profile/filter_test.go
lvng: obsolete.

profile/filter_test.go
lvng: obsolete

profile/filter_test.go
lvng: obsolete

profile/filter_test.go
lvng: i merged your branch into mine, and removed the old helper function.

profile/filter_test.go
lvng: now that ive added test cases for inline function, there are two profiles defined outside the test case.

lvng:
PTAL

profile/filter_test.go
```diff
@@ -15,60 +15,398 @@
...
 	"regexp"
+	"strings"
 	"testing"
+
+	"github.com/google/pprof/internal/proftest"
 )
 
-func TestFilter(t *testing.T) {
-	// Perform several forms of filtering on the test profile.
+var simpleProfileMappings = []*Mapping{
```
aalexand: Nit: you don't have other mappings var, so simply naming it `mappings` would be sufficient.

profile/filter_test.go
```diff
@@ -15,60 +15,398 @@
...
+	{ID: 1, Start: 0x10000, Limit: 0x40000, File: "map0", HasFunctions: true, HasFilenames: true, HasLineNumbers: true, HasInlineFrames: true},
+	{ID: 2, Start: 0x50000, Limit: 0x70000, File: "map1", HasFunctions: true, HasFilenames: true, HasLineNumbers: true, HasInlineFrames: true},
+}
 
-	type filterTestcase struct {
-		focus, ignore, hide, show *regexp.Regexp
-		fm, im, hm, hnm           bool
-	}
+var simpleProfilefunctions = []*Function{
```
aalexand: Similarly, just `functions`.

profile/filter_test.go
```diff
@@ -15,60 +15,398 @@
...
+	{ID: 6, Name: "fun5", SystemName: "fun5", Filename: "file5"},
+	{ID: 7, Name: "fun6", SystemName: "fun6", Filename: "file6"},
+	{ID: 8, Name: "fun7", SystemName: "fun7", Filename: "file7"},
+	{ID: 9, Name: "fun8", SystemName: "fun8", Filename: "file8"},
+	{ID: 10, Name: "fun9", SystemName: "fun9", Filename: "file9"},
+	{ID: 11, Name: "fun10", SystemName: "fun10", Filename: "file10"},
+}
 
-	for tx, tc := range []filterTestcase{
+var simpleProfilelocations = []*Location{
```
aalexand: "Simple" is subjective. How about expressing the exact purpose in the name: `noInlinesLocs`?

profile/filter_test.go
```diff
@@ -15,60 +15,398 @@
...
+	{ID: 5, Mapping: simpleProfileMappings[0], Address: 0x5000, Line: []Line{{Function: simpleProfilefunctions[4], Line: 1}}},
+	{ID: 6, Mapping: simpleProfileMappings[0], Address: 0x6000, Line: []Line{{Function: simpleProfilefunctions[5], Line: 1}}},
+	{ID: 7, Mapping: simpleProfileMappings[0], Address: 0x7000, Line: []Line{{Function: simpleProfilefunctions[6], Line: 1}}},
+	{ID: 8, Mapping: simpleProfileMappings[0], Address: 0x8000, Line: []Line{{Function: simpleProfilefunctions[7], Line: 1}}},
+	{ID: 9, Mapping: simpleProfileMappings[0], Address: 0x9000, Line: []Line{{Function: simpleProfilefunctions[8], Line: 1}}},
+	{ID: 10, Mapping: simpleProfileMappings[0], Address: 0x10000, Line: []Line{{Function: simpleProfilefunctions[9], Line: 1}}},
+	{ID: 11, Mapping: simpleProfileMappings[1], Address: 0x11000, Line: []Line{{Function: simpleProfilefunctions[10], Line: 1}}},
+}
+
+var simpleProfile = &Profile{
```
aalexand: `noInlinesProfile`

profile/filter_test.go
```diff
@@ -15,60 +15,398 @@
...
+	Location:      simpleProfilelocations,
+	Sample: []*Sample{
+		{Value: []int64{1}, Location: []*Location{simpleProfilelocations[0], simpleProfilelocations[1], simpleProfilelocations[2], simpleProfilelocations[3]}},
+		{Value: []int64{2}, Location: []*Location{simpleProfilelocations[4], simpleProfilelocations[5], simpleProfilelocations[1], simpleProfilelocations[6]}},
+		{Value: []int64{3}, Location: []*Location{simpleProfilelocations[7], simpleProfilelocations[8]}},
+		{Value: []int64{4}, Location: []*Location{simpleProfilelocations[9], simpleProfilelocations[4], simpleProfilelocations[10], simpleProfilelocations[7]}},
+	},
+}
+
+var allSimpleSampleFuncs = []string{
```
aalexand: `allNoInlinesSampleFuncs`

profile/filter_test.go
```diff
@@ -15,60 +15,398 @@
...
+	"fun0 fun1 fun2 fun3: 1",
+	"fun4 fun5 fun1 fun6: 2",
+	"fun7 fun8: 3",
+	"fun9 fun4 fun10 fun7: 4",
+}
+
+var inlineProfilelocations = []*Location{
```
aalexand: `inlinesLocs`

profile/filter_test.go
```diff
@@ -15,60 +15,398 @@
...
+	{ID: 1, Mapping: simpleProfileMappings[0], Address: 0x1000, Line: []Line{{Function: simpleProfilefunctions[0], Line: 1}, {Function: simpleProfilefunctions[1], Line: 1}}},
+	{ID: 2, Mapping: simpleProfileMappings[0], Address: 0x2000, Line: []Line{{Function: simpleProfilefunctions[2], Line: 1}, {Function: simpleProfilefunctions[3], Line: 1}}},
+	{ID: 3, Mapping: simpleProfileMappings[0], Address: 0x3000, Line: []Line{{Function: simpleProfilefunctions[4], Line: 1}, {Function: simpleProfilefunctions[5], Line: 1}, {Function: simpleProfilefunctions[6], Line: 1}}},
+}
+
+var inlineProfile = &Profile{
```
aalexand: `inlinesProfile`

profile/filter_test.go
aalexand: Also, curious - why separate profiles with inlines and no inlines? These could be just separate samples in the same profile - some that have multi-line locations and some that don't?

lvng: done

profile/filter_test.go
lvng: done.

profile/filter_test.go
lvng: done.

profile/filter_test.go
lvng: done.

profile/filter_test.go
lvng: done.

profile/filter_test.go
lvng: done.

profile/filter_test.go
lvng: done.

profile/filter_test.go
lvng: couple reasons:
- these test cases are largely duplicated from the cloud profiler test cases, so i didnt have to design them from scratch, and i dont have to worry missing an edge case that would lead to incompatibility.
- i think smaller test input are easier to understand and reason about.
- i think smaller test inputs are also easier to maintain, because the test cases dont all depend on the same profile.  if we used one profile for everything, then any modification to it would require changing all test cases. with multiple small profiles its easier to create new tests without breaking old ones.

PR #360, opened 2018-04-16
Add ShowFrom and HideFrom filters to profile filter lib.

lvng:
- Added HideFrom and ShowFrom. 
- They are not yet exposed as command line flags. This change is pretty big so I thought it would be helpful to keep it minimal but I could add that here if desired.
- Added lots of test cases for interactions between filters, as well as handling of locations with empty lines.


aalexand:
Could you squash your initial commit history and re-push? It's just noise.

lvng:
To squash the history I copied changes into a separate PR: https://github.com/google/pprof/pull/361

PR #361, opened 2018-04-16
Add ShowFrom and HideFrom filters to profile filter lib

lvng:
- Added HideFrom and ShowFrom filters.
- They are not yet exposed as command line flags. This change is pretty big so I thought it would be helpful to keep it minimal but I could add that here if desired. I plan to add that in another change.
- Added lots of test cases for interactions between filters, as well as handling of locations with empty lines.

### DIFF
--- a/profile/filter.go
+++ b/profile/filter.go
@@ -21,9 +21,19 @@ import "regexp"
 // FilterSamplesByName filters the samples in a profile and only keeps
 // samples where at least one frame matches focus but none match ignore.
 // Returns true is the corresponding regexp matched at least one sample.
-func (p *Profile) FilterSamplesByName(focus, ignore, hide, show *regexp.Regexp) (fm, im, hm, hnm bool) {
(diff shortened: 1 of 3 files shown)


profile/filter.go
```diff
@@ -21,9 +21,19 @@ import "regexp"
 // FilterSamplesByName filters the samples in a profile and only keeps
 // samples where at least one frame matches focus but none match ignore.
 // Returns true is the corresponding regexp matched at least one sample.
-func (p *Profile) FilterSamplesByName(focus, ignore, hide, show *regexp.Regexp) (fm, im, hm, hnm bool) {
+func (p *Profile) FilterSamplesByName(focus, ignore, showFrom, hideFrom, show, hide *regexp.Regexp) (fm, im, sfm, hfm, sm, hm bool) {
```
aalexand: This function has become too complicated now. I would make it a separate function, something like ShowHideFrom and put it to prune.go, next to PruneFrom, since the functionality is rather related to pruning. Having the function separate would make the code and the tests more focused and easier to read.

aalexand:
This PR is superseded by https://github.com/google/pprof/pull/372, closing.

PR #362, opened 2018-04-17
Add command-line editing support for interactive pprof

### LINKED ISSUE #57: Feature Request: Call back previous commands in interactive mode
I believe that this repo is the code that powers `go tool pprof`. If I'm wrong on that count, disregard everything else I say.

I'm using `pprof` on macOS 10.12.1 with go 1.7.3. I'm constantly using it in interactive mode, and it would be really nice to be able to call back previous commands with the "up" arrow key. Instead, of course, it prints `^[[A`, which I assume means that this isn't supported?

I completely understand if this isn't worth the development team's time, but I thought I would suggest it.

rauls5382:
Implement basic command-line editing, using github.com/chzyer/readline. This will not affect the golang
distribution as it uses its own pprof driver, with its own driver.

Fixes #57.
Only tested on Linux.
TODO: Implement auto-completion and persistent history.

### DIFF
--- a/pprof.go
+++ b/pprof.go
@@ -35,8 +35,7 @@ func main() {
 // readlineUI implements the driver.UI interface using the
 // github.com/chzyer/readline library
 type readlineUI struct {
-	rl        *readline.Instance
-	completer func(string) string
+	rl *readline.Instance
 }
 
 func newUI() driver.UI {
--- a/pprof.go
+++ b/pprof.go
@@ -19,13 +19,86 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
+	"github.com/chzyer/readline"
 	"github.com/google/pprof/driver"
 )
 
 func main() {
-	if err := driver.PProf(&driver.Options{}); err != nil {
+	if err := driver.PProf(&driver.Options{UI: newUI()}); err != nil {
 		fmt.Fprintf(os.Stderr, "pprof: %v\n", err)
 		os.Exit(2)
 	}
+}
+
+// readlineUI implements the driver.UI interface using the
+// github.com/chzyer/readline library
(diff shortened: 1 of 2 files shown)


rauls5382:
I did not consider using other terminal implementations. I'm flexible if there's consensus. 
chzyer/readline seems to be popular, though, and not sure if adding dependences into something in crypto/.*/ssh would have other downsides.

In any case, terminal support has to be explicitly ported/reimplemented into golang/go's custom pprof.go: we're technically not constrained to using the same terminal implementation.

The golang distribution does not exclude pprof.go from google/pprof, but I believe it is not compiled when building the distro. They could exclude this file if they really wanted to scrub the dependence.

Adding @hyangah for their opinion.

hyangah:
Please go ahead. 

As @rauls5382 said, the pprof cmd included in go distribution is built from cmd/pprof/pprof.go and we need to create the similar change there. The vendored main package is unused. I just briefly tested and the additional dependency in the vendored pprof seems ok.

pprof.go
```diff
@@ -19,13 +19,85 @@ package main
...
 func main() {
-	if err := driver.PProf(&driver.Options{}); err != nil {
+	if err := driver.PProf(&driver.Options{UI: newUI()}); err != nil {
 		fmt.Fprintf(os.Stderr, "pprof: %v\n", err)
 		os.Exit(2)
 	}
 }
+
+// readlineUI implements the driver.UI interface using the
+// github.com/chzyer/readline library
```
aalexand: I'd add a sentence to summarize what we discussed which is "It is contained in pprof.go to avoid the readline dependency in the vendored copy of pprof in Go distribution (which does not use this file)".

PR #367, opened 2018-04-27
Render icicle graph in "Flame Graph" view

### LINKED ISSUE #359: Icicle graph: inverted flame graph
In the comments for a recent Netflix's article about their FlameScope tool @paulirish has [noticed an interesting point][1]: in many cases, Flame Graphs are much higher than our display's height, so user has to start her's observation by scrolling to the bottom to look at the graph's root. Icicle graphs (inverted flame graph) fixes this behaviour.

To my knowledge, Chrome's performance profiler is the most widely used example of such view, while other tools / services also support it. For example [go-tourch](https://github.com/uber/go-torch) has `--inverted` flag.

In my opinion, it would be useful if pprof used inverted Flame Graph, or provided an option to transform one to another.

I use current pprof's master.

### What did you do?

```
% pprof -http :10080 cpu.prof
```

Open View → Flame Graph

[1]: https://medium.com/@paul_irish/great-stuff-bbc02caa144f

narqo:
This PR changes Flame Graph to be rendered as "icicle graph" (inverted flame graph).

The `inverted` option was introduced in d3_flame_graph.js 2.x, so the component was updated to current latest 2.0.0-alpha4.

In addition to that, some visual changes were done:

- Graph details were moved above the graph itself.
- Nodes are labelled with package name instead of full package path, that is `<package>.<func>`. Graph details bar still shows full package path `<github.com/owner/package>.<func>`.
- d3-tooltip has been disabled since the data it showed is always visible above the graph.

screen shot 2018-04-28 at 01 53 11

Fixes  #359


internal/driver/webhtml.go
```diff
@@ -1031,49 +1031,51 @@ function viewer(baseUrl, nodes) {
...
+    <div id="flamegraphdetails" class="flamegraph-details"></div>
     <div class="flamegraph-content">
       <div id="chart"></div>
     </div>
-    <div id="flamegraphdetails" class="flamegraph-details"></div>
   </div>
   {{template "script" .}}
   <script>viewer(new URL(window.location.href), {{.Nodes}});</script>
   <script>{{template "d3script" .}}</script>
-  <script>{{template "d3tipscript" .}}</script>
```
aalexand: Since this isn't used now, remove d3tip from third_party?

third_party_update.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
aalexand: This seems to only update the flamegraph directory, so maybe name the file as flamegraph_update.sh or something like that?

aalexand:
@narqo Left a couple comments, other than that looks good - thanks a lot for doing this!

narqo:
Addressed the comments and updated the PR. Please take another look.

Issue #323, opened 2018-02-15
Fix percentage calculation in diff mode (-base) and remove -positive_percentages 


nolanmar511:
Right now, the denominator used when computing percentages when there is a base profile is the sum of the absolute value for all samples in the merged profile (which is source profile merged with negative base profile), before samples are aggregated by function, line number, file, etc. Usually, this results in the denominator used when computing percentages being the sum of the total of samples in the source profile(s) and sum of the total of samples in the base profile, because the addresses for samples are generally different. However, this is not always the case; for example, when the source and base profile are identical, the denominator used would be zero.

This is useful when looking at cumulative profiles, but not useful when using the -base flag to compare profiles. So, we intend to add a new option `-diff_base`, which would be used similarly to `-base`, but compute the denominator for percentages in a way which is more useful when comparing profiles.

We want to make the denominator used when computing percentages when there is a diff base profile equal to the sum of all samples in the diff base profile. There are several advantages to this over the current method. Most notably, the denominator will be consistent, regardless of the address space of the profiles being compared. Also, percentages when comparing two different source profiles to the same base profile will be comparable, since the same denominator was used to compute both.

To do this, prior to merging the source and base profiles, a tag `pprof::base` with a value "true" will be added to each sample in the diff base profile. Then, when computing the denominator, we will return the absolute value of the sum of all profiles with the `pprof::base` tag if this tag is present in the profile; otherwise we will return the sum of the absolute value of all samples (which is currently done for all profiles). There are several advantages to this approach for computing the denominator. First, it will be a relatively small code change compared to other methods considered (for example, we considered maintaining a separate copy of the base profile through most steps which pprof does). Next, it provides some consistency when using a merged profile with a diff base -- if this profile is written out, the samples from the diff base profile will still have the `pprof::diff` tag, and so if pprof is used to inspect that profile, it will still compute the denominator as the sum of samples in the base profile.

Since this change means that the denominator used when there is a base profile will no longer generally be the sum of positive sample from the source profile and the absolute value of negative samples from the base profile, we are also removing the `-positive_percentages` flag, which allowed users to choose to only include positive samples when computing the denominator. This change makes this flag somewhat unnecessary -- since `-positive_percentages` was generally used to make the denominator equal to the sum of samples in one profile, which is very similar to what `-diff_base` does. This flag would also be problematic if used with the method for computing the denominator outline above, since, if it were used when computing the denominator for a merged profile as described above, the denominator would 0.

PR #369, opened 2018-05-01
add -diff flag for better profile comparision

nolanmar511:
Part of fixing #323.

Changes made to design:
* Using key "base" and value "1" to mark profiles in the base profile, rather than key "pprof::base" (or "pprof::diff") and value "true", because we don't want to use long values (the are lots of samples, so adding lots of tags might increase memory usage too much).
* Using `-diff_base`, rather than `-diff` as the flag name. It should be clearer.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -216,6 +216,36 @@ search for them in a directory pointed to by the environment variable
 * **-weblist= _regex_:** Generates a source/assembly combined annotated listing for
   functions matching *regex*, and starts a web browser to display it.
 
+## Comparing profiles
+
+pprof can subtract one profile's samples from another in order to compare them.
+For that, use the **-diff_base= _profile_** option, where *profile* is the filename
+or URL for the profile to be subtracted (referred to as the base profile). When
+using this mode, percentages reported in the output will be relative to the
+total of the base profile. To separate samples that are part of the diff base
+profile from those in the source profile, a string tag, "base"="1" is added to
+all samples in the base profile. This means that if one saves the merged profile
+and looks at it again with pprof, it will still be treated as a profile with a 
+diff base. It also means that any other samples which also have this tag will
+be treated as part of the diff base profile.
+
+If one is interested in the relative differences between the source and base
+profiles (for example, which profile has a larger percentage of CPU time used
+in a particular function), it can be useful to use the **-normalize** option.
+With this option, the source profile is scaled so that the total of samples in
+the source profile is equal to the total of samples in the base profile prior
+to merging the profiles.
+
+When comparing two cumulative profiles, for example two contention profiles
+collected from the same program at different times, use the **-base** flag
+in place of the **-diff_base** flag. When this flag is used and no negative
+values appear in the merged profile when aggregated at the address-level, 
+which would be the case when comparing cummulative profiles collected from the
+same program, percentages reported will be relative to the the difference
+between the total for the source profile and the total for the base profile.
+In the general case, percentages will be relative to the total of the absolute
+value of all samples when aggregated at the address level.
+
 # Fetching profiles
 
 pprof can read profiles from a file or directly from a URL over http. Its native
@@ -236,11 +266,6 @@ If multiple profiles are specified, pprof will fetch them all and merge
 them. This is useful to combine profiles from multiple processes of a
 distributed job. The profiles may be from different programs but must be
 compatible (for example, CPU profiles cannot be combined with heap profiles).
-
-pprof can subtract a profile from another in order to compare them. For that,
-use the **-base= _profile_** option, where *profile* is the filename or URL for the
-profile to be subtracted. This may result on some report entries having negative
-values.
 
 ## Symbolization
 
@@ -286,4 +311,3 @@ the symbolization handler.
 
 * **-symbolize=demangle=templates:** Demangle, and trim function parameters, but
   not template parameters.
-
--- a/doc/README.md
+++ b/doc/README.md
@@ -240,38 +240,42 @@ compatible (for example, CPU profiles cannot be combined with heap profiles).
 
 ## Comparing profiles
 
-pprof can subtract a profile from another, provided the profiles are of 
-compatible types, in order to compare them. For that, use the 
-**-diff_base= _profile_** option, where *profile* is the filename or URL for the
-profile to be subtracted. This may result in some report entries having negative
-values. Percentages in the report are relative to the total of 
-the diff base. 
+pprof can subtract one profile from another, provided the profiles are of
+compatible types. pprof has two options which can be used to specify the
+filename or URL for a profile to be subtracted from the source profile:
 
-If the merged profile is output as a protocol buffer, all samples in
-the diff base profile will have a label with the key "pprof::base" and a value
-of "true". If pprof is then used to look at the merged profile, it will behave
-as if separate source and base profiles were passed in.
+* **-diff_base= _profile_:** useful for comparing any two profiles of compatible
+types (i.e. any two heap profiles). Percentages in the output are relative to
+the total of samples in the diff base profile.
 
-The **-base** option can be used in the place of the **-diff_base** option, and
-may be preferrable when comparing two cumulative profiles, for example two
-contention profiles collected from the same program at different times. When 
-this flag is used and no negative values appear in the merged profile when
-aggregated at the address-level, percentages reported will be relative to the 
-the difference between the total for the source profile and the total for the
-base profile. In the general case, percentages will be relative to the total of
-the absolute value of all samples when aggregated at the address level.
+* **-base= _profile_:** useful for subtracting a cumulative profile, like a
+[golang block profile](https://golang.org/doc/diagnostics.html#profiling),
+from another cumulative profile collected from the same program at a later time.
+When comparing cumulative profiles collected on the same profile, percentages in
+the output are relative to the the difference between the total for the source
+profile and the total for the base profile.
 
-The **-normalize** option can be useful for determining the relative differences 
-between profiles, for example, which profile has a larger percentage of CPU time
-used in a particular function. With this option, the source profile is scaled so
-that the total of samples in the source profile is equal to the total of samples
-in the base profile prior to merging the profiles. 
+The following flag can be used when a base profile is specified with either the
+`-diff_base` or the `-base` option:
 
-This flag can be only be used when a diff base or a base profile is specified, 
-so for example:
+* **-normalize**: Scales the source profile so that that the total of samples 
+in the source profile is equal to the total of samples in the base profile prior
+to subtracting the base profile from the source profile. Useful for determining
+the relative differences between profiles, for example, which profile has a
+larger percentage of CPU time used in a particular function.
 
-    pprof <format> -normalize -diff_base=base source
-    
+When using the **-diff_base** option, some report entries may have negative
+values. If the merged profile is output as a protocol buffer, all samples in the
+diff base profile will have a label with the key "pprof::base" and a value of
+"true". If pprof is then used to look at the merged profile, it will behave as
+if separate source and base profiles were passed in.
+
+When using the **-base** option to subtract one cumulative profile from another
+collected on the same program at a later time, percentages will be relative to
+the difference between the total for the source profile and the total for
+the base profile, and all values will be positive. In the general case, some 
+report entries may have negative values and percentages will be relative to the
+total of the absolute value of all samples when aggregated at the address level.
 
 ## Symbolization
 
--- a/doc/README.md
+++ b/doc/README.md
@@ -216,35 +216,9 @@ search for them in a directory pointed to by the environment variable
 * **-weblist= _regex_:** Generates a source/assembly combined annotated listing for
   functions matching *regex*, and starts a web browser to display it.
 
-## Comparing profiles
-
-pprof can subtract one profile's samples from another in order to compare them.
-For that, use the **-diff_base= _profile_** option, where *profile* is the filename
-or URL for the profile to be subtracted (referred to as the base profile). When
-using this mode, percentages reported in the output will be relative to the
-total of the base profile. To separate samples that are part of the diff base
-profile from those in the source profile, a string tag, "base"="1" is added to
-all samples in the base profile. This means that if one saves the merged profile
-and looks at it again with pprof, it will still be treated as a profile with a 
-diff base. It also means that any other samples which also have this tag will
-be treated as part of the diff base profile.
-
-If one is interested in the relative differences between the source and base
-profiles (for example, which profile has a larger percentage of CPU time used
-in a particular function), it can be useful to use the **-normalize** option.
-With this option, the source profile is scaled so that the total of samples in
-the source profile is equal to the total of samples in the base profile prior
-to merging the profiles.
-
-When comparing two cumulative profiles, for example two contention profiles
-collected from the same program at different times, use the **-base** flag
-in place of the **-diff_base** flag. When this flag is used and no negative
-values appear in the merged profile when aggregated at the address-level, 
-which would be the case when comparing cummulative profiles collected from the
-same program, percentages reported will be relative to the the difference
-between the total for the source profile and the total for the base profile.
-In the general case, percentages will be relative to the total of the absolute
-value of all samples when aggregated at the address level.
+Samples in a profile may have tags. These tags have a name and a value; this
+value can be either numeric or a string. pprof can select samples from a
+profile based on these tags using the `-tagfocus` and `-tagignore` options.
 
 # Fetching profiles
 
@@ -266,6 +240,42 @@ If multiple profiles are specified, pprof will fetch them all and merge
 them. This is useful to combine profiles from multiple processes of a
 distributed job. The profiles may be from different programs but must be
 compatible (for example, CPU profiles cannot be combined with heap profiles).
+
+
+## Comparing profiles
+
+pprof can subtract a profile from another, provided the profiles are of 
+compatible types, in order to compare them. For that, use the 
+**-diff_base= _profile_** option, where *profile* is the filename or URL for the
+profile to be subtracted. This may result in some report entries having negative
+values.  Percentages in the report are relative to the total of 
+the diff base. 
+
+If the merged profile is output as a protocol buffer, all samples in 
+the diff base profile will have a label with the key "base" and a value of "1".
+If pprof is then used to look at the merged profile, it will behave as if 
+separate source and base profiles were passed in. 
+
+The **-base** option can be used in the place of the **-diff_base** option, and
+may be preferrable when comparing two cumulative profiles, for example two
+contention profiles collected from the same program at different times. When 
+this flag is used and no negative values appear in the merged profile when
+aggregated at the address-level, percentages reported will be relative to the 
+the difference between the total for the source profile and the total for the
+base profile. In the general case, percentages will be relative to the total of
+the absolute value of all samples when aggregated at the address level.
+
+The **-normalize** option can be useful for determining the relative differences 
+between profiles, for example, which profile has a larger percentage of CPU time
+used in a particular function. With this option, the source profile is scaled so
+that the total of samples in the source profile is equal to the total of samples
+in the base profile prior to merging the profiles. 
+
+This flag can be only be used when a diff base or a base profile is specified, 
+so for example:
+
+    pprof <format> -normalize -diff_base=base source
+    
 
 ## Symbolization
 
--- a/doc/README.md
+++ b/doc/README.md
@@ -216,6 +216,44 @@ search for them in a directory pointed to by the environment variable
 * **-weblist= _regex_:** Generates a source/assembly combined annotated listing for
   functions matching *regex*, and starts a web browser to display it.
 
+## Comparing profiles
+
+pprof can subtract one profile from another, provided the profiles are of
+compatible types (i.e. two heap profiles). pprof has two options which can be
+used to specify the filename or URL for a profile to be subtracted from the
+source profile:
+
+* **-diff_base= _profile_:** useful for comparing two profiles. Percentages in
+the output are relative to the total of samples in the diff base profile.
+
+* **-base= _profile_:** useful for subtracting a cumulative profile, like a
+[golang block profile](https://golang.org/doc/diagnostics.html#profiling),
+from another cumulative profile collected from the same program at a later time.
+When comparing cumulative profiles collected on the same profile, percentages in
+the output are relative to the difference between the total for the source
+profile and the total for the base profile.
+
+The **-normalize** flag can be used when a base profile is specified with either 
+the `-diff_base` or the `-base` option. This flag scales the source profile so
+that the total of samples in the source profile is equal to the total of samples
+in the base profile prior to subtracting the base profile from the source
+profile. Useful for determining the relative differences between profiles, for
+example, which profile has a larger percentage of CPU time used in a particular
+function.
+
+When using the **-diff_base** option, some report entries may have negative
+values. If the merged profile is output as a protocol buffer, all samples in the
+diff base profile will have a label with the key "pprof::base" and a value of
+"true". If pprof is then used to look at the merged profile, it will behave as
+if separate source and base profiles were passed in.
+
+When using the **-base** option to subtract one cumulative profile from another
+collected on the same program at a later time, percentages will be relative to
+the difference between the total for the source profile and the total for
+the base profile, and all values will be positive. In the general case, some 
+report entries may have negative values and percentages will be relative to the
+total of the absolute value of all samples when aggregated at the address level.
+
 # Fetching profiles
 
 pprof can read profiles from a file or directly from a URL over http. Its native
@@ -236,46 +274,6 @@ If multiple profiles are specified, pprof will fetch them all and merge
 them. This is useful to combine profiles from multiple processes of a
 distributed job. The profiles may be from different programs but must be
 compatible (for example, CPU profiles cannot be combined with heap profiles).
-
-
-## Comparing profiles
-
-pprof can subtract one profile from another, provided the profiles are of
-compatible types. pprof has two options which can be used to specify the
-filename or URL for a profile to be subtracted from the source profile:
-
-* **-diff_base= _profile_:** useful for comparing any two profiles of compatible
-types (i.e. any two heap profiles). Percentages in the output are relative to
-the total of samples in the diff base profile.
-
-* **-base= _profile_:** useful for subtracting a cumulative profile, like a
-[golang block profile](https://golang.org/doc/diagnostics.html#profiling),
-from another cumulative profile collected from the same program at a later time.
-When comparing cumulative profiles collected on the same profile, percentages in
-the output are relative to the the difference between the total for the source
-profile and the total for the base profile.
-
-The following flag can be used when a base profile is specified with either the
-`-diff_base` or the `-base` option:
-
-* **-normalize**: Scales the source profile so that that the total of samples 
-in the source profile is equal to the total of samples in the base profile prior
-to subtracting the base profile from the source profile. Useful for determining
-the relative differences between profiles, for example, which profile has a
-larger percentage of CPU time used in a particular function.
-
-When using the **-diff_base** option, some report entries may have negative
-values. If the merged profile is output as a protocol buffer, all samples in the
-diff base profile will have a label with the key "pprof::base" and a value of
-"true". If pprof is then used to look at the merged profile, it will behave as
-if separate source and base profiles were passed in.
-
-When using the **-base** option to subtract one cumulative profile from another
-collected on the same program at a later time, percentages will be relative to
-the difference between the total for the source profile and the total for
-the base profile, and all values will be positive. In the general case, some 
-report entries may have negative values and percentages will be relative to the
-total of the absolute value of all samples when aggregated at the address level.
 
 ## Symbolization
 
--- a/doc/README.md
+++ b/doc/README.md
@@ -229,7 +229,7 @@ the output are relative to the total of samples in the diff base profile.
 * **-base= _profile_:** useful for subtracting a cumulative profile, like a
 [golang block profile](https://golang.org/doc/diagnostics.html#profiling),
 from another cumulative profile collected from the same program at a later time.
-When comparing cumulative profiles collected on the same profile, percentages in
+When comparing cumulative profiles collected on the same program, percentages in
 the output are relative to the difference between the total for the source
 profile and the total for the base profile.
 
--- a/doc/README.md
+++ b/doc/README.md
@@ -216,10 +216,6 @@ search for them in a directory pointed to by the environment variable
 * **-weblist= _regex_:** Generates a source/assembly combined annotated listing for
   functions matching *regex*, and starts a web browser to display it.
 
-Samples in a profile may have tags. These tags have a name and a value; this
-value can be either numeric or a string. pprof can select samples from a
-profile based on these tags using the `-tagfocus` and `-tagignore` options.
-
 # Fetching profiles
 
 pprof can read profiles from a file or directly from a URL over http. Its native
@@ -248,13 +244,13 @@ pprof can subtract a profile from another, provided the profiles are of
 compatible types, in order to compare them. For that, use the 
 **-diff_base= _profile_** option, where *profile* is the filename or URL for the
 profile to be subtracted. This may result in some report entries having negative
-values.  Percentages in the report are relative to the total of 
+values. Percentages in the report are relative to the total of 
 the diff base. 
 
-If the merged profile is output as a protocol buffer, all samples in 
-the diff base profile will have a label with the key "base" and a value of "1".
-If pprof is then used to look at the merged profile, it will behave as if 
-separate source and base profiles were passed in. 
+If the merged profile is output as a protocol buffer, all samples in
+the diff base profile will have a label with the key "pprof::base" and a value
+of "true". If pprof is then used to look at the merged profile, it will behave
+as if separate source and base profiles were passed in.
 
 The **-base** option can be used in the place of the **-diff_base** option, and
 may be preferrable when comparing two cumulative profiles, for example two
--- a/internal/driver/cli.go
+++ b/internal/driver/cli.go
@@ -142,7 +142,7 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 		Comment:      *flagAddComment,
 	}
 
-	if err := source.addBaseProfiles(flagBase, flagDiffBase); err != nil {
+	if err := source.addBaseProfiles(*flagBase, *flagDiffBase); err != nil {
 		return nil, nil, err
 	}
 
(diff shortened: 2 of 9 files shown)


internal/report/report.go
```diff
@@ -261,6 +261,21 @@ func (rpt *Report) newGraph(nodes graph.NodeSet) *graph.Graph {
 		}
 		s.NumLabel = numLabels
 		s.NumUnit = numUnits
+
+		// Also remove pprof::diff=true string tag.
```
aalexand: It would be good to clarify why. Also this code is not super clear. Perhaps make a separate function, and look into how to simplify it. I would also look into whether code could be simpler if the pprof::diff tag was removed in the aggregate() code.

internal/report/report_test.go
```diff
@@ -285,3 +285,122 @@ func TestLegendActiveFilters(t *testing.T) {
 		}
 	}
 }
+
+func TestComputTotal(t *testing.T) {
```
aalexand: Typo

internal/report/report.go
```diff
@@ -1211,10 +1226,10 @@ func NewDefault(prof *profile.Profile, options Options) *Report {
 	return New(prof, o)
 }
 
-// computeTotal computes the sum of all sample values. This will be
-// used to compute percentages.
+// computeTotal computes the sum of the absolute value of all sample values.
```
aalexand: Doc should be updated when such non-trivial logic is added in the function code.

internal/report/report.go
```diff
@@ -1224,13 +1239,19 @@ func computeTotal(prof *profile.Profile, value, meanDiv func(v []int64) int64) i
 		if v < 0 {
 			v = -v
 		}
-		ret += v
+		total += v
+		if sample.HasTag("pprof::diff", "true") {
+			diffTotal += v
+		}
 		div += d
```
aalexand: Assuming that it's intentional that accumulating the div is not under the tag condition, can you add a comment on why so?

profile/profile.go
```diff
@@ -674,6 +674,36 @@ func numLabelsToString(numLabels map[string][]int64, numUnits map[string][]strin
 	return strings.Join(ls, " ")
 }
 
+// AddTag adds a tag with the specified key and value to all samples in the
+// profile.
+func (p *Profile) AddTag(key, value string) {
```
aalexand: I'd maybe do this as SetTag(), it would simplify the code and I don't think we really need to support multiple values per tag in this PR.

internal/report/report_test.go
```diff
@@ -285,3 +285,122 @@ func TestLegendActiveFilters(t *testing.T) {
...
+			Label:    map[string][]string{"pprof::diff": {"true"}},
+		},
+		{
+			Location: []*profile.Location{testL[2], testL[1], testL[0]},
+			Value:    []int64{1000, -10},
+		},
+		{
+			Location: []*profile.Location{testL[2], testL[1], testL[0]},
+			Value:    []int64{-9000, 3},
+			Label:    map[string][]string{"pprof::diff": {"true"}},
```
aalexand: Given that "pprof::diff" and "true" are going to be stored as strings for each sample, I wonder if we should come up with shorter strings to save memory.

internal/report/report.go
nolanmar511: Would removing the tag in aggregation() prevent it from being used in later to separate out source and base profiles?

For now, I'm separating this into a different function.

internal/report/report_test.go
nolanmar511: I agree -- maybe "1"?

profile/profile.go
nolanmar511: Done.

profile/profile.go
```diff
@@ -674,6 +674,36 @@ func numLabelsToString(numLabels map[string][]int64, numUnits map[string][]strin
...
+	for _, sample := range p.Sample {
+		if sample.HasTag(key, value) {
```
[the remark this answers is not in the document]
nolanmar511: Changing AddTag to SetTag, so I think this no longer applies.

internal/report/report_test.go
nolanmar511: Done.

internal/report/report.go
nolanmar511: Using the SetTag function to clear pprof::diff tag.
And adding comment to explain why this tag shouldn't be set.

internal/report/report.go
```diff
@@ -261,6 +261,21 @@ func (rpt *Report) newGraph(nodes graph.NodeSet) *graph.Graph {
 		}
 		s.NumLabel = numLabels
 		s.NumUnit = numUnits
+
+		// Also remove pprof::diff=true string tag.
+		if values, ok := s.Label["pprof::diff"]; ok {
+			newVals := []string{}
```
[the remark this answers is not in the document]
nolanmar511: This code has been removed.

internal/report/report.go
nolanmar511: done.

nolanmar511:
PTAL

doc/pprof.md
```diff
@@ -214,6 +214,10 @@ search for them in a directory pointed to by the environment variable
 * **-weblist= _regex_:** Generates a source/assembly combined annotated listing for
   functions matching *regex*, and starts a web browser to display it.
 
+Samples in a profile may have tags. These tags have a name and a value; this
```
aalexand: This added paragraph sounds a bit out of tune in this "Annotated code" section. Why here?

doc/pprof.md
```diff
@@ -235,10 +239,41 @@ them. This is useful to combine profiles from multiple processes of a
...
+the diff base. 
+
+If the merged profile is output as a protocol buffer, all samples in
+the diff base profile will have a label with the key "pprof::base" and a value
+of "true". If pprof is then used to look at the merged profile, it will behave
+as if separate source and base profiles were passed in.
+
+The **-base** option can be used in the place of the **-diff_base** option, and
+may be preferrable when comparing two cumulative profiles, for example two
+contention profiles collected from the same program at different times. When
```
aalexand: This should be more specific that we are talking about contention reports from Go programs. Linking to Go's documentation would be good.

doc/pprof.md
```diff
@@ -235,10 +239,41 @@ them. This is useful to combine profiles from multiple processes of a
...
+If the merged profile is output as a protocol buffer, all samples in
+the diff base profile will have a label with the key "pprof::base" and a value
+of "true". If pprof is then used to look at the merged profile, it will behave
+as if separate source and base profiles were passed in.
+
+The **-base** option can be used in the place of the **-diff_base** option, and
+may be preferrable when comparing two cumulative profiles, for example two
+contention profiles collected from the same program at different times. When 
+this flag is used and no negative values appear in the merged profile when
+aggregated at the address-level, percentages reported will be relative to the
```
aalexand: I'd somehow leave out the detail about address level or move it closer to the end of the section. Telling the whole story feels overwhelming a bit.

doc/pprof.md
```diff
@@ -235,10 +239,41 @@ them. This is useful to combine profiles from multiple processes of a
...
 compatible (for example, CPU profiles cannot be combined with heap profiles).
 
-pprof can subtract a profile from another in order to compare them. For that,
-use the **-base= _profile_** option, where *profile* is the filename or URL for the
-profile to be subtracted. This may result on some report entries having negative
-values.
+
+## Comparing profiles
+
+pprof can subtract a profile from another, provided the profiles are of
```
aalexand: Maybe start this section with a very high-level note that there are two ways to compare profiles, briefly characterize each of them, then talk about each in more detail.

doc/pprof.md
```diff
@@ -235,10 +239,41 @@ them. This is useful to combine profiles from multiple processes of a
...
-profile to be subtracted. This may result on some report entries having negative
-values.
+
+## Comparing profiles
+
+pprof can subtract a profile from another, provided the profiles are of 
+compatible types, in order to compare them. For that, use the 
+**-diff_base= _profile_** option, where *profile* is the filename or URL for the
+profile to be subtracted. This may result in some report entries having negative
+values.  Percentages in the report are relative to the total of
```
aalexand: Nit: double space is a typo?

doc/pprof.md
```diff
@@ -235,10 +239,41 @@ them. This is useful to combine profiles from multiple processes of a
...
+The **-base** option can be used in the place of the **-diff_base** option, and
+may be preferrable when comparing two cumulative profiles, for example two
+contention profiles collected from the same program at different times. When 
+this flag is used and no negative values appear in the merged profile when
+aggregated at the address-level, percentages reported will be relative to the 
+the difference between the total for the source profile and the total for the
+base profile. In the general case, percentages will be relative to the total of
+the absolute value of all samples when aggregated at the address level.
+
+The **-normalize** option can be useful for determining the relative differences
```
aalexand: Did we document -normalize elsewhere before? I'd make sure we don't double-document it now.

internal/driver/cli.go
```diff
@@ -140,12 +142,30 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 		Comment:      *flagAddComment,
 	}
 
+	var base []string
```
aalexand: This fragment of code of dealing with base and diffBase flags is a bit long now. Consider refactoring it.

internal/report/report.go
```diff
@@ -1225,13 +1230,21 @@ func computeTotal(prof *profile.Profile, value, meanDiv func(v []int64) int64) i
 		if v < 0 {
 			v = -v
 		}
-		ret += v
+		total += v
+		if sample.HasTag("pprof::base", "true") {
+			diffTotal += v
+			diffDiv += d
+		}
 		div += d
```
aalexand: I think I had this comment but it was not answered before: Assuming that it's intentional that accumulating the div is not under the tag condition, can you add a comment on why so?

profile/profile.go
```diff
@@ -674,6 +674,29 @@ func numLabelsToString(numLabels map[string][]int64, numUnits map[string][]strin
 	return strings.Join(ls, " ")
 }
 
+// SetTag sets the specified key to the specified value for all samples in the
+// profile.
+func (p *Profile) SetTag(key string, value []string) {
+	for _, sample := range p.Sample {
+		if sample.Label == nil {
+			sample.Label = make(map[string][]string)
```
aalexand: `sample.Label = map[string]string{}` is a bit cheaper.

profile/profile_test.go
```diff
@@ -912,6 +945,134 @@ func locationHash(s *Sample) string {
...
+	var testcases = []struct {
+		desc       string
+		sample     *Sample
+		key        string
+		value      string
+		wantHasTag bool
+	}{
+		{
+			desc: "empty label does not have tag",
+			sample: &Sample{
```
aalexand: I think you could avoid repeating the Sample and its Location and Value in every test case. You only care about Label field, you could reuse the rest.

profile/profile_test.go
```diff
@@ -912,6 +945,134 @@ func locationHash(s *Sample) string {
...
+func TestSetTag(t *testing.T) {
+	var testcases = []struct {
+		desc     string
+		profile  *Profile
+		setKey   string
+		setVal   []string
+		wantTags []map[string][]string
+	}{
+		{
+			desc:    "some samples have tag already",
```
aalexand: Did you mean to add more test cases? E.g. what if sample doesn't have the tags.

doc/pprof.md
nolanmar511: Made an error when moving docs around. Fixed.

doc/pprof.md
nolanmar511: We haven't documented normalize before.

internal/report/report.go
nolanmar511: Fixed.

internal/driver/cli.go
nolanmar511: Did some refactoring.

profile/profile.go
nolanmar511: done.

profile/profile_test.go
nolanmar511: Done.
Also did some refactoring of this test.

nolanmar511:
PTAL, except for at docs.
(Will add another PTAL when docs are ready to be looked at again)

doc/pprof.md
nolanmar511: I can't find much in go documentation on contention reports.

This is the main thing I found (from [here](https://golang.org/src/runtime/pprof/pprof.go?h=block+profile)):
```
//	goroutine    - stack traces of all current goroutines
//	heap         - a sampling of all heap allocations
//	threadcreate - stack traces that led to the creation of new OS threads
//	block        - stack traces that led to blocking on synchronization primitives
//	mutex        - stack traces of holders of contended mutexes
```

I'm guessing that block profiles are what we would consider contention profiles?
I could look for blog posts?

nolanmar511:
Also addressed (tried to address) comments on docs. 
Those are ready for PTAL, too.

doc/pprof.md
aalexand: Maybe link to https://golang.org/doc/diagnostics.html#profiling? It's not exclusively about contention but has something.

doc/README.md
```diff
@@ -237,10 +237,45 @@ them. This is useful to combine profiles from multiple processes of a
 distributed job. The profiles may be from different programs but must be
 compatible (for example, CPU profiles cannot be combined with heap profiles).
 
-pprof can subtract a profile from another in order to compare them. For that,
-use the **-base= _profile_** option, where *profile* is the filename or URL for the
-profile to be subtracted. This may result on some report entries having negative
-values.
+
+## Comparing profiles
```
aalexand: It is a bit strange that "Comparing profiles" is a section under "Fetching profiles", is it intentional? I'd expect it to be more related to viewing than to fetching.

doc/README.md
```diff
@@ -237,10 +237,45 @@ them. This is useful to combine profiles from multiple processes of a
...
+
+pprof can subtract one profile from another, provided the profiles are of
+compatible types. pprof has two options which can be used to specify the
+filename or URL for a profile to be subtracted from the source profile:
+
+* **-diff_base= _profile_:** useful for comparing any two profiles of compatible
```
aalexand: Both -diff_base and -base require the profiles to be compatible so the focus on the compatibility of profiles in this paragraph is somewhat unexpected to me? I'd mention that in the paragraph above.

doc/README.md
```diff
@@ -237,10 +237,45 @@ them. This is useful to combine profiles from multiple processes of a
...
+types (i.e. any two heap profiles). Percentages in the output are relative to
+the total of samples in the diff base profile.
+
+* **-base= _profile_:** useful for subtracting a cumulative profile, like a
+[golang block profile](https://golang.org/doc/diagnostics.html#profiling),
+from another cumulative profile collected from the same program at a later time.
+When comparing cumulative profiles collected on the same profile, percentages in
+the output are relative to the the difference between the total for the source
```
aalexand: "the the"

doc/README.md
```diff
@@ -237,10 +237,45 @@ them. This is useful to combine profiles from multiple processes of a
...
+profile and the total for the base profile.
+
+The following flag can be used when a base profile is specified with either the
```
aalexand: Since there is only one flag listed here, the pattern "The following flag can be used ... : <flag description>" looks a bit heavyweight - this pattern would make sense if there would be multiple flags, but for one it doesn't.

internal/report/report.go
nolanmar511: fixed.

doc/pprof.md
nolanmar511: Tried to change this.
Possibly fixed.

doc/pprof.md
nolanmar511: fixed.

internal/report/report.go
```diff
@@ -264,6 +264,10 @@ func (rpt *Report) newGraph(nodes graph.NodeSet) *graph.Graph {
 		s.NumUnit = numUnits
 	}
 
+	// Remove tag marking samples from the base profiles, so it does not appear
+	// as a nodelet in the graph view.
+	prof.SetTag("pprof::base", []string{})
```
aalexand: I think this will set the tag on all samples even on those that do not have it, needlessly consuming memory?

profile/profile.go
```diff
@@ -674,6 +674,29 @@ func numLabelsToString(numLabels map[string][]int64, numUnits map[string][]strin
 	return strings.Join(ls, " ")
 }
 
+// SetTag sets the specified key to the specified value for all samples in the
+// profile.
+func (p *Profile) SetTag(key string, value []string) {
```
aalexand: The field is called "Label", should SetTag and HasTag be SetLabel and HasLabel?

doc/README.md
nolanmar511: Makes sense. Moved.

doc/README.md
nolanmar511: done.

doc/README.md
nolanmar511: Done -- thanks

doc/README.md
nolanmar511: Done.

internal/driver/cli.go
```diff
@@ -158,6 +158,37 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	return source, cmd, nil
 }
 
+// addBaseProfiles adds the list of base profiles or diff base profiles to
+// the source. This function will return an error if both base and diff base
+// profiles are specified.
+func (source *source) addBaseProfiles(flagBase, flagDiffBase *[]*string) error {
```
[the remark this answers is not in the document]
nolanmar511: Fixed.

internal/driver/cli.go
```diff
@@ -158,6 +158,37 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
...
+	if len(diffBase) > 0 {
+		source.Base = diffBase
+		source.DiffBase = true
+	}
+	return nil
+}
+
+// parseStringListFlag list takes a StringList flag, and outputs an array of non-empty
+// strings associated with the flag.
+func parseStringListFlag(list *[]*string) []string {
```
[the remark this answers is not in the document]
nolanmar511: done.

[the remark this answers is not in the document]
nolanmar511: That's much better!
Done.

internal/driver/fetch.go
```diff
@@ -57,7 +57,7 @@ func fetchProfiles(s *source, o *plugin.Options) (*profile.Profile, error) {
 		})
 	}
 
-	p, pbase, m, mbase, save, err := grabSourcesAndBases(sources, bases, o.Fetch, o.Obj, o.UI)
+	p, pbase, m, mbase, save, err := grabSourcesAndBases(sources, bases, s.DiffBase, o.Fetch, o.Obj, o.UI)
```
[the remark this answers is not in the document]
nolanmar511: We had had some concerns about fetchProfiles getting to large.
I think this is cleaner, though.

internal/report/report.go
nolanmar511: It would.
Fixed.

profile/profile.go
nolanmar511: Fixed.

nolanmar511:
PTAL

internal/driver/cli.go
```diff
@@ -158,6 +158,37 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
...
 
+// addBaseProfiles adds the list of base profiles or diff base profiles to
+// the source. This function will return an error if both base and diff base
+// profiles are specified.
+func (source *source) addBaseProfiles(flagBase, flagDiffBase []*string) error {
+	base := dropEmpty(flagBase)
+	diffBase := dropEmpty(flagDiffBase)
+
+	if len(base) > 0 && len(diffBase) > 0 {
+		return fmt.Errorf("-base and -diff_base flags cannot both be specified")
```
aalexand: When formatting is not needed, use errors.New()

internal/driver/cli.go
```diff
@@ -158,6 +158,37 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	return source, cmd, nil
 }
 
+// addBaseProfiles adds the list of base profiles or diff base profiles to
+// the source. This function will return an error if both base and diff base
+// profiles are specified.
+func (source *source) addBaseProfiles(flagBase, flagDiffBase []*string) error {
+	base := dropEmpty(flagBase)
```
aalexand: Nit: base, diffBase := dropEmpty(flagBase), dropEmpty(flagDiffBase)

internal/driver/cli.go
```diff
@@ -158,6 +158,37 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
...
+	diffBase := dropEmpty(flagDiffBase)
+
+	if len(base) > 0 && len(diffBase) > 0 {
+		return fmt.Errorf("-base and -diff_base flags cannot both be specified")
+	}
+
+	source.Base = base
+	if len(diffBase) > 0 {
+		source.Base = diffBase
```
aalexand: Maybe: `source.Base, source.DiffBase = diffBase, true`

internal/driver/fetch_test.go
```diff
@@ -210,112 +210,299 @@ func TestFetchWithBase(t *testing.T) {
...
 	}
 
 	testcases := []testcase{
 		{
 			"not normalized base is same as source",
 			[]string{path + "cppbench.contention"},
 			[]string{path + "cppbench.contention"},
+			[]string{},
+			false,
+			[]WantSample{},
```
aalexand: Use `nil` for empty slices.

internal/driver/fetch_test.go
```diff
@@ -210,112 +210,299 @@ func TestFetchWithBase(t *testing.T) {
...
+		src, _, err := parseFlags(o)
 
-			p, err := fetchProfiles(src, o)
-			pprofVariables = baseVars
-			if err != nil {
-				t.Fatal(err)
+		if tc.wantErrorMsg != "" {
+			if err == nil {
+				t.Errorf("%s: want error %q, got nil", tc.desc, tc.wantErrorMsg)
+				continue
```
aalexand: Use t.Run() to run the test cases? Then you can just use t.Fatalf() instead of t.Errorf() + continue. And you can avoid the boilerplate of formatting in tc.desc manually.

internal/driver/fetch_test.go
```diff
@@ -210,112 +210,299 @@ func TestFetchWithBase(t *testing.T) {
...
+			t.Errorf("%s: %v", tc.desc, err)
+			continue
+		}
+
+		if want, got := len(tc.wantSamples), len(p.Sample); want != got {
+			t.Errorf("%s: want %d samples got %d", tc.desc, want, got)
+			continue
+		}
+
+		if len(p.Sample) > 0 {
```
aalexand: I think this check is redundant - if p.Sample is an empty slice the loop will immediately terminate anyway.

internal/report/report_test.go
```diff
@@ -287,3 +287,122 @@ func TestLegendActiveFilters(t *testing.T) {
...
+			},
+			wantTotal: 9111,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.desc, func(t *testing.T) {
+			gotTotal := computeTotal(tc.prof, tc.value, tc.meanDiv)
+			if gotTotal != tc.wantTotal {
+				t.Errorf("want total %d, got %v", tc.wantTotal, gotTotal)
```
aalexand: "got" should go first, here and everywhere.

internal/driver/cli.go
nolanmar511: done.

internal/driver/cli.go
nolanmar511: done.

internal/driver/cli.go
nolanmar511: true

internal/driver/fetch_test.go
nolanmar511: done

internal/driver/fetch_test.go
nolanmar511: Done

internal/report/report_test.go
nolanmar511: Fixed.

internal/driver/fetch_test.go
nolanmar511: I'd had some issues with this test failing on xcode6 with against go master when running in parallel.
(https://travis-ci.org/google/pprof/jobs/374039614)

Will try to use `runtime.GOMAXPROCS` to increase parallelism.

nolanmar511:
PTAL

internal/driver/fetch_test.go
aalexand: "In parallel" - why? The only request was to use t.Run.

internal/driver/fetch_test.go
```diff
@@ -210,72 +210,250 @@ func TestFetchWithBase(t *testing.T) {
...
+			[]string{path + "cppbench.contention"},
+			[]string{path + "cppbench.contention"},
+			false,
+			[]WantSample{},
+			"-base and -diff_base flags cannot both be specified",
 		},
 	}
 
+	// Allow more test cases to run in parallel.
+	runtime.GOMAXPROCS(len(testcases))
```
aalexand: I do not understand why this is needed. You are not calling https://golang.org/pkg/testing/#T.Parallel here.

aalexand:
@nolanmar511 Please LMK when you remove the `runtime.GOMAXPROCS(len(testcases))` code and I'll look again.

nolanmar511:
Removed the `runtime.GOMAXPROCS(len(testcases))` code.
PTAL

internal/report/report.go
```diff
@@ -264,6 +264,10 @@ func (rpt *Report) newGraph(nodes graph.NodeSet) *graph.Graph {
 		s.NumUnit = numUnits
 	}
 
+	// Remove tag marking samples from the base profiles, so it does not appear
```
aalexand: tag -> label

internal/report/report.go
```diff
@@ -1212,10 +1216,11 @@ func NewDefault(prof *profile.Profile, options Options) *Report {
 	return New(prof, o)
 }
 
-// computeTotal computes the sum of all sample values. This will be
-// used to compute percentages.
+// computeTotal computes the sum of the absolute value of all sample values.
+// If any samples have the tag "pprof::base" with value "true", then the total
```
aalexand: tag -> label

internal/driver/fetch_test.go
```diff
@@ -289,30 +464,37 @@ func TestFetchWithBase(t *testing.T) {
 			})
 			src, _, err := parseFlags(o)
 
+			if tc.wantErrorMsg != "" {
+				if err == nil {
+					t.Fatalf("got nil, want error %q", tc.wantErrorMsg)
+				}
+				gotErrMsg := err.Error()
```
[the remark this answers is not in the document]
nolanmar511: Done.

internal/driver/fetch_test.go
```diff
@@ -289,30 +464,37 @@ func TestFetchWithBase(t *testing.T) {
...
 			p, err := fetchProfiles(src, o)
-			pprofVariables = baseVars
+
 			if err != nil {
 				t.Fatal(err)
 			}
 
-			if want, got := len(tc.expectedSamples), len(p.Sample); want != got {
-				t.Fatalf("want %d samples got %d", want, got)
+			if want, got := len(tc.wantSamples), len(p.Sample); want != got {
```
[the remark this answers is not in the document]
nolanmar511: Done

internal/report/report.go

internal/report/report.go

internal/report/report.go
nolanmar511: Sorry -- just pushed.

internal/driver/fetch_test.go
```diff
@@ -210,72 +210,247 @@ func TestFetchWithBase(t *testing.T) {
...
 		t.Run(tc.desc, func(t *testing.T) {
 			pprofVariables = baseVars.makeCopy()
-
 			base := make([]*string, len(tc.bases))
 			for i, s := range tc.bases {
 				base[i] = &s
 			}
 
+			diffBase := make([]*string, len(tc.diffBases))
+			for i, s := range tc.diffBases {
```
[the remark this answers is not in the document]
nolanmar511: Done.

nolanmar511:
PTAL

profile/profile.go
```diff
@@ -674,6 +674,39 @@ func numLabelsToString(numLabels map[string][]int64, numUnits map[string][]strin
 	return strings.Join(ls, " ")
 }
 
+// SetLabel sets the specified key to the specified value for all samples in the
+// profile.
+func (p *Profile) SetLabel(key string, value []string) {
+	for _, sample := range p.Sample {
+		if sample.Label == nil {
```
aalexand: Could avoid lookup on the "no key" case as
```
if sample.Label == nil {
  sample.Label = map[string][]string{key: value}
} else {
  sample.Label[key] = value
}
```

profile/profile.go
```diff
@@ -674,6 +674,39 @@ func numLabelsToString(numLabels map[string][]int64, numUnits map[string][]strin
...
+		}
+		sample.Label[key] = value
+	}
+}
+
+// RemoveLabel removes all labels associated with the specified key for all
+// samples in the profile.
+func (p *Profile) RemoveLabel(key string) {
+	for _, sample := range p.Sample {
+		if _, ok := sample.Label[key]; ok {
```
aalexand: https://blog.golang.org/go-maps-in-action: "The delete function doesn't return anything, and will do nothing if the specified key doesn't exist.", so I think you could drop the check?

profile/profile.go
```diff
@@ -674,6 +674,39 @@ func numLabelsToString(numLabels map[string][]int64, numUnits map[string][]strin
...
+			delete(sample.Label, key)
+		}
+	}
+}
+
+// HasLabel returns true if a sample has a label with indicated key and value.
+func (s *Sample) HasLabel(key, value string) bool {
+	if values, ok := s.Label[key]; ok {
```
aalexand: I think you could drop the "ok" check and just do
```
for _, v := range s.Label[key] {
```
as it will return nil slice if the key is missing which will immediately terminate the loop anyway.

profile/profile_test.go
```diff
@@ -912,6 +912,271 @@ func locationHash(s *Sample) string {
...
+			for i, sample := range profile.Sample {
+				wantLabels := tc.wantLabels[i]
+				if got, want := len(sample.Label), len(wantLabels); got != want {
+					t.Errorf("got %v label keys for sample %v, want %v", got, i, want)
+					continue
+				}
+				for wantKey, wantValues := range wantLabels {
+					if gotValues, ok := sample.Label[wantKey]; ok {
+						if !reflect.DeepEqual(gotValues, wantValues) {
+							t.Fatalf("for key %s, got values %v, want values %v", wantKey, gotValues, wantValues)
```
aalexand: Why Fatalf?

internal/driver/cli.go
```diff
@@ -140,15 +143,13 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
...
-			source.Base = append(source.Base, *s)
-		}
+	if err := source.addBaseProfiles(*flagBase, *flagDiffBase); err != nil {
+		return nil, nil, err
 	}
 
 	normalize := pprofVariables["normalize"].boolValue()
 	if normalize && len(source.Base) == 0 {
-		return nil, nil, fmt.Errorf("Must have base profile to normalize by")
+		return nil, nil, errors.New("Must have base profile to normalize by")
```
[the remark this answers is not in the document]
nolanmar511: Done

internal/driver/cli.go
```diff
@@ -158,6 +159,35 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
...
+	}
+
+	source.Base = base
+	if len(diffBase) > 0 {
+		source.Base, source.DiffBase = diffBase, true
+	}
+	return nil
+}
+
+// dropEmpty list takes a StringList flag, and outputs an array of non-empty
```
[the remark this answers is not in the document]
nolanmar511: Done.

internal/driver/cli.go
```diff
@@ -158,6 +159,35 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	return source, cmd, nil
 }
 
+// addBaseProfiles adds the list of base profiles or diff base profiles to
+// the source. This function will return an error if both base and diff base
+// profiles are specified.
+func (source *source) addBaseProfiles(flagBase, flagDiffBase []*string) error {
+	base, diffBase := dropEmpty(flagBase), dropEmpty(flagDiffBase)
+
```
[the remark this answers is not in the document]
nolanmar511: Done

internal/driver/driver_test.go
```diff
@@ -355,7 +355,11 @@ func (f testFlags) StringVar(p *string, s, d, c string) {
 
 func (f testFlags) StringList(s, d, c string) *[]*string {
 	if t, ok := f.stringLists[s]; ok {
-		return &t
+		tp := make([]*string, len(t))
```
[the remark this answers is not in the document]
nolanmar511: This was to address a previous comment for simplifying testing.
I changed the type of the `stringLists` field in the `testFlags` structure from `map[string][]*string` to `map[string][]string` to make it easier to construct flags for tests. 

Added a comment.

internal/report/report_test.go
```diff
@@ -287,3 +287,122 @@ func TestLegendActiveFilters(t *testing.T) {
...
+			value: func(v []int64) int64 {
+				return v[0]
+			},
+			wantTotal: 9111,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.desc, func(t *testing.T) {
+			gotTotal := computeTotal(tc.prof, tc.value, tc.meanDiv)
```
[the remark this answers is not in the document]
nolanmar511: Done.

profile/profile.go
nolanmar511: Done.

profile/profile.go
nolanmar511: Done

profile/profile.go
nolanmar511: Done.

profile/profile_test.go
nolanmar511: Errorf works. 
Done.

nolanmar511:
PTAL

internal/driver/fetch_test.go
```diff
@@ -210,72 +210,237 @@ func TestFetchWithBase(t *testing.T) {
...
+			"",
 		},
 		{
 			"normalized single source, multiple base (all profiles same)",
 			[]string{path + "cppbench.contention"},
 			[]string{path + "cppbench.contention", path + "cppbench.contention"},
+			nil,
 			true,
-			[][]int64{},
+			[]WantSample{},
```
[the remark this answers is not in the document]
nolanmar511: Done.

internal/driver/fetch_test.go
```diff
@@ -289,30 +454,37 @@ func TestFetchWithBase(t *testing.T) {
...
+
+				if gotErrMsg := err.Error(); gotErrMsg != tc.wantErrorMsg {
+					t.Fatalf("got error %q, want error %q", gotErrMsg, tc.wantErrorMsg)
+				}
+				return
+			}
+
 			if err != nil {
-				t.Fatalf("%s: %v", tc.desc, err)
+				t.Fatal(err)
```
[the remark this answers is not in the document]
nolanmar511: Done.

nolanmar511:
PTAL

PR #371, opened 2018-05-03
Add GetBase support for ASLR kernel mappings

rauls5382:
ASLR kernel mappings include offset == start.
Currently GetBase() cannot handle such a case.

This behavior can be seen from a perf.report data with samples in the kernel:

$ sudo perf report -D perf.data | grep MMAP
0 0x108 [0x50]: PERF_RECORD_MMAP -1/0: [0xffffffffaf800000(0x1087a000) @ 0xffffffffaf800000]: x [kernel.kallsyms]_text

Also added a testcase to reproduce and document.


internal/elfexec/elfexec.go
```diff
@@ -208,13 +208,13 @@ func GetBase(fh *elf.FileHeader, loadSegment *elf.ProgHeader, stextOffset *uint6
...
 		}
-		if start >= loadSegment.Vaddr && limit > start && (offset == 0 || offset == pageOffsetPpc64) {
+		if start >= loadSegment.Vaddr && limit > start && (offset == 0 || offset == pageOffsetPpc64 || offset == start) {
 			// Some kernels look like:
 			//       VADDR=0xffffffff80200000
 			// stextOffset=0xffffffff80200198
 			//       Start=0xffffffff83200000
 			//       Limit=0xffffffff84200000
-			//      Offset=0 (0xc000000000000000 for PowerPC64)
+			//      Offset=0 (0xc000000000000000 for PowerPC64) (==Start for Linux Perf)
```
aalexand: I think this case corresponds to when KASLR is active, so I'd reflect that in the comment and PR description. Some experimentation:
```
$ sudo cat /proc/kallsyms | grep _stext
ffffffff866002b8 T _stext
$ sudo perf record sleep 5
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.018 MB perf.data (7 samples) ]
$ pprof -raw perf.data | tail -3
Mappings
1: 0x0/0x1/0x0 perf_4.9 <build ID redacted> 
2: 0xffffffff86600000/0xffffffffc030a000/0xffffffff86600000 [kernel.kallsyms] <build ID redacted>
```
Then rebooted the machine and did the same (note changed addresses):
```
$ sudo cat /proc/kallsyms | grep _stext
[sudo] password for aalexand: 
ffffffff892002b8 T _stext
$ sudo perf record sleep 5
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.018 MB perf.data (2 samples) ]
$ pprof -raw perf.data | tail -3
Mappings
1: 0x0/0x1/0x0 perf_4.9 <build ID redacted> 
2: 0xffffffff89200000/0xffffffffc0338000/0xffffffff89200000 [kernel.kallsyms] <build ID redacted>
```

internal/elfexec/elfexec_test.go
```diff
@@ -55,6 +55,8 @@ func TestGetBase(t *testing.T) {
 		{"exec offset 2", fhExec, lsOffset, nil, 0x200000, 0x600000, 0, 0, false},
 		{"exec nomap", fhExec, nil, nil, 0, 0, 0, 0, false},
 		{"exec kernel", fhExec, kernelHeader, uint64p(0xffffffff81000198), 0xffffffff82000198, 0xffffffff83000198, 0, 0x1000000, false},
+		{"exec kernel perf offset0", fhExec, kernelHeader, uint64p(0xffffffff810002b8), 0xffffffff81000000, 0xffffffffa0000000, 0x0, 0x0, false},
+		{"exec kernel perf", fhExec, kernelHeader, uint64p(0xffffffff810002b8), 0xffffffff81000000, 0xffffffffa0000000, 0xffffffff81000000, 0x0, false},
 		{"exec PPC64 kernel", fhExec, ppc64KernelHeader, uint64p(0xc000000000000000), 0xc000000000000000, 0xd00000001a730000, 0xc000000000000000, 0x0, false},
```
aalexand: Looks like this case passes the new check so the condition for pageOffsetPpc64 and the constant can be removed?

internal/elfexec/elfexec_test.go
```diff
@@ -55,6 +55,8 @@ func TestGetBase(t *testing.T) {
 		{"exec offset 2", fhExec, lsOffset, nil, 0x200000, 0x600000, 0, 0, false},
 		{"exec nomap", fhExec, nil, nil, 0, 0, 0, 0, false},
 		{"exec kernel", fhExec, kernelHeader, uint64p(0xffffffff81000198), 0xffffffff82000198, 0xffffffff83000198, 0, 0x1000000, false},
+		{"exec kernel perf offset0", fhExec, kernelHeader, uint64p(0xffffffff810002b8), 0xffffffff81000000, 0xffffffffa0000000, 0x0, 0x0, false},
```
aalexand: Similar to other comments, I'd maybe say "KASLR" instead of "perf" in this and next test case since I am not sure there is much specifics to Linux perf, I think it's more about the mechanics of kernel loading.

aalexand:
For the record, #299 also fixed some issues with processing KASLR'ed kernels. The MMAP in the perf.data there was
```
0 0x108 [0x50]: PERF_RECORD_MMAP -1/0: [0xffffffff89200000(0x37138000) @ 0xffffffff89200000]: x [kernel.kallsyms]_text
```
which I think the same what we are seeing here, not sure why #299 worked there but not here.

rauls5382:
The only way I can see #299 work is if loadSegment.Vaddr == loadSegment.Off
That is not the case on the files I'm seeing, though.

PR #373, opened 2018-05-04
internal/driver: skip tests requiring tcp on js

mvdan:
This commit upstreams the changes made in golang.org/cl/110096. As the
change says, it is acceptable for the new js/wasm target to only support
what nacl supports right now.


mvdan:
/cc @neelance @hyangah @aalexand 

This change is in preparation for updating the vendored pprof. Otherwise, this patch might have been reverted at some point in the future.

Also note that the original change disabled two tests, but I can see three "tcp available" skips. So I've done it for all three to be consistent.

mvdan:
My bad for forgetting to fetch the latest HEAD. Hopefully you won't leave a merge commit in the master commit history :)

mvdan:
Also - I'd like to wait for @neelance's OK on me skipping a third test. Perhaps the original patch was correct and mine isn't.

aalexand:
> Hopefully you won't leave a merge commit in the master commit history :)

@mvdan We squash on merging PRs.

> I'd like to wait for @neelance's OK on me skipping a third test.

@mvdan OK - waiting for @neelance's OK.

neelance:
Yes, also skipping the third test is correct.

Thanks for bringing these changes upstream, should have done so myself. I disabled the tests while working on the support for WebAssembly and forgot about them.

PR #380, opened 2018-05-21
Make readlineUI the default UI

hyangah:
The current implementation placed the dependency on
github.com/chzyer/readline in pprof.go instead of placing
it in the default ui just because the go repository that
vendors pprof may not want the dependency.

I evaluated another option (golang.org/x/crypto/ssh/terminal)
for the Go repository and there exist subtle differences
in the behavior. Inconsistency in UI is undesirable so golang
decided to vendor the chzyer readline as well for now.

Update golang/go#14041

### DIFF
--- a/internal/driver/options.go
+++ b/internal/driver/options.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/chzyer/readline"
 	"github.com/google/pprof/internal/binutils"
 	"github.com/google/pprof/internal/plugin"
 	"github.com/google/pprof/internal/symbolizer"
@@ -44,7 +45,7 @@ func setDefaults(o *plugin.Options) *plugin.Options {
 		d.Obj = &binutils.Binutils{}
 	}
 	if d.UI == nil {
-		d.UI = &stdUI{r: bufio.NewReader(os.Stdin)}
+		d.UI = defaultUI()
 	}
 	if d.Sym == nil {
 		d.Sym = &symbolizer.Symbolizer{Obj: d.Obj, UI: d.UI}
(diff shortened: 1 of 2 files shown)


internal/driver/options.go
```diff
@@ -107,6 +108,15 @@ func (goFlags) Parse(usage func()) []string {
 	return args
 }
 
+func defaultUI() plugin.UI {
+	rl, err := readline.New("")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "fall back to the default UI due to a failure in initializing readline: %v", err)
```
aalexand: "failed to init terminal-based UI, falling back to stdin/stdout: %v"

internal/driver/options.go
```diff
@@ -143,6 +153,68 @@ func (ui *stdUI) fprint(f *os.File, args []interface{}) {
 	f.WriteString(text)
 }
 
+// readlineUI implements the driver.UI interface using the
+// github.com/chzyer/readline library.
+// This is contained in pprof.go to avoid adding the readline
```
aalexand: The comment needs update.

rauls5382:
It is not clear to me that we want this dependence in the core pprof packages.
Was it not possible to add a similar dependence into golang's pprof.go?

I think it is useful to be able to easily build pprof without this dependence.
In the past we've seen reports of platforms where terminal I/O packages were not available.

hyangah:
Sure, it's possible to copy the code into golang's fork of pprof.go.
But, that's one more thing to remember in maintaining the code - if someone makes a change to add more features such as autocomplete, etc, I hope someone remembers and reflects the change to the half- vendored version of go in go repo. 

If it's undesirable to place this in the core pprof logic, is it possible to have this in a place so we can wire up the implementation used in this pprof tool to golang's pprof easily?

aalexand:
@hyangah This PR should be dropped / closed now?

hyangah:
Dropping for now.

PR #381, opened 2018-05-24
apply additional command overrides based on report format

nolanmar511:


internal/driver/driver.go
```diff
@@ -149,7 +151,7 @@ func generateReport(p *profile.Profile, cmd []string, vars variables, o *plugin.
 	return out.Close()
 }
 
-func applyCommandOverrides(cmd []string, v variables) variables {
+func applyCommandOverrides(cmd []string, outputFormat int, v variables) variables {
```
aalexand: How about changing the first argument to be just "cmd string"? It doesn't seem to be necessary to pass the array.

internal/driver/driver.go
```diff
@@ -176,6 +178,12 @@ func applyCommandOverrides(cmd []string, v variables) variables {
 			v.set("nodecount", "80")
 		}
 	}
+
+	if outputFormat == report.Proto {
```
aalexand: Why did you keep `case "proto", "raw":` in the switch, isn't it redundant now?

aalexand: Side question: why does "peek" case resets the filter but not not the tagfilter?

internal/driver/driver.go
```diff
@@ -176,6 +178,12 @@ func applyCommandOverrides(cmd []string, v variables) variables {
...
+		trim, tagfilter, filter = false, false, false
```
aalexand: Perhaps for future unless there is an easy way to do it: it would be good to have a way to generate reports in the proto format keeping the filter. E.g. using the internal -flame command with a -tagfocus filter to upload a subset of the profile.

internal/driver/driver.go
nolanmar511: It's now redundant. Done.

internal/driver/driver.go
nolanmar511: Done.

internal/driver/driver.go
nolanmar511: I'm assuming that this could be done just by not setting `filter` and `tagfilter` to false. Not sure if we would want to also, for example, add a comment to the profile to indicate filtering was done.

Will experiment, then look into changing.

internal/driver/driver.go
nolanmar511: It does seem odd. My best guess is that tagfilter was added later, and this is an oversight, but I'll look into the commit history to see if there's evidence for that.

nolanmar511:
PTAL

internal/driver/driver.go
aalexand: So we should fix it?

internal/driver/driver.go
aalexand: This can be done later.

internal/driver/driver.go
nolanmar511: Tag filtering was added after the peek command, (in [this](https://github.com/google/pprof/pull/33/files) PR), which suggests to me this is an oversight.

I've changed it in this PR. Let me know if you think it ought to be a separate PR.

aalexand:
@nolanmar511 Appveyor build failed?

nolanmar511:
Appveyor has an 503 error when installing graphviz with chocolatey. 

I've retried and am still getting this error. Not sure what's up, though I thinking waiting a bit could help (but that's maybe not the best solution).

aalexand:
@nolanmar511 It would be good to comment on the PR before merging about how the Appveyor failure was fixed - did you do something? did it just go away? something else?

PR #382, opened 2018-05-29
profile: fix legacy format Go heap profile parsing

hyangah:
This CL fixes a long-lasting bug that prevented pprof from recognizing
Legacy heap profile produced by Go. Go reports four types of samples
at once so the profile includes alloc_objects, alloc_space,
inuse_objects, and inuse_space. The bug caused pprof to misclassify Go
heap profile data and prevent selection of correct filtering/pruning
patterns in analysis.

Update golang/go#25096

Tested with the profile samples included in golang.org/issues/25096
(pprof hides the runtime functions with this change)

### DIFF
--- a/profile/legacy_profile.go
+++ b/profile/legacy_profile.go
@@ -1103,6 +1103,7 @@ var heapzSampleTypes = [][]string{
 	{"objects", "space"},
 	{"inuse_objects", "inuse_space"},
 	{"alloc_objects", "alloc_space"},
+	{"alloc_objects", "alloc_space", "inuse_objects", "inuse_space"}, // Go pprof legacy profiles
 }
 var contentionzSampleTypes = [][]string{
 	{"contentions", "delay"},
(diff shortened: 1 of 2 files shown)


nolanmar511:
Based on the [compatible](https://github.com/google/pprof/blob/master/profile/merge.go#L455-L473) function, it still won't be possible to merge legacy go heap profiles with go heap profiles. I think the error message from that function will still be clear, if someone tries to merge a legacy go heap profile with another heap profile. 

LGTM, but I'll let @aalexand approve.

hyangah:
@nolanmar511 I am not familiar with the issue - why do people want to merge Go heap profile with other (non-Go?) profile?

nolanmar511:
@hyangah -- My guess is that users (generally) don't want to merge a go heap profile with other non-go profiles.

But if a user is using a legacy profile, it seems possible to me that users could want to compare a legacy go profile with a non-legacy go profile. This wouldn't be possible right now, since the profiles would have a different number of sample types. It still won't be possible to merge a legacy and non-legacy go profile after this change, but the error message generated if a user tries this will still be clear (I'd been worried that the error message might become unclear).

My comment was a bit of an aside, and a bit confusing -- apologies.

aalexand:
The change LGTM so I'll merge it, but @rauls5382 FYI if you can take a look.

PR #383, opened 2018-05-31
Only colorize error messages if stderr is a terminal.

aalexand:


aalexand:
@rauls5382 Could you look at this change?

aalexand:
@rsc Thanks for pointing this out. The golang-embedded pprof uses x/crypto/ssh/terminal while standalone one uses chzyer/readline, their quirks may be different. Your enter-is-not-working issue is against standalone pprof? How do I set up a not-ANSI-enabled terminal like you have?

Well, experimenting more I do see at least strange behavior in emacs' shell. In the pprof interactive shell within emacs shell on each Enter the entered command prints out twice for some reason:

```
(pprof) hello
(pprof) hello
Unrecognized command: "hello"
```

rauls5382:
I agree we should disable readline for TERM="" and TERM="dumb" (any others?)
I think this commit is still useful, though.

PR #384, opened 2018-06-01
record diff base profile label key/value constant

nolanmar511:


profile/profile.go
```diff
@@ -704,6 +704,21 @@ func (s *Sample) HasLabel(key, value string) bool {
 	return false
 }
 
+// GetSourceAndDiffBaseSamples returns source samples for diff and diff base
+// profiles, separated into two slices.
+func (p *Profile) GetSourceAndDiffBaseSamples() ([]*Sample, []*Sample) {
```
aalexand: This looks overly specialized and too low-level in terms of its return types to me to be a part of API when I look at it. I would probably just expose a string constant like DiffBaseLabel for "pprof::base" and keep the rest of the code in the caller.

nolanmar511: Done.

nolanmar511:
PTAL

internal/driver/fetch_test.go
```diff
@@ -398,27 +398,27 @@ func TestFetchWithBase(t *testing.T) {
 				},
 				{
 					values: []int64{-2700, -608881724},
-					labels: map[string][]string{"pprof::base": {"true"}},
+					labels: map[string][]string{profile.DiffBaseLabelKey: {profile.DiffBaseLabelValue}},
```
aalexand: I probably wouldn't bother using the constants in the tests. It's fine to have some duplication.

Exposing both key and value constants looks noisy. Would it make sense to have function like Sample.DiffBaseSample instead?

nolanmar511: Done (for both)

aalexand:
@nolanmar511 Looks like tests are failing
```
=== RUN   TestDiffBaseSample/label_with_many_keys_and_values,_not_including_diff_base_label
--- FAIL: TestDiffBaseSample (0.00s)
    --- PASS: TestDiffBaseSample/empty_label_does_not_have_label (0.00s)
    --- FAIL: TestDiffBaseSample/label_with_one_key_and_value,_not_including_diff_base_label (0.00s)
    	profile_test.go:1031: sample.DiffBaseSample() got false, want true
    --- PASS: TestDiffBaseSample/label_with_one_key_and_value,_not_including_diff_base_label#01 (0.00s)
    --- FAIL: TestDiffBaseSample/label_with_many_keys_and_values,_including_diff_base_label (0.00s)
    	profile_test.go:1031: sample.DiffBaseSample() got false, want true
    --- PASS: TestDiffBaseSample/label_with_many_keys_and_values,_not_including_diff_base_label (0.00s)
```

nolanmar511:
Sorry -- tests fixed. PTAL.

internal/report/report_test.go
```diff
@@ -330,7 +330,7 @@ func TestComputeTotal(t *testing.T) {
 		{
 			Location: []*profile.Location{testL[2], testL[1], testL[0]},
 			Value:    []int64{-10, 3},
-			Label:    map[string][]string{"pprof::base": {"true"}},
+			Label:    map[string][]string{profile.DiffBaseLabelKey: {profile.DiffBaseLabelValue}},
```
aalexand: I would not make this change. https://github.com/google/pprof/pull/384#discussion_r192537247 applied to all tests.

nolanmar511: Sorry -- missed this test.
Should all be fixed now.

nolanmar511:
PTAL

profile/profile.go
```diff
@@ -30,6 +30,14 @@ import (
 	"time"
 )
 
+// DiffBaseLabelKey is the key for the label which is used to mark profile
+// samples which are a part of the diff base profile.
+const DiffBaseLabelKey = "pprof::base"
```
aalexand: With DiffBaseSample function exposed, does the constants really need to be public?

nolanmar511: These don't need to be exposed. Done.

PR #388, opened 2018-06-06
do not record function multiple times in topproto

nolanmar511:


internal/report/report.go
```diff
@@ -338,11 +340,15 @@ func printTopProto(w io.Writer, rpt *Report) error {
 
 type functionMap map[string]*profile.Function
 
-func (fm functionMap) FindOrAdd(ni graph.NodeInfo) *profile.Function {
+// FindOrAdd takes a node representing a function, adds the function represented
```
aalexand: Please fit comments in 80 chars.

internal/report/report_test.go
```diff
@@ -253,15 +253,22 @@ func TestFunctionMap(t *testing.T) {
...
 	}
 
-	want := []profile.Function{
-		{ID: 1, Name: "fun1"},
-		{ID: 2, Name: "fun2", Filename: "filename"},
-		{ID: 1, Name: "fun1"},
-		{ID: 3, Name: "fun2", Filename: "filename2"},
+	want := []struct {
+		function profile.Function
+		wasAdded bool
```
aalexand: wantAdded or wantWasAdded

internal/report/report_test.go
```diff
@@ -253,15 +253,22 @@ func TestFunctionMap(t *testing.T) {
...
+		{profile.Function{ID: 2, Name: "fun2", Filename: "filename"}, true},
+		{profile.Function{ID: 1, Name: "fun1"}, false},
+		{profile.Function{ID: 3, Name: "fun2", Filename: "filename2"}, true},
 	}
 
 	for i, tc := range nodes {
-		if got, want := fm.FindOrAdd(tc), want[i]; *got != want {
+		gotFunc, gotAdded := fm.FindOrAdd(tc)
+		if got, want := gotFunc, want[i].function; *got != want {
+			t.Errorf("%d: want %v, got %v", i, want, got)
```
aalexand: got should go first, please refresh on https://github.com/golang/go/wiki/CodeReviewComments

internal/report/report.go
```diff
@@ -309,7 +309,10 @@ func printTopProto(w io.Writer, rpt *Report) error {
 	}
 	functionMap := make(functionMap)
 	for i, n := range g.Nodes {
-		f := functionMap.FindOrAdd(n.Info)
+		f, added := functionMap.FindOrAdd(n.Info)
+		if added {
+			out.Function = append(out.Function, f)
+		}
 		flat, cum := n.FlatValue(), n.CumValue()
 		l := &profile.Location{
```
aalexand: Doesn't have to be in this PR, but I wonder if locations should be de-duped like functions?

internal/driver/driver_test.go
```diff
@@ -712,6 +715,73 @@ func cpuProfileSmall() *profile.Profile {
 	}
 }
 
+func cpuProfileNoFunctions() *profile.Profile {
```
aalexand: I may be missing something: the bug fix is about not properly handling multiple instances of the same function in a profile. Why do you add a profile with no functions here rather than extending some existing profile to trigger the issue? Why is it important to have a profile with no function information? It's good to have the number of test inputs to necessary minimum imo.

internal/report/report.go
```diff
@@ -338,11 +340,15 @@ func printTopProto(w io.Writer, rpt *Report) error {
 
 type functionMap map[string]*profile.Function
 
-func (fm functionMap) FindOrAdd(ni graph.NodeInfo) *profile.Function {
+// FindOrAdd takes a node representing a function, adds the function represented
+// by the node to the if the function is not already present, and returns the
+// function the node represents. This also returns a boolean, which is true if
+// the function was added and false otherwise.
+func (fm functionMap) FindOrAdd(ni graph.NodeInfo) (*profile.Function, bool) {
```
aalexand: Does this function have to be public?

internal/driver/driver_test.go
```diff
@@ -62,7 +62,8 @@ func TestParse(t *testing.T) {
 		{"text,lines,cum,hide=line[X3]0", "cpu"},
 		{"text,lines,cum,show=[12]00", "cpu"},
 		{"text,lines,cum,hide=line[X3]0,focus=[12]00", "cpu"},
-		{"topproto,lines,cum,hide=mangled[X3]0", "cpu"},
+		// {"topproto,lines,cum,hide=mangled[X3]0,", "cpu"},
```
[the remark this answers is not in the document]
nolanmar511: Done.

internal/report/report.go

internal/report/report.go
nolanmar511: No, done.

internal/report/report_test.go
nolanmar511: done

internal/report/report_test.go
nolanmar511: done

internal/report/report.go
nolanmar511: Would we expect the that locations to be de-duplicated when the graph is constructed?

internal/driver/driver_test.go
nolanmar511: I added a case with repeated functions.
The case with no functions is similar to the input on the issue.

nolanmar511:
PTAL

internal/driver/driver_test.go
```diff
@@ -63,6 +63,8 @@ func TestParse(t *testing.T) {
 		{"text,lines,cum,show=[12]00", "cpu"},
 		{"text,lines,cum,hide=line[X3]0,focus=[12]00", "cpu"},
 		{"topproto,lines,cum,hide=mangled[X3]0", "cpu"},
+		{"topproto,lines", "cpu_repeated_function"},
```
aalexand: Can we extend one of existing cases to have a repeated function? It doesn't seem the case with a function referenced by several locations needs to be special and having so many almost-the-same test cases becomes messy, let's try to simplify things.

internal/driver/driver_test.go
aalexand: I think the "no functions" case is redundant even if it's similar to the input in the bug report since the root cause is not handling case when a function referenced by several locations. I'd remove it unless you think it has potential to expose some other bugs.

internal/report/report.go
```diff
@@ -309,7 +309,10 @@ func printTopProto(w io.Writer, rpt *Report) error {
 	}
 	functionMap := make(functionMap)
```
aalexand: The coverage data says this code block is not covered by tests. Is it a bug with the coverage data or there is something in it?

internal/report/report.go
nolanmar511: I've checked with a number of large profiles and have not found one where locations a repeated (though I've not constructed any profile specifically for this purpose).

I've also confirmed that there is one node in the graph per location in the profile ( code confirming this [here](https://github.com/google/pprof/blob/master/internal/graph/graph.go#L520)).

So duplication of locations shouldn't be possible, (and if we see duplicated locations in the -topproto output, then the locations would be duplicated in pprof's internal proto representation, so the problem should be fixed elsewhere.)

internal/driver/driver_test.go
nolanmar511: Done. 
Did more testing on master, and current cpuProfile() has repeated functions, so can be used for this.

internal/driver/driver_test.go
nolanmar511: Done.
Did more testing on master, and current cpuProfile() has repeated functions, so can be used for this.

internal/report/report.go
nolanmar511: I'm incline to think there is a big with the coverage data.
Several of functions for printing the output format aren't covered, and I've added are tests that go from passing to failing when this function is modified.

nolanmar511:
PTAL

internal/driver/driver_test.go
```diff
@@ -63,6 +63,7 @@ func TestParse(t *testing.T) {
 		{"text,lines,cum,show=[12]00", "cpu"},
 		{"text,lines,cum,hide=line[X3]0,focus=[12]00", "cpu"},
 		{"topproto,lines,cum,hide=mangled[X3]0", "cpu"},
+		{"topproto,lines", "cpu"},
```
aalexand: Could you confirm this test fails without your fix? I am curious why the test one line above did not reproduce the issue.

nolanmar511: I've confirmed this.
I think the line above doesn't reproduce because the hide filter hides all the repeated functions.

aalexand: Interesting that topproto format applied the filters, unlike the proto one. Relevant to the other discussion where we decided to stop discarding the filters for the proto format.

PR #390, opened 2018-06-20
document diff_base flag (#384)

nolanmar511:


internal/driver/cli.go
```diff
@@ -312,7 +312,8 @@ var usageMsgSrc = "\n\n" +
 	"    -buildid              Override build id for main binary\n" +
 	"    -add_comment          Free-form annotation to add to the profile\n" +
 	"                          Displayed on some reports or with pprof -comments\n" +
-	"    -base source          Source of profile to use as baseline\n" +
+	"    -diff_base source     Source for base profile for comparison\n" +
```
aalexand: Nit here and line below: maybe leave the the "Source of" wording rather than "Source for"? That reads somewhat better to me.

nolanmar511: Done. Also switched to "of" in parseFlags() function.

internal/driver/cli.go
```diff
@@ -45,8 +45,8 @@ type source struct {
 func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flag := o.Flagset
 	// Comparisons.
-	flagBase := flag.StringList("base", "", "Source for base profile for profile subtraction")
-	flagDiffBase := flag.StringList("diff_base", "", "Source for diff base profile for comparison")
+	flagBase := flag.StringList("base", "", "Source of base profile for profile subtraction")
```
aalexand: Nit: you list base flag here first and diff_base second, but in the help message it's reversed. List diff_base flag here first to emphasize that's the flag most people should use?

nolanmar511: Done.

nolanmar511:
PTAL

PR #391, opened 2018-06-22
log error message when negative samples appear in non-diff_base profiles

### LINKED ISSUE #323: Fix percentage calculation in diff mode (-base) and remove -positive_percentages

nolanmar511:
This is the last change needed for improving diff mode, so fixes #323.


internal/report/report.go
```diff
@@ -1237,9 +1239,12 @@ func computeTotal(prof *profile.Profile, value, meanDiv func(v []int64) int64) i
 			diffDiv += d
 		}
 	}
+
 	if diffTotal > 0 {
 		total = diffTotal
 		div = diffDiv
+	} else if negSamples {
+		ui.PrintErr("Negative sample values appeared in the profile.\nIf using the -base flag to compare profiles, consider using the -diff_base flag instead.")
```
aalexand: It's a bit strange that printing out this user interface messages goes down as deeply as to this code that computes the total. As well as that report.New has to take the UI now. I think it would be better to make computeTotal return a boolean (in addition to the int64 like now) indicating whether negative values were seen, store that in the report object, and print the message in Generate() making that function take UI.

nolanmar511:
PTAL

internal/report/report_test.go
```diff
@@ -213,6 +238,41 @@ var testProfile = &profile.Profile{
 	Mapping:  testM,
 }
 
+var testProfile2 = &profile.Profile{
```
aalexand: I think it would be easier to massage testProfile where needed in the test to produce this profile. As far as I can tell, it's just negating the value in one of the samples.

internal/report/report_test.go
```diff
@@ -406,9 +468,13 @@ func TestComputeTotal(t *testing.T) {
 
 	for _, tc := range testcases {
 		t.Run(tc.desc, func(t *testing.T) {
-			if gotTotal := computeTotal(tc.prof, tc.value, tc.meanDiv); gotTotal != tc.wantTotal {
+			gotTotal, gotUnexpNegSamples := computeTotal(tc.prof, tc.value, tc.meanDiv)
+			if gotTotal != tc.wantTotal {
 				t.Errorf("got total %d, want %v", gotTotal, tc.wantTotal)
 			}
+			if gotUnexpNegSamples != tc.wantUnexpNegSamples {
+				t.Errorf("got  unexpected negative samples %v, want %v", gotUnexpNegSamples, tc.wantUnexpNegSamples)
```
aalexand: Nit: double space after "got"?

internal/report/report_test.go
```diff
@@ -42,39 +43,63 @@ func TestSource(t *testing.T) {
...
 			want: path + "source.dot",
 		},
+		{
+			rpt: New(
+				testProfile2.Copy(),
+				&Options{
+					OutputFormat: Dot,
+					CallTree:     true,
+					Symbol:       regexp.MustCompile(`.`),
+					TrimPath:     "/some/path",
```
aalexand: What do you specify TrimPath in this test for?

internal/report/report_test.go
```diff
@@ -42,39 +43,63 @@ func TestSource(t *testing.T) {
...
+					SampleUnit:  testProfile1.SampleType[1].Unit,
 				},
 			),
 			want: path + "source.dot",
 		},
+		{
+			rpt: New(
+				testProfile2.Copy(),
+				&Options{
+					OutputFormat: Dot,
```
aalexand: For this test, is Dot format really necessary? Which kind of issues would it help to catch?

internal/report/report.go
```diff
@@ -1202,8 +1206,8 @@ func New(prof *profile.Profile, o *Options) *Report {
 		}
 		return measurement.ScaledLabel(v, o.SampleUnit, o.OutputUnit)
 	}
-	return &Report{prof, computeTotal(prof, o.SampleValue, o.SampleMeanDivisor),
-		o, format}
+	total, unexpNegSamples := computeTotal(prof, o.SampleValue, o.SampleMeanDivisor)
```
aalexand: unexpNegSamples -> hasNegativeSamples, here and in other places.

internal/report/report.go
```diff
@@ -82,7 +82,11 @@ type Options struct {
 }
 
 // Generate generates a report as directed by the Report.
-func Generate(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
+func Generate(w io.Writer, rpt *Report, obj plugin.ObjTool, ui plugin.UI) error {
+	if rpt.unexpNegSamples {
+		ui.PrintErr("Negative sample values appeared in the profile.\nIf using the -base flag to compare profiles, consider using the -diff_base flag instead.")
```
aalexand: It would be good to shorten the message a bit (without any loss of information), but I couldn't come up with a good version...

internal/report/report.go
```diff
@@ -1246,20 +1254,23 @@ func computeTotal(prof *profile.Profile, value, meanDiv func(v []int64) int64) i
...
 type Report struct {
-	prof        *profile.Profile
-	total       int64
-	options     *Options
-	formatValue func(int64) string
+	prof            *profile.Profile
+	total           int64
+	options         *Options
+	formatValue     func(int64) string
+	unexpNegSamples bool
```
aalexand: Maybe move it up to be after "total" since these are related.

internal/report/report_test.go
nolanmar511: Done.
I negated all samples for simplicity, but this works.

internal/report/report_test.go
nolanmar511: Fixed.

internal/report/report_test.go
nolanmar511: Switched to list

internal/report/report.go
Not sure if this could be confusing, because -diff_base profiles will have negative samples, but this will be false.

internal/report/report_test.go
nolanmar511: Removed this.

internal/report/report.go
nolanmar511: I'm thinking about it still... not sure if I'll come up with something.

internal/report/report.go
aalexand: I see. Well, maybe leave the old name then. Up to you.

internal/report/report.go

internal/report/report.go
nolanmar511: Switching back to old name.

nolanmar511:
PTAL

internal/report/report.go
```diff
@@ -82,7 +82,11 @@ type Options struct {
 }
 
 // Generate generates a report as directed by the Report.
-func Generate(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
+func Generate(w io.Writer, rpt *Report, obj plugin.ObjTool, ui plugin.UI) error {
+	if rpt.unexpNegSamples {
+		ui.PrintErr("Profile has negative sample values.\nIf using the -base flag to compare profiles, consider using the -diff_base flag instead.")
```
aalexand: Could we check if `-base` flag was actually used and print this warning (or this part of the warning) only in that case?

nolanmar511: It's possible, but I don't think it's trivial. 

I see a few possible options:
1. Call `report.Generate()` with a boolean indicating if `-base` was used to specify a base profile. This requires also passing that boolean into `interactive()` and `generateReport()` or extending `plugin.Options` with a field for specifying if there is a `-base` profile.
2. Add a boolean to profile.Profile to track if a base profile was specified with the `base`.
3. Start tracking -base somehow in `pprofVariables` (this seems undesirable, because I think specifying a base profile from the interactive mode would fail silently).

Any other ideas or preferences?

(There's also the smaller complication of what to name a boolean indicating the a profile has a "base: profile but not a "diff base" profile)

nolanmar511:
PTAL at response to comment.

internal/report/report.go
aalexand: I wonder if it would make more sense to print out this warning in fetchProfiles around where pbase and s.DiffBase are checked. The complication is that at that point the sample index is not known, but we could print a warning if any of the metrics has a negative sample and mention all such metrics in the message. The advantage of placing it there is that the message would be printed only once as opposed to for every report in the interactive mode. What do you think?

nolanmar511: I'm happy with that solution. 
It may lead to the message printing out when people are using -base to look at cumulative profiles, but I think that'll be okay.

aalexand: That's a good point, we don't want the message to be printed out when someone is doing `pprof -sample_index=alloc_space profile1.pb.gz -base profile2.pb.gz`.

aalexand: Maybe shorten the message a bit to have it one line: "Profile has negative values. If using -base, consider using -diff_base instead."

internal/report/report.go
```diff
@@ -1246,20 +1254,23 @@ func computeTotal(prof *profile.Profile, value, meanDiv func(v []int64) int64) i
 	if diffTotal > 0 {
 		total = diffTotal
 		div = diffDiv
+		// negative samples are expected in diff base profiles.
```
aalexand: Nit: "Negative" (capitalize documentation sentences)

internal/report/report.go
nolanmar511: done

internal/report/report.go

nolanmar511:
PTAL

internal/report/report.go
```diff
@@ -82,7 +82,11 @@ type Options struct {
 }
 
 // Generate generates a report as directed by the Report.
-func Generate(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
+func Generate(w io.Writer, rpt *Report, obj plugin.ObjTool, ui plugin.UI) error {
+	if rpt.unexpNegSamples {
+		ui.PrintErr("Profile has negative values. If using -base, consider using -diff_base instead.")
```
aalexand: Sorry, let's do one more tweak to clarify _why_ someone would need to consider using -diff_base instead. How about

`Profile has negative values, percentage values may be incorrect. Profile has negative values. If using -base, consider using -diff_base instead.`

aalexand: Edited the comment, I had a mistake there. The right version is `Profile has negative values, percentage values may be incorrect. If using -base, consider using -diff_base instead.`.

nolanmar511:
PTAL

aalexand:
Actually closing the PR. Also to clarify, the exact reason for why the change shouldn't go in is https://github.com/golang/go/issues/26638.

Issue #389, opened 2018-06-14
browser list used for invokeVisualizer needs be rethought


hyangah:
pprof invokes web browsers when the output of the command (e.g. web or weblist) 
needs to be visualized in a web browser. 

golang/go#25853 exhibited a couple of issues around the browser invocation logic.

1) the list of browsers populated in [browsers](https://github.com/google/pprof/blob/8b03ce837f3401233d4e32de3a5f0ea3ff1c13c2/internal/driver/commands.go#L339 ) needs to be rethought.

If there is BROWSER env var set, I think it should be picked as the first choice. Currently, not.
FYI, go's [cmd/internal/browser](https://github.com/golang/go/blob/46076c37578fba9b49059584ef896099c9240fb2/src/cmd/internal/browser/browser.go#L18) package has an implementation for the similar purpose. 

Also, note that commands like chrome or google-chrome are not used on Mac. If they exist, they are likely not the browser command we expect. 

2) [invokeVisualize](https://github.com/google/pprof/blob/8b03ce837f3401233d4e32de3a5f0ea3ff1c13c2/internal/driver/commands.go#L422) returns once the command binary is located and can launch a process. 

The command's stderr is directed to the pprof's stderr so users could be able to see the error message but *only if* the command outputs the error message to stderr.  I don't know what's the best way to deal with it.  Will it be useful if the tool prints out the exact command? Or, wait for a second, and if the process seems to exit with error code already, try again?

marwan-at-work:
Thanks for opening this issue!

I think it might be helpful if the `web` and `weblist` commands communicated a little more to the user about what they did. So that if there's some weird behavior, the user can probably figure it out quickly.

I was unfamiliar with pprof and how the `web` command worked internally. And so typing `web` and seeing it break, was intimidating and confusing until 4-5 hours later. 

So maybe when you type `web` the following will be logged: 

```
attempting to find chrome binary
chrome binary not found.
attempting to open Firefox through firefox binary
```

etc etc

When `chrome` actually opened for me, but it didn't work, I thought there was some fundamental problem, and not that I just had the wrong binary. I was able to tell because the `which chrome` command gave me a suspicious path. 

Maybe more robustly, if an OS had a consistent way of opening a user's default browser, then that would probably be best.

nolanmar511:
First part (changing the order of browsers) is fixed with #392.

nolanmar511:
[...]

PR #392, opened 2018-06-26
update order of available browsers

nolanmar511:
Fixes first issue mentioned in #389
Order of commands matches go's [cmd/internal/browser Commands()](https://github.com/golang/go/blob/46076c37578fba9b49059584ef896099c9240fb2/src/cmd/internal/browser/browser.go#L18) function, except that "sensible-browser" is added when no on windows or darwin, and "xdg-open" is added even when "DISPLAY" env variable is not set.

### DIFF
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -337,21 +337,23 @@ func listHelp(c string, redirect bool) string {
 
 // browsers returns a list of commands to attempt for web visualization.
 func browsers() []string {
-	cmds := []string{"chrome", "google-chrome", "firefox"}
+	var cmds []string
+	if userBrowser := os.Getenv("BROWSER"); userBrowser != "" {
+		cmds = append(cmds, userBrowser)
+	}
 	switch runtime.GOOS {
 	case "darwin":
-		return append(cmds, "/usr/bin/open")
+		cmds = append(cmds, "/usr/bin/open")
 	case "windows":
-		return append(cmds, "cmd /c start")
+		cmds = append(cmds, "cmd /c start")
 	default:
-		userBrowser := os.Getenv("BROWSER")
-		if userBrowser != "" {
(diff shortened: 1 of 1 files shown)


internal/driver/commands.go
```diff
@@ -337,21 +337,19 @@ func listHelp(c string, redirect bool) string {
 
 // browsers returns a list of commands to attempt for web visualization.
 func browsers() []string {
```
aalexand: Could we match the behavior of Go's cmd/internal/browser/Commands function exactly? Only adding xdg-open when DISPLAY is set is I think reasonable. Not sure why Go's code doesn't use sensible-browser command which is Debian's way to open a default browser. In reality using just xdg-open should be enough, maybe that's why. So how about doing the same, dropping sensible-browser from the list? @hyangah FYI.

hyangah: I don't know why sensible-browser is missing. I guess it's just because that's new in pprof code as well and no one in go side bothered to add it - according to the initial commit of cmd/internal/browser 
https://go-review.googlesource.com/c/go/+/29160/2/src/cmd/internal/pprof/commands/commands.go

We can update cmd/internal/browser to match behaviors.

aalexand: @nolanmar511 Let's probably leave sensible-browser in, after xdg-open in the list. Only xdg-open should be guarded by DISPLAY presence perhaps (since sensible-browser can probably resolve to a terminal browser like elinks). Though frankly I am not sure opening any kind of browser in a non-GUI environment is that useful. I'd probably expect that if DISPLAY is not set, the tool would just print out the URL the user should open by copying and pasting to their browser (think an SSH terminal).

aalexand: Thanks. I assume the code now is line-wise equivalent to the cmd/internal/browser one except `cmds = append(cmds, "sensible-browser")` line. Please merge the PR once CI is green.

PR #393, opened 2018-06-26
apply filters to -proto output format

nolanmar511:
We've had some discussion, and concluded that user-specified filters should be applied to the -proto output format.

This change makes it so that filters are applied to the -proto output.

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -152,7 +152,7 @@ func generateReport(p *profile.Profile, cmd []string, vars variables, o *plugin.
 }
 
 func applyCommandOverrides(cmd string, outputFormat int, v variables) variables {
-	trim, tagfilter, filter := v["trim"].boolValue(), true, true
+	trim := v["trim"].boolValue()
 
 	switch cmd {
 	case "callgrind", "kcachegrind":
@@ -162,7 +162,7 @@ func applyCommandOverrides(cmd string, outputFormat int, v variables) variables 
 		trim = false
 		v.set("addressnoinlines", "t")
 	case "peek":
-		trim, tagfilter, filter = false, false, false
+		trim = false
 	case "list":
 		v.set("nodecount", "0")
 		v.set("lines", "t")
@@ -185,17 +185,6 @@ func applyCommandOverrides(cmd string, outputFormat int, v variables) variables 
 		v.set("nodecount", "0")
 		v.set("nodefraction", "0")
 		v.set("edgefraction", "0")
-	}
-	if !tagfilter {
-		v.set("tagfocus", "")
-		v.set("tagignore", "")
-	}
-	if !filter {
-		v.set("focus", "")
-		v.set("ignore", "")
-		v.set("hide", "")
-		v.set("show", "")
-		v.set("show_from", "")
 	}
 	return v
 }
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -177,7 +177,7 @@ func applyCommandOverrides(cmd string, outputFormat int, v variables) variables 
 	}
 
 	if outputFormat == report.Proto || outputFormat == report.Raw {
-		trim, tagfilter, filter = false, false, false
+		trim = false
 		v.set("addresses", "t")
 	}
 
--- a/internal/driver/driver_test.go
+++ b/internal/driver/driver_test.go
@@ -116,21 +116,21 @@ func TestParse(t *testing.T) {
 			}
 			defer os.Remove(protoTempFile.Name())
 			defer protoTempFile.Close()
-			f.bools["lines"] = true
 			f.strings["output"] = protoTempFile.Name()
 
 			if flags[0] == "topproto" {
 				f.bools["proto"] = false
 				f.bools["topproto"] = true
+				f.bools["addresses"] = true
 			}
 
 			// First pprof invocation to save the profile into a profile.proto.
-			o1 := setDefaults(nil)
-			o1.Flagset = f
-			o1.Fetch = testFetcher{}
-			o1.Sym = testSymbolizer{}
-			o1.UI = testUI
-			if err := PProf(o1); err != nil {
+			o := setDefaults(nil)
+			o.Flagset = f
+			o.Fetch = testFetcher{}
+			o.Sym = testSymbolizer{}
+			o.UI = testUI
+			if err := PProf(o); err != nil {
 				t.Fatalf("%s %q:  %v", tc.source, tc.flags, err)
 			}
 			// Reset the pprof variables after the proto invocation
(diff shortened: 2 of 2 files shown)


internal/driver/driver.go
```diff
@@ -177,7 +177,7 @@ func applyCommandOverrides(cmd string, outputFormat int, v variables) variables
 	}
 
 	if outputFormat == report.Proto || outputFormat == report.Raw {
-		trim, tagfilter, filter = false, false, false
+		trim = false
```
aalexand: Should we also stop resetting the filters for the peek command? We would be able to remove the whole filter / tagfilter machinery form the applyCommandOverrides function then. I can't think of a good reason to keep resetting the filters for the peek command.

internal/driver/driver_test.go
```diff
@@ -125,6 +125,17 @@ func TestParse(t *testing.T) {
 				f.bools["topproto"] = true
 			}
 
+			// Filters should not be applied on first pprof invocation.
```
aalexand: Why? Is this a hack to make the old tests work or is it truly to right thing to do?

internal/driver/driver.go
aalexand: Might be good to add a test for the raw command to driver_test.go with a filter making sure it fails before and passes now. A dedicated test for proto command would be good too, though not sure how elegant adding that would be given it's a binary format.

internal/driver/driver_test.go
```diff
@@ -151,13 +162,15 @@ func TestParse(t *testing.T) {
...
+				addFlags(&f, flags)
 				solution = solutionFilename(tc.source, &f)
 				delete(f.bools, "topproto")
 				f.bools["text"] = true
 			} else {
 				delete(f.bools, "proto")
-				addFlags(&f, flags[:1])
+				addFlags(&f, flags)
 				solution = solutionFilename(tc.source, &f)
+				fmt.Printf(solution)
```
aalexand: Is this a debugging leftover?

internal/driver/driver.go
rauls5382: +1 to resetting it for peek as well.

The original intent was for peek to provide full details no matter what the settings were (to inspect at a very low level), but I think it is just confusing, and the user can always reset their filters if they so desire.

aalexand:
@nolanmar511 Note that the Travis CI is failing.

internal/driver/driver_test.go
nolanmar511: Fixed.

internal/driver/driver.go
nolanmar511: Done.

internal/driver/driver_test.go
nolanmar511: I feel like the right thing may be to re-write this test so that we can specify flags for the first invocation of pprof (which would have to have -proto or -topproto  as format so a protobuf is written out), and then separate flags to 

This would make the test work for testing proto and topproto, without changing the test too much or duplicating a lot of this code. 

So, most current test cases would look like: 
```
{
  firstInvkFlags: "proto", 
  secondInvkFlags: "text,functions,flat", 
  profile: "cpu"
},
```

Or, to simplify writing test cases, we can have firstInvkFlags be set to "proto" or "proto,lines" when that value is an empty string.

What do you think of this?

internal/driver/driver.go
nolanmar511: I replied to comment above with a way to modify TestParse, which would work nicely for testing that the proto command works as intended. We can confirm filters applied to proto in first invocation of have stuck in the second.

Want to check in before making changes, since I think making these changes will have a reasonable impact on TestParse and likely the names of the testdata files used.

internal/driver/driver_test.go
aalexand: That seems overkill to me. I feel the first invocation serves only one purpose: check that the proto generation works as expected under different circumstances. Could we just run the first invocation without any flags?

Alternatively, it might be even cleaner if we'd have just one invocation, not two, and a separate test case for "proto" format, with whatever flags we want to test. The only wrinkle there is how to compare the output since proto is a binary file.

nolanmar511:
PTAL

internal/driver/driver_test.go
```diff
@@ -104,31 +104,29 @@ func TestParse(t *testing.T) {
 
 			testUI := &proftest.TestUI{T: t, AllowRx: "Generating report in|Ignoring local file|expression matched no samples|Interpreted .* as range, not regexp"}
 
-			f := baseFlags()
-			f.args = []string{tc.source}
+			f1 := baseFlags()
```
aalexand: Nit: could you just reassign the `f` variable below rather than renaming? It will make the diff a bit cleaner.

nolanmar511: Done.

nolanmar511:
PTAL

internal/driver/driver_test.go
```diff
@@ -109,16 +109,14 @@ func TestParse(t *testing.T) {
...
-			addFlags(&f, flags[1:])
-
 			// Encode profile into a protobuf and decode it again.
 			protoTempFile, err := ioutil.TempFile("", "profile_proto")
 			if err != nil {
 				t.Errorf("cannot create tempfile: %v", err)
 			}
 			defer os.Remove(protoTempFile.Name())
 			defer protoTempFile.Close()
+			f.bools["lines"] = true
```
aalexand: Is this needed? For the proto output format the default is addresses anyway.

nolanmar511:
Travis seems to be failing with `go get` failures.
Seems unrelated to this change. Going to wait for a bit and try to re-run the travis tests.

internal/driver/driver_test.go
```diff
@@ -124,15 +121,16 @@ func TestParse(t *testing.T) {
...
 			}
 
 			// First pprof invocation to save the profile into a profile.proto.
-			o1 := setDefaults(nil)
-			o1.Flagset = f
-			o1.Fetch = testFetcher{}
-			o1.Sym = testSymbolizer{}
-			o1.UI = testUI
-			if err := PProf(o1); err != nil {
+			o := setDefaults(nil)
```
aalexand: Could you keep the original variable names? Sorry, I probably wasn't clear previously when I asked to use re-assignment for the `f` variable - I don't really care much one way or another, I am just suggesting we limit unnecessary styling changes so that the diff is more telling.

nolanmar511: Done.

nolanmar511:
PTAL

aalexand:
@nolanmar511 Travis is failing?

nolanmar511:
Travis failed during go get for some builds:
```
10.67s$ go get -u github.com/golang/lint/golint honnef.co/go/tools/cmd/...
# honnef.co/go/tools/lint/lintutil
../../../honnef.co/go/tools/lint/lintutil/util.go:273: undefined: types.SizesFor
```

The Travis failure seems unrelated to this change -- I'll wait for a bit, re-run the tests, and merge when they pass.

PR #411, opened 2018-08-17
Add support for fat Mach-O files

Timmmm:
Fixes #1033. Note that this does not do the "correct" thing, of checking the main executable to see what architecture it is and trying to match that. Or maybe checking the architecture in the profile file if it is present.

Instead it just uses the host architecture to decide which subfile of the fat archive to open.

Also I haven't added tests, sorry. Might have time later but not at the moment. It doesn't print any error messages though so that seems good.


aalexand:
Sorry for the delay, I'll take a look later today.

internal/binutils/binutils.go
```diff
@@ -223,19 +224,17 @@ func (bu *Binutils) Open(name string, start, limit, offset uint64) (plugin.ObjFi
 		return f, nil
 	}
 	if machoMagicLittle == macho.MagicFat || machoMagicBig == macho.MagicFat {
-		// TODO: #1033
-		return nil, fmt.Errorf("fat Mach-O archives are currently unsupported")
+		f, err := b.openFatMachO(name, start, limit, offset)
+		if err != nil {
+			return nil, fmt.Errorf("error reading fat Mach-O file %s: %s", name, err)
```
aalexand: Using "%v" for the `err` argument is more common.

internal/binutils/binutils.go
```diff
@@ -258,6 +257,51 @@ func (b *binrep) openMachO(name string, start, limit, offset uint64) (plugin.Obj
...
+		return nil, fmt.Errorf("error parsing %s: %v", name, err)
+	}
+	defer of.Close()
+
+	if len(of.Arches) == 0 {
+		return nil, fmt.Errorf("empty fat Mach-O file: %s", name)
+	}
+
+	var arch macho.Cpu
+	// Use the host architecture. Not ideal.
```
aalexand: Maybe expand on why not ideal as
```
// Use the host architecture.
// TODO: Match the arch where the profile was recorded instead.
```

PR #412, opened 2018-08-18
Expose sample_index argument through web ui

nvanbenschoten:
The option is exposed in the web UI's query string under a
new `?si=<sample_index>` parameter. This is particularly
useful for heap profiles, where switching between the four
indices (alloc_objects, alloc_space, inuse_objects, and inuse_space)
is a common task.

We could add a menu to the UI to make discovery of these
options easier, but for now this seems sufficient.

### DIFF
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -201,6 +201,7 @@ func openBrowser(url string, o *plugin.Options) {
 		{"s", "show"},
 		{"i", "ignore"},
 		{"h", "hide"},
+		{"si", "sample_index"},
 	} {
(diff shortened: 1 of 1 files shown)


aalexand:
You could as well just pass `-sample_index` flag to the pprof invocation when starting the web server. Is that not sufficient because you want to switch between different sample indices without having to restart the server?

nvanbenschoten:
> Is that not sufficient because you want to switch between different sample indices without having to restart the server?

Yes, exactly. Restarting the entire pprof server just to change the default `-sample_index` flag seems like the wrong way to approach this. That's especially true when the pprof web server is embedded inside of a larger project.

PR #414, opened 2018-08-24
Add sample index selector to web ui

nvanbenschoten:
This is a follow-up change to #412 which adds a selector to the web
ui for selecting the sample index of a profile to view. The selector
is hidden for profiles with only a single sample type.

screen shot 2018-08-23 at 10 41 13 pm

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -249,6 +249,21 @@ table tr td {
     </div>
   </div>
 
+  {{$sampleLen := len .SampleTypes}}
+  {{if gt $sampleLen 1}}
+  <div id="sample" class="menu-item">
+    <div class="menu-name">
+      Sample
+      <i class="downArrow"></i>
+    </div>
+    <div class="submenu">
+      {{range .SampleTypes}}
+      <a href="?si={{.}}" id="{{.}}">{{.}}</a>
+      {{end}}
+    </div>
+  </div>
+  {{end}}
+
   <div id="refine" class="menu-item">
     <div class="menu-name">
       Refine
@@ -718,9 +733,18 @@ function viewer(baseUrl, nodes) {
     return str.replace(/([\\\.?+*\[\](){}|^$])/g, '\\$1');
   }
 
+  function setSampleIndexLink(id) {
+    const elem = document.getElementById(id);
+    if (elem != null) {
+      setHrefParams(elem, function (params) {
+        params.set("si", id);
+      });
+    }
+  }
+
   // Update id's href to reflect current selection whenever it is
   // liable to be followed.
-  function makeLinkDynamic(id) {
+  function makeSearchLinkDynamic(id) {
     const elem = document.getElementById(id);
     if (elem == null) return;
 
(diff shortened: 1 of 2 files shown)


internal/driver/webui.go
```diff
@@ -264,6 +265,7 @@ func (ui *webInterface) render(w http.ResponseWriter, tmpl string,
 	data.Title = file + " " + profile
 	data.Errors = errList
 	data.Total = rpt.Total()
+	data.SampleTypes = sampleTypes(rpt.Profile())
```
aalexand: Can you just use ui.prof?

internal/report/report.go
```diff
@@ -1262,6 +1262,9 @@ type Report struct {
 	formatValue func(int64) string
 }
 
+// Profile returns the profile used to create a report.
+func (rpt *Report) Profile() *profile.Profile { return rpt.prof }
```
aalexand: See my comment in webui.go - I don't think this accessor is needed.

internal/driver/webhtml.go
```diff
@@ -736,35 +758,37 @@ function viewer(baseUrl, nodes) {
     elem.addEventListener('touchstart', updater);
 
     function updater() {
-      elem.href = updateUrl(new URL(elem.href), param);
+      // The selection can be in one of two modes: regexp-based or
+      // list-based.  Construct regular expression depending on mode.
+      let re = regexpActive
```
aalexand: Since you are changing the code around handling the filters, could you describe / confirm which manual testing is done? There are no automated tests for the web UI feature unfortunately.

internal/driver/webui.go
nvanbenschoten: Yes and I actually had this written that way at first. I switched to this approach because `ui.prof` was used to generate the `rpt`, which is the input to this method, so it felt appropriate to look at the profile from `rpt`. This also avoids leaking the abstraction that `ui.prof == rpt.prof`.

Happy to switch it back if you'd like, I don't have strong feelings.

internal/driver/webhtml.go
nvanbenschoten: Yeah, I played with this for a little while and everything seemed correct (after my most recent cleanup). Here's an incomplete list of things I tested:
- All "Refine" options still work with empty and non-empty regexp without sample index
- All "Refine" options still work with empty and non-empty regexp with sample index
- "Reset" button works
- Sample index selection does not remove "Refine" options
- "Refine" option does not remove sample index selection
- No errors in the web console
- Sample button not show when profile only has one sample

internal/driver/webui.go
aalexand: Adding an accessor for the profile in the report is more leaking since it changes the report public API surface, while here we are just dealing with local context of a single file implementation, so yes, please change. There isn't much motivation for passing the report object here, BTW, so we could as well just pass the total which is the only thing the render function uses.

nvanbenschoten: Done.

internal/report/report.go
nvanbenschoten: Done.

nvanbenschoten:
Looks like CI flaked with the error:

```
The command "go get -u github.com/golang/lint/golint honnef.co/go/tools/cmd/..." failed and exited with 2 during .
```

PR #419, opened 2018-09-17
Change order of commands returned by browser to prioritize commands used to open browsers on windows and mac

### LINKED ISSUE #402: pprof -web opens an svg image on my linux machine
3e1e08e68eaed0faefa67fe1c1e10e004cec2c9a (latest commit at the time of posting this)

If you run pprof from GitHub, what's the Git revision?

Linux (ubuntu)

### What did you do?
I called `pprof -web /path/to/profile.pb.gz`

I expected the graph to open in chrome.

### What did you see instead?
An svg image opened.

I think this is related to https://github.com/google/pprof/pull/392

nolanmar511:
Order selected based on discussion from #418.

Still fixes #402.


internal/driver/commands.go
```diff
@@ -336,27 +336,28 @@ func listHelp(c string, redirect bool) string {
...
+		// browser).
+		cmds = append(cmds, "sensible-browser")
 		if os.Getenv("DISPLAY") != "" {
+			// xdg-open is only for use in a desktop environment.
 			cmds = append(cmds, "xdg-open")
 		}
-		cmds = append(cmds, "sensible-browser")
 	}
-	return cmds
+	return append(cmds, []string{"chrome", "google-chrome", "chromium", "firefox"}...)
```
aalexand: I feel it's a bit weird that "sensible-browser" priority is different from these explicit browser commands. I think the point behind "sensible-browsers" is that sometimes this command may be missing (I think it's Debian / Ubuntu only?) and so we fall back to "chrome" etc. explicitly. Having anything in between "sensible-browser" and "chrome" and friends doesn't look right.

nolanmar511: Modified to only use chrome (and friends) on linux.

internal/driver/commands.go
```diff
@@ -336,25 +336,22 @@ func listHelp(c string, redirect bool) string {
...
 		cmds = append(cmds, userBrowser)
 	}
-	cmds = append(cmds, []string{"chrome", "google-chrome", "chromium", "firefox"}...)
 	switch runtime.GOOS {
 	case "darwin":
 		cmds = append(cmds, "/usr/bin/open")
 	case "windows":
 		cmds = append(cmds, "cmd /c start")
 	default:
+		cmds = append(cmds, []string{"chrome", "google-chrome", "chromium", "firefox", "sensible-browser"}...)
```
aalexand: Could you keep your comment about why the browser list is prioritized over xdg-open?

internal/driver/commands.go
```diff
@@ -347,6 +347,10 @@ func browsers() []string {
 	case "windows":
 		cmds = append(cmds, "cmd /c start")
 	default:
+		// commands opening browsers are prioritized over xdg-open, so browser()
```
aalexand: Nit: "Commands" - please keep comments proper sentences.

internal/driver/commands.go
nolanmar511: Done.

internal/driver/commands.go
nolanmar511: Done.

internal/driver/commands.go
```diff
@@ -336,25 +336,26 @@ func listHelp(c string, redirect bool) string {
...
 	case "windows":
 		cmds = append(cmds, "cmd /c start")
 	default:
+		// Commands opening browsers are prioritized over xdg-open, so browser()
+		// command can be used on linux to open the .svg file generated by the -web
+		// command (the .svg file includes embedded javascript so is best viewed in
+		// a browser).
+		cmds = append(cmds, []string{"chrome", "google-chrome", "chromium", "firefox", "sensible-browser"}...)
 		if os.Getenv("DISPLAY") != "" {
+			// xdg-open is only for use in a desktop environment.
```
aalexand: FWIW, I don't think there is much sense in opening Chrome when DISPLAY is not set either, but we can save it for a future change.

Issue #415, opened 2018-09-04
Discrepency between pprof list and weblist


ardan-bkennedy:
### What version of Go are you using (`go version`)?
```go version go1.11 darwin/amd64```

### Does this issue reproduce with the latest release?
YES

### What operating system and processor architecture are you using (`go env`)?
```
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/bill/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/bill/code/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/p6/l5y36lvx3zvdc3gtfh0g_43w0000gn/T/go-build393058492=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

### What did you do?

Using this code:
https://github.com/ardanlabs/gotraining/tree/master/topics/go/profiling/memcpu

I ran this set of commands:
```
$ go test -gcflags "-m -m" -run none -bench . -benchtime 3s -benchmem -memprofile p.out  
$ go tool pprof p.out  
(pprof) weblist algOne  
```

### What did you expect to see?

I saw the `weblist` view was different (and more accurate) then the `list` view

**weblist view**
```
Type: alloc_space
Time: Sep 3, 2018 at 10:39am (PDT)
Total: 325.53MB
github.com/ardanlabs/gotraining/topics/go/profiling/memcpu.algOne
/Users/bill/code/go/src/github.com/ardanlabs/gotraining/topics/go/profiling/memcpu/stream.go

  Total:    325.53MB   325.53MB (flat, cum)   100%
     78            .          .            
     79            .          .           // algOne is one way to solve the problem. 
[...]

PR #110, opened 2017-03-10
Remove filenames when showing function (default) granularity

rauls5382:
Previously we were showing filenames together with function names
when generating reports at function granularity. This was an attempt
to provide more information to the user, but it clutters the output.
Also, in cases where a function is associated to multiple files
(eg template instantiations in C++), keeping the filename may prevent
us from combining nodes from the same function.

Will stop keeping filenames at the default function granularity, and will
remove the "functionnameonly" option which was previously providing this
functionality.

This fixes #106

PR #420, opened 2018-09-20
Make '-noinlines' a separate flag, introduce '-filefunctions' granularity

aalexand:
This change consists of two relatively independent parts, but they are
both about handling the granularity, so bundling them together.

First, in #415 there is a discussion that having a granularity mode by
source lines but with inlines hidden would be useful. The agreement is
also that adding `-linenoinlines` granularity would make the granularity
flags too messy (and they are already somewhat messy with `-addresses`
and `-addressnoinlines`. So, it was proposed to make `-noinlines` a
separate flag, which is what this change does. Note that the flag is now
pulled out of the granularity group so it's a bit of backward
incompatible change but I think it is acceptable. For the example in
issue #415 the user would now be able to specify `-list foo -noinlines`
to produce annotated source where the metrics from the inlined functions
are attributed to the calling inliner line.

With this change, I am also dropping the `-addressnoinlines` granularity
which is now supported as `-addresses -noinlines` combination. I
couldn't find any usage of this option at least internally at Google, so
I think it's safe to remove it.

Second, I am adding a separate `-filefunctions` granularity which groups
the data by both function and file. This is a follow-up to #110 from the
past where we changed the `-functions` granularity to not group by file
(it used to), and since then there was a couple of reports where using
just function name alone would over-aggregate the data in cases when a
function with the same name is contained in multiple source files (e.g.
see b/18874275 internally).

Also, make a number of assorted documentation and `-help` fixes.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -80,21 +80,21 @@ other.
 
 Some common pprof options are:
 
-* **-flat [default]:** Sort entries based on their flat weight, on text reports.
-* **-cum:** Sort entries based on cumulative weight, on text reports.
-* **-functions [default]:** Accumulate samples at the function level; profile
-  locations that describe the same function will be merged into a report entry.
-* **-lines:** Accumulate samples at the source line level; profile locations that
-  describe the same function will be merged into a report entry.
-* **-addresses:** Accumulate samples at the instruction address; profile locations
-  that describe the same function address will be merged into a report entry.
+* **-flat** [default], **-cum**: Sort entries based on their flat or cumulative
+  weight respectively, on text reports.
+* **-functions** [default], **-filefunctions**, **-files**, **-lines**,
+  **-addresses**: Generate the report using the specified granularity.
+* **-noinlines**: Attribute inlined functions to their first out-of-line caller.
+  For example, a command like `pprof -list foo -noinlines profile.pb.gz` can be
+  used to produce the annotated source listing attributing the metrics in the
+  inlined functions to the out-of-line calling line.
 * **-nodecount= _int_:** Maximum number of entries in the report. pprof will only print
   this many entries and will use heuristics to select which entries to trim.
 * **-focus= _regex_:** Only include samples that include a report entry matching
   *regex*.
 * **-ignore= _regex_:** Do not include samples that include a report entry matching
   *regex*.
-* **-show\_from= _regex_:** Do not show entries above the first one that 
+* **-show\_from= _regex_:** Do not show entries above the first one that
   matches *regex*.
 * **-show= _regex_:** Only show entries that match *regex*.
 * **-hide= _regex_:** Do not show entries that match *regex*.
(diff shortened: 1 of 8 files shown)


doc/README.md
```diff
@@ -80,21 +80,21 @@ other.
...
 
-* **-flat [default]:** Sort entries based on their flat weight, on text reports.
-* **-cum:** Sort entries based on cumulative weight, on text reports.
-* **-functions [default]:** Accumulate samples at the function level; profile
-  locations that describe the same function will be merged into a report entry.
-* **-lines:** Accumulate samples at the source line level; profile locations that
-  describe the same function will be merged into a report entry.
-* **-addresses:** Accumulate samples at the instruction address; profile locations
-  that describe the same function address will be merged into a report entry.
+* **-flat** [default], **-cum** Sort entries based on their flat or cumulative
```
hyangah: need ':' before ' Sort'

internal/driver/driver.go
```diff
@@ -155,17 +155,17 @@ func applyCommandOverrides(cmd string, outputFormat int, v variables) variables
...
 	case "disasm", "weblist":
 		trim = false
-		v.set("addressnoinlines", "t")
+		v.set("addresses", "t")
+		v.set("noinlines", "t")
 	case "peek":
 		trim = false
 	case "list":
-		v.set("nodecount", "0")
+		trim = false
```
hyangah: explanation for this will be helpful for readers

aalexand: Done - added explanatory text above near the `trim` variable declaration.

aalexand:
@nolanmar511 @rauls5382 Could you re-approve the most recent changes please?

ardan-bkennedy:
What version will this change in be @aalexand

ardan-bkennedy:
The default on the console is -noinlines, I need the opposite, I want the list command to use inlining.

aalexand:
> What version will this change in be @aalexand

Do you mean which Go version? I am not 100% sure about that, Go team would know. @hyangah do you know if 1.12 is going to have this change in?

> The default on the console is -noinlines, I need the opposite, I want the list command to use inlining.

Is that with pprof from master branch of this repo?

ardan-bkennedy:
Inside pprof I need to be able to say

`list algOne -inline`

The default is to use `-noinlines`.

It hurts to have to use `weblist` to see the `inline` view.

hyangah:
@aalexand Vendored version in 1.12 is google/pprof@fde099a545debf81bf2a96a0ec13d7da2c2a6663, so includes google/pprof@7dadf64105bba38bab598048e7721cc9d34f7e75 

(while we are here - if there is any important fix to be cherry picked for 1.12, please file an issue)

ardan-bkennedy:
I had this issue and it was closed?

https://github.com/google/pprof/issues/415

Do I need to create a new one from this?

aalexand:
@ardan-bkennedy Do you use pprof from the master branch? The list command should have -inlines as the default, see https://github.com/google/pprof/blob/985cf9b4fdd2b479e4563ec57352005f69d5f9f6/internal/driver/driver.go#L180.

ardan-bkennedy:
Sorry, I didn't try the master branch. If this is the case, then we are good. Thanks for making that change. I can't wait for it to land.

PR #422, opened 2018-10-01
Add a comment on why 'noinlines' is forced for disasm reports.

aalexand:


aalexand:
@nolanmar511 Could you please re-approve this? I had a minor comment change.

PR #425, opened 2018-10-12
Fix ELF base calculation for exec mapping with offset != 0

aalexand:
I was looking at a report where pprof wouldn't symbolize the data
collected for a Chrome binary using Linux perf (b/112303003). The mmap
information is:

start: 000000000272e000, limit: 0000000006b09000, offset: 000000000252f000

The ELF file header:

elf.FileHeader{Class:elf.ELFCLASS64, Data:elf.ELFDATA2LSB, Version:elf.EV_CURRENT, OSABI:elf.ELFOSABI_NONE, ABIVersion:0x0, ByteOrder:binary.LittleEndian, Type:elf.ET_EXEC, Machine:elf.EM_X86_64, Entry:0x272e000}

The code segment:

elf.ProgHeader{Type:elf.PT_LOAD, Flags:elf.PF_X+elf.PF_R, Off:0x252f000,
Vaddr:0x272e000, Paddr:0x272e000, Filesz:0x43da610, Memsz:0x43da610,
Align:0x1000}

The dynamic loader here mapped 0x6b09000-0x272e000 = 0x43db000 bytes
starting 0x252f000 file offset into 0x272e000 virtual address, exactly
as instructed by the program header (so, no ASLR). Thus, the base
adjustment should be zero. Yet, the current GetBase produced the base of
0x252f000 which is wrong. The reason for that is that the ET_EXEC branch
of GetBase doesn't handle the general case of non-zero mmap file offset,
but rather only supports a couple of special cases. This change makes
handling the case of user-mode ET_EXEC more generic.

### DIFF
--- a/internal/elfexec/elfexec.go
+++ b/internal/elfexec/elfexec.go
@@ -178,8 +178,7 @@ func GetBase(fh *elf.FileHeader, loadSegment *elf.ProgHeader, stextOffset *uint6
 		pageOffsetPpc64 = 0xc000000000000000
 	)
 
-	if start == 0 && offset == 0 &&
-		(limit == ^uint64(0) || limit == 0) {
+	if start == 0 && offset == 0 && (limit == ^uint64(0) || limit == 0) {
 		// Some tools may introduce a fake mapping that spans the entire
 		// address space. Assume that the address has already been
 		// adjusted, so no additional base adjustment is necessary.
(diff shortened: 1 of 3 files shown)


internal/elfexec/elfexec.go
```diff
@@ -189,9 +188,15 @@ func GetBase(fh *elf.FileHeader, loadSegment *elf.ProgHeader, stextOffset *uint6
...
 	case elf.ET_EXEC:
 		if loadSegment == nil {
-			// Fixed-address executable, no adjustment.
+			// Assume fixed-address executable and so no adjustment.
 			return 0, nil
 		}
+		if stextOffset == nil && start > 0 && start < 0x8000000000000000 {
+			// A regular user-mode executable. Compute the base offset using same
+			// arithmetics as in ET_DYN case below, see the explanation there.
+			return start - offset + loadSegment.Off - loadSegment.Vaddr, nil
```
gmarin13: I think this formula is the general way to compute the base, at least for user mode addresses, so I am happy with this change.
But, can you add a comment on what stextOffset represents?

aalexand: Done.

internal/elfexec/elfexec.go
```diff
@@ -189,9 +188,21 @@ func GetBase(fh *elf.FileHeader, loadSegment *elf.ProgHeader, stextOffset *uint6
 	switch fh.Type {
 	case elf.ET_EXEC:
 		if loadSegment == nil {
-			// Fixed-address executable, no adjustment.
+			// Assume fixed-address executable and so no adjustment.
 			return 0, nil
 		}
+		if stextOffset == nil && start > 0 && start < 0x8000000000000000 {
```
gmarin13: Thanks for the explanation on the meaning of stextOffset.
If that value is optional, do you think it is safe to handle all user mode addresses here, regardless if stextOffset is nil or not?
If stextOffset != nil, then we fall into one of the crazy special cases below.

gmarin13: Oh, rereading your comment I see that the _stext symbol is in the vmlinux image.

PR #435, opened 2018-11-13
Update .travis.yml

AlekSi:
* bump Go versions;
* bump macOS / Xcode versions;
* use homebrew addon to install graphviz.


nolanmar511:
@aalexand -- Is there any reason we still need to support as mac OS earlier than xcode8.3, or golang before 1.10?

I assume we should upgrade, but wanted to check.

PR #436, opened 2018-11-13
add license to profile.proto

nolanmar511:
Not sure about the year for the license.

### DIFF
--- a/proto/profile.proto
+++ b/proto/profile.proto
@@ -1,3 +1,17 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


proto/profile.proto
```diff
@@ -1,3 +1,17 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
aalexand: What was the year in the previous version of the license header? Use that?

nolanmar511: Other than the accidentally added "// Copyright 2016 Google Inc.", we've never had any year here (I looked through past commits, and didn't see any apache header files)

I'm going to use 2016, I guess.

aalexand:
@nolanmar511 Please merge once CI passes and let the user who requested the change know.

PR #438, opened 2018-11-18
Support sign in tagfocus

fowles:
Allow tagfocus expressions to have explicit +/- prefix

### DIFF
--- a/internal/driver/driver_test.go
+++ b/internal/driver/driver_test.go
@@ -1152,6 +1152,20 @@ func TestTagFilter(t *testing.T) {
 			map[string][]string{"key1": {"tag1"}},
 			false,
 		},
+		{
+			"Match negative key and range of values, value matches",
+			"bytes=-512b:-128b",
+			map[string][]int64{"bytes": {-256}},
+			map[string]string{"bytes": "bytes"},
+			true,
+		},
+		{
+			"Match negative key and range of values, value outside range",
+			"bytes=-512b:-128b",
+			map[string][]int64{"bytes": {-2048}},
+			map[string]string{"bytes": "bytes"},
+			false,
+		},
 	}
 	for _, test := range tagFilterTests {
 		t.Run(test.desc, func(*testing.T) {
--- a/internal/driver/driver_test.go
+++ b/internal/driver/driver_test.go
@@ -1152,20 +1152,6 @@ func TestTagFilter(t *testing.T) {
 			map[string][]string{"key1": {"tag1"}},
 			false,
 		},
-		{
-			"Match negative key and range of values, value matches",
-			"bytes=-512b:-128b",
-			map[string][]int64{"bytes": {-256}},
-			map[string]string{"bytes": "bytes"},
-			true,
-		},
-		{
-			"Match negative key and range of values, value outside range",
-			"bytes=-512b:-128b",
-			map[string][]int64{"bytes": {-2048}},
-			map[string]string{"bytes": "bytes"},
-			false,
-		},
 	}
 	for _, test := range tagFilterTests {
 		t.Run(test.desc, func(*testing.T) {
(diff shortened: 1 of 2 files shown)


aalexand:
@fowles Can you add tests? For both negative and positive values.

Also, CI is failing on go master - if you could look into whether there is an easy way to fix it, it would be good. It's probably unrelated to your changes but we need to get it back green.

fowles:
Caught me. ;) 

Added a test.  The failure looks like something related to:

$ go tool vet -all .
vet: invalid command: want .cfg file (this reduced version of vet is intended to be run only by the 'go vet' command)
The command "go tool vet -all ." exited with 1.

but I don't actually know go (at all really), so I would prefer if you looked into this yourself.

internal/driver/driver_test.go
```diff
@@ -96,6 +96,7 @@ func TestParse(t *testing.T) {
 		{"peek=line.*01", "cpu"},
 		{"weblist=line[13],addresses,flat", "cpu"},
 		{"tags,tagfocus=400kb:", "heap_request"},
+		{"tags,tagfocus=+400kb:", "heap_request"},
```
aalexand: As I said, add a test for negative values as well?

fowles: man you are catching all my laziness.  That requires me understanding how you have rigged up all these which is more than a bit of logic...

fowles:
any progress in fixing your CI server and getting this merged?

internal/driver/driver_test.go
fowles: nope, as noted doing that requires understanding your tests.  The current test is a large integration test so it is not particularly amenable to understanding.  The code is simple enough and the positive case hits the same code path, so I strongly prefer to stop there.

aalexand: Sorry, I can't merge this without a test.

fowles: the test provided hits the correct path.  What you are asking is essentially to test if go's parse int routine handles negative signs.  (it does)

aalexand: There may be other reasons the code may not work nor or with future refactorings, other than parsing. For example, there could be some unsigned types along the path of handling the numeric tag filtering, as an example.

aalexand:
@fowles Thanks for adding the test. CI fails with a build failure now, please fix. Also rebase on top of current master to get #442 in.

fowles:
rebased on top.  I really have no idea where that build error is coming from.  It works on my local test...

aalexand:
@fowles It repros for me, see below. Check that you are running "go test ./...", i.e. recursively.

```
@azina:~$ cd ~/go/src/github.com/google/pprof

@azina:pprof$ git fetch origin pull/438/head:pr438
remote: Enumerating objects: 30, done.
remote: Counting objects: 100% (30/30), done.
remote: Compressing objects: 100% (18/18), done.
remote: Total 30 (delta 17), reused 23 (delta 12), pack-reused 0
Unpacking objects: 100% (30/30), done.
From https://github.com/google/pprof
 * [new ref]         refs/pull/438/head -> pr438
@azina:pprof$ git co pr438
M       internal/driver/driver.go
M       profile/filter.go
M       profile/profile.go
Switched to branch 'pr438'

@azina:pprof$ go test ./...
# github.com/google/pprof/internal/driver [github.com/google/pprof/internal/driver.test]
internal/driver/driver_test.go:1158:22: cannot use map[string][]int64 literal (type map[string][]int64) as type map[string][]string in field value
internal/driver/driver_test.go:1159:21: cannot use map[string]string literal (type map[string]string) as type bool in field value
internal/driver/driver_test.go:1160:4: too many values in composite literal
internal/driver/driver_test.go:1165:22: cannot use map[string][]int64 literal (type map[string][]int64) as type map[string][]string in field value
internal/driver/driver_test.go:1166:21: cannot use map[string]string literal (type map[string]string) as type bool in field value
internal/driver/driver_test.go:1167:4: too many values in composite literal
?       github.com/google/pprof [no test files]
?       github.com/google/pprof/driver  [no test files]
ok      github.com/google/pprof/fuzz    0.021s
ok      github.com/google/pprof/internal/binutils       0.104s
FAIL    github.com/google/pprof/internal/driver [build failed]
ok      github.com/google/pprof/internal/elfexec        0.050s
ok      github.com/google/pprof/internal/graph  0.016s
ok      github.com/google/pprof/internal/measurement    0.039s
?       github.com/google/pprof/internal/plugin [no test files]
?       github.com/google/pprof/internal/proftest       [no test files]
ok      github.com/google/pprof/internal/report 0.146s
ok      github.com/google/pprof/internal/symbolizer     0.055s
ok      github.com/google/pprof/internal/symbolz        0.048s
?       github.com/google/pprof/internal/transport      [no test files]
ok      github.com/google/pprof/profile 0.061s
?       github.com/google/pprof/third_party/d3  [no test files]
?       github.com/google/pprof/third_party/d3flamegraph        [no test files]
?       github.com/google/pprof/third_party/svgpan      [no test files]
```

fowles:
problem was that I was manually exporting diffs from google3/ into my git repo and had an error in that step.  Sorry, should be fixed now

PR #445, opened 2018-11-30
[Bug Fix] Propagate query args during URL redirect to preserve heap allocation type

richardartoul:
Master branch of PPROF has a bug introduced 8 months ago by #348 which breaks invocations like:

`pprof -http localhost:8080 -alloc_space <FILE>`

If you run that command, it will display `inuse_space` instead of `alloc_space`. The reason is that the decision to display `alloc_space` is propagated as a query arg, and the `http.RedirectHandler` that the existing code is using doesn't propagate them.

I verified the bug exists (and is fixed in this P.R) by running the same command described above.

### DIFF
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -184,13 +184,10 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
 	return s.Serve(ln)
 }
 
-func redirectWithQuery(url string) http.HandlerFunc {
-	urlWithQuery := url
+func redirectWithQuery(path string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		if queryStr := r.URL.Query().Encode(); queryStr != "" {
-			urlWithQuery = urlWithQuery + "?" + queryStr
-		}
-		http.Redirect(w, r, urlWithQuery, http.StatusTemporaryRedirect)
+		pathWithQuery := &gourl.URL{Path: path, RawQuery: r.URL.RawQuery}
+		http.Redirect(w, r, pathWithQuery.String(), http.StatusTemporaryRedirect)
 	}
 }
 
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -185,11 +185,12 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
 }
 
 func redirectWithQuery(url string) http.HandlerFunc {
+	urlWithQuery := url
 	return func(w http.ResponseWriter, r *http.Request) {
 		if queryStr := r.URL.Query().Encode(); queryStr != "" {
-			url = url + "?" + queryStr
+			urlWithQuery = urlWithQuery + "?" + queryStr
 		}
-		http.Redirect(w, r, url, http.StatusTemporaryRedirect)
+		http.Redirect(w, r, urlWithQuery, http.StatusTemporaryRedirect)
 	}
 }
 
--- a/internal/driver/webui.go
+++ b/internal/driver/webui.go
@@ -186,14 +186,10 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
 
 func redirectWithQuery(url string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		var (
-			urlWithQuery = url
-			queryStr     = r.URL.Query().Encode()
-		)
-		if queryStr != "" {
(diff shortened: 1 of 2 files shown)


internal/driver/webui.go
```diff
@@ -179,11 +179,24 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
...
+	mux.Handle("/", redirectWithQuery("/ui"))
 	s := &http.Server{Handler: mux}
 	return s.Serve(ln)
 }
 
+func redirectWithQuery(url string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var (
+			urlWithQuery = url
+			queryStr     = r.URL.Query().Encode()
```
nolanmar511: I feel like using `urlWithQuery := url` instead of var() would fit more in with the style of the code we have so far.

nolanmar511: I would probably be tempted to write lines 189-195  as:
```
if queryStr := r.URL.Query().Encode(); queryString != "" {
     url = url + "?" + queryStr
}
```

Then use url instead of urlWithQuery (I think that's equivalent).

nolanmar511:
We currently don't have testing for the web UI (hoping to get to it -- we know this bad), so could you comment on the manual testing in your description?

I can reproduce this bug -- Thanks for the fix!

Seems good to me overall (I added a comment, but seems like a nit). I'll let @aalexand take a look as well.

internal/driver/webui.go
richardartoul: sure, done

richardartoul:
@nolanmar511 Cool, updated as requested

internal/driver/webui.go
```diff
@@ -179,11 +179,20 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
...
-	mux.Handle("/", http.RedirectHandler("/ui/", http.StatusTemporaryRedirect))
+	mux.Handle("/", redirectWithQuery("/ui"))
 	s := &http.Server{Handler: mux}
 	return s.Serve(ln)
 }
 
+func redirectWithQuery(url string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if queryStr := r.URL.Query().Encode(); queryStr != "" {
+			url = url + "?" + queryStr
```
nolanmar511: Thanks for making this fix, I realized I had an error in my correction (really sorry about that).

Looking at the code now, it looks a bit risky to reassign the value of url within a nested function. Probably better to go back to something more like you originally had, so like:
```
urlWithQuery := url
if ... {
....
}
http.Redirect(w, r, urlWithQuery, http.StatusTemporaryRedirect)
```

richardartoul: no worries, updated it

internal/driver/webui.go
```diff
@@ -179,11 +179,21 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
...
-	mux.Handle("/", http.RedirectHandler("/ui/", http.StatusTemporaryRedirect))
+	mux.Handle("/", redirectWithQuery("/ui"))
 	s := &http.Server{Handler: mux}
 	return s.Serve(ln)
 }
 
+func redirectWithQuery(url string) http.HandlerFunc {
+	urlWithQuery := url
+	return func(w http.ResponseWriter, r *http.Request) {
+		if queryStr := r.URL.Query().Encode(); queryStr != "" {
```
aalexand: I feel this could be just

```
mux.Handle("/", func(w http.ResponseWriter, r *http.Request) {
  http.Redirect(w, r, url.URL{Path: "/ui/", RawQuery: r.URL.RawQuery}.String(), http.StatusTemporaryRedirect)
}
```

?

aalexand: I would inline the function to be a lambda to avoid mental load "oh but where redirectWithQuery is used? ah, just this one place" which is obvious with the inlined lambda. But it's fine as is.

richardartoul: I think it would have to be wrapped in http.HandlerFunc(...) which I didn't think was very readable, but I can make the change if you feel strongly

aalexand: I hoped it would implicitly cast, but I may be wrong. Not feeling strongly. Waiting for CI to pass before I merge.

aalexand: Thanks for the fix BTW!

richardartoul: yeah no problem, thanks for the useful tool, I use it almost everyday

PR #448, opened 2019-01-08
Use demangled symbols in disassembly report

kalyanac:
In the current disassembly report, not all symbols are demangled. This results in readability challenges with examining the report. 

This PR uses the -C option with objdump to demangle symbols, and updates disassembly parser to extract demangled symbols and ignore the corresponding demangled symbols.

This PR does not demangle the symbols itself but relies on 'objdump' to perform the demangling.

### DIFF
--- a/internal/binutils/disasm.go
+++ b/internal/binutils/disasm.go
@@ -25,10 +25,11 @@ import (
 )
 
 var (
-	nmOutputRE            = regexp.MustCompile(`^\s*([[:xdigit:]]+)\s+(.)\s+(.*)`)
-	objdumpAsmOutputRE    = regexp.MustCompile(`^\s*([[:xdigit:]]+):\s+(.*)`)
-	objdumpOutputFileLine = regexp.MustCompile(`^(.*):([0-9]+)`)
-	objdumpOutputFunction = regexp.MustCompile(`^(\S.*)\(\):`)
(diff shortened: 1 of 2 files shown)


aalexand:
@kalyanac Could you add a test that fails before the fix and passes after? It would also be good to describe in more detail which problem this PR addresses.

@rauls5382 FYI

internal/binutils/disasm.go
```diff
@@ -25,10 +25,11 @@ import (
...
 
 var (
-	nmOutputRE            = regexp.MustCompile(`^\s*([[:xdigit:]]+)\s+(.)\s+(.*)`)
-	objdumpAsmOutputRE    = regexp.MustCompile(`^\s*([[:xdigit:]]+):\s+(.*)`)
-	objdumpOutputFileLine = regexp.MustCompile(`^(.*):([0-9]+)`)
-	objdumpOutputFunction = regexp.MustCompile(`^(\S.*)\(\):`)
+	nmOutputRE                     = regexp.MustCompile(`^\s*([[:xdigit:]]+)\s+(.)\s+(.*)`)
+	objdumpAsmOutputRE             = regexp.MustCompile(`^\s*([[:xdigit:]]+):\s+(.*)`)
+	objdumpOutputFileLine          = regexp.MustCompile(`^(.*):([0-9]+)`)
+	objdumpOutputFunction          = regexp.MustCompile(`^(\S.*)\(\):`)
```
rauls5382: This is now unused. remove it.

PR #449, opened 2019-01-10
use simplified demangled symbol names in disassembly report

kalyanac:
In the current disassembly report, not all symbols are demangled. This results in readability challenges with examining the report.

This PR demangles all symbol names while generating the disassembly report. By default, demangling of function and template parameters is disabled to generate a simplified symbol name.

If a symbol cannot be demangled, it is printed as is.


aalexand:
@kalyanac Add a test? The code change is small but I think we should still cover it.

kalyanac:
sorry for the delay, added a test as requested

kalyanac:
the failures are on OS X with objdump. I will take a look at modifying the test for OS X.

kalyanac:
Mac tests using XCode 8.3 are all failing because `objdump` included in XCode 8.3 does not support `-start-address` and `-stop-address`

```
./objdump -version
Apple LLVM version 8.1.0 (clang-802.0.42)
  Optimized build.
  Default target: x86_64-apple-darwin18.2.0
  Host CPU: broadwell

./objdump -disassemble-all -no-show-raw-insn -line-numbers -start-address=0x100001cd0 -stop-address=0x100001cef ~/go/src/github.com/google/pprof/internal/report/testdata/disasm.mac.bin
objdump: Unknown command line argument '-start-address=0x100001cd0'.  Try: './objdump -help'
objdump: Did you mean '-headers=0x100001cd0'?
objdump: Unknown command line argument '-stop-address=0x100001cef'.  Try: './objdump -help'
objdump: Did you mean '-headers=0x100001cef'?
```

```
objdump -version
Apple LLVM version 10.0.0 (clang-1000.11.45.2)
  Optimized build.
  Default target: x86_64-apple-darwin18.2.0
  Host CPU: skylake

objdump -disassemble-all -no-show-raw-insn -line-numbers -start-address=0x100001cd0 -stop-address=0x100001cef ~/go/src/github.com/google/pprof/internal/report/testdata/disasm.mac.bin

/Users/kchadalavada/go/src/github.com/google/pprof/internal/report/testdata/disasm.mac.bin:	file format Mach-O 64-bit x86-64

Disassembly of section __TEXT,__text:
__text:
100001cd0:	pushq	%rbp
100001cd1:	movq	%rsp, %rbp
100001cd4:	movl	%edi, -4(%rbp)
100001cd7:	movl	%esi, -8(%rbp)
100001cda:	movl	-4(%rbp), %esi
100001cdd:	cmpl	-8(%rbp), %esi
100001ce0:	sete	%al
100001ce3:	andb	$1, %al
100001ce5:	movzbl	%al, %eax
100001ce8:	popq	%rbp
100001ce9:	retq
100001cea:	nopw	(%rax,%rax)

___clang_call_terminate:

__ZNSt3__111char_traitsIcE11eq_int_typeEii:
100001cd0:	pushq	%rbp
100001cd1:	movq	%rsp, %rbp
100001cd4:	movl	%edi, -4(%rbp)
100001cd7:	movl	%esi, -8(%rbp)
100001cda:	movl	-4(%rbp), %esi
100001cdd:	cmpl	-8(%rbp), %esi
100001ce0:	sete	%al
100001ce3:	andb	$1, %al
100001ce5:	movzbl	%al, %eax
100001ce8:	popq	%rbp
100001ce9:	retq
100001cea:	nopw	(%rax,%rax)
```

Since XCode 8.3 is pretty [old](https://en.wikipedia.org/wiki/Xcode#8.x_series), can we drop support for it?

kalyanac:
Added code to look for objdump before starting the test. 
The test function will not have access to binutils initialization routine from within the test function. So the test function tries to determine the objdump version only if it is in the path.

internal/report/report.go
```diff
@@ -428,7 +432,13 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
...
-		fmt.Fprintf(w, "ROUTINE ======================== %s\n", s.sym.Name[0])
+		if strings.HasPrefix(s.sym.Name[0], "__Z") {
+			// Workaround to handle double leading underscores on Mac
+			// which are not properly handled by the demangle.Filter
+			fmt.Fprintf(w, "ROUTINE ======================== %s\n", demangle.Filter(s.sym.Name[0][1:], options...))
+		} else {
+			fmt.Fprintf(w, "ROUTINE ======================== %s\n", demangle.Filter(s.sym.Name[0], options...))
+		}
 		for _, name := range s.sym.Name[1:] {
 			fmt.Fprintf(w, "    AKA ======================== %s\n", name)
```
rauls5382: Should we demangle these ones too?

internal/report/report.go
```diff
@@ -428,7 +432,13 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 
 		ns := annotateAssembly(insts, sns, s.base)
 
-		fmt.Fprintf(w, "ROUTINE ======================== %s\n", s.sym.Name[0])
+		if strings.HasPrefix(s.sym.Name[0], "__Z") {
+			// Workaround to handle double leading underscores on Mac
+			// which are not properly handled by the demangle.Filter
```
rauls5382: Could we open an issue against the demangle package for this?

internal/report/report.go
```diff
@@ -413,6 +414,9 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 		}
 	}
 
+	// demangle symbols by default for disassembly report
+	options := []demangle.Option{demangle.NoParams, demangle.NoTemplateParams}
```
rauls5382: I would've preferred to have a way to control this, (eg with symbolize=demangle=xxx). Sometimes having the mangled name is useful

internal/binutils/binutils.go
```diff
@@ -163,6 +163,14 @@ func (bu *Binutils) Disasm(file string, start, end uint64) ([]plugin.Inst, error
 		fmt.Sprintf("--start-address=%#x", start),
 		fmt.Sprintf("--stop-address=%#x", end),
 		file)
+
+	if runtime.GOOS == "darwin" {
```
rauls5382: So without this disasm doesn't work on Mac at all? If that's the case perhaps we should put this on its own PR

internal/report/report.go
kalyanac: https://github.com/ianlancetaylor/demangle/issues/12

internal/binutils/binutils.go
kalyanac: https://github.com/google/pprof/pull/534

nolanmar511:
@kalyanac -- Does it make sense to try to get this PR merged in still?

kalyanac:
No, it should be addressed by #534

PR #462, opened 2019-03-08
Add option to disable browser in HTTP mode

### LINKED ISSUE #458: HTTP mode needs a !browser option

As downloaded via `go get -u github.com/google/pprof` today. Not sure how to get a better version ID. Go version 1.10.2.

Linux x86-64.

### What did you do?

I frequently run pprof on a remote machine via SSH. The HTTP interface--which is great, btw, I really wish I'd had this years ago!--launches a browser on that remote machine.  This is not useful for me (I do in fact run Xforwarding, but the remote browser isn't particularly good.) I'd much rather use my local workstation's browswer to visit <remote:port>, which does in fact work.  But I'd like to skip launching the browser.

Via source digging, I worked out that [it checks if stdout is interactive](https://github.com/google/pprof/blob/e84dfd68c163c45ea47aa24b3dc7eaa93f6675b1/pprof.go#L101) and thus the very silly-looking "pprof -http : <profile> > /dev/null" works but this is, well, silly, and hardly discoverable. This wants a flag or similar way to say "no thanks, skip the browser please."

Related, but ironic: if you don't open a browser, there's no command-line output of the path we're serving on. Since by default (with -http : or similar) the port is randomly chosen, we can't know it. So even with a  >/dev/null or | tee hack, I have to write out "-http :12345" or something like that so I can know where to look. So I feel pretty strongly we should log something like: 

"Serving UI on http://localhost:19553"

(I don't care at all about the format so long as there's a pastable URL in there.)

Thanks!

ahh:
Fixes #458


kalyanac:
thank you for the PR. 
Can you also please add the new option to the usage text? It can be added to usageMsgVars in cli.go.

driver/driver.go
```diff
@@ -228,6 +228,9 @@ type UI interface {
 	// WantBrowser indicates whether browser should be opened with the -http option.
 	WantBrowser() bool
 
+	// DisableBrowser forces WantBrowser() to return false.
+	DisableBrowser()
```
aalexand: I wouldn't want to make changes to this interface unless it's absolutely required. Can we instead make the value of the "no browser" flag available to webui.go and handle the decision of not opening the browser there?

ahh:
Added usage text as requested.

driver/driver.go
ahh: Sure. I figured you'd prefer to have "WantBrowser" return the truthful value, but it's just as easy to pass it through as a bool.

ahh: Done.

internal/driver/cli.go
```diff
@@ -118,6 +120,12 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 		return nil, nil, errors.New("-http is not compatible with an output format on the command line")
 	}
 
+	if *flagNoBrowser {
```
aalexand: `if *flagNoBrowser && *flagHTTP == "" {`

pprof.go
```diff
@@ -42,12 +42,17 @@ type readlineUI struct {
 	rl *readline.Instance
 }
 
+func isTerminal() bool {
```
aalexand: It doesn't look like changes in this file are required?

ahh: Reverted.

internal/driver/cli.go
ahh: Done.

internal/driver/cli.go
```diff
@@ -66,6 +67,7 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flagTools := flag.String("tools", os.Getenv("PPROF_TOOLS"), "Path for object tool pathnames")
 
 	flagHTTP := flag.String("http", "", "Present interactive web based UI at the specified http host:port")
```
aalexand: "interactive web based UI" -> "interactive web UI" to make it a bit shorter and consistent with description of the new flag one line below.

aalexand:
@ahh Oh, could you also name the PR properly? The current name makes no sense.

internal/driver/cli.go
ahh: Done.

ahh:
Renamed.

aalexand:
The CI seems to be failing with build error below. When fixing that, consider adding a test for the new code btw?

```
ok  	github.com/google/pprof/fuzz	0.124s
# github.com/google/pprof/internal/driver [github.com/google/pprof/internal/driver.test]
internal\driver\webui_test.go:57:22: not enough arguments in call to serveWebInterface
	have (string, *profile.Profile, *plugin.Options)
	want (string, *profile.Profile, *plugin.Options, bool)
=== RUN   TestAddr2Liner
```

ahh:
Woops!

Fixed the test failure.

I am not clear if this is testable; openBrowswer just invokes a bunch of
commands like "google-chrome" etc.  I don't see an easy way to check if
that's happening without nasty side effects in tests like opening browsers
(hence why the test UI instance sets WantBrowser() = false unconditionally.)

Any thoughts?

Issue #338, opened 2018-03-21
Tests fail on CentOS 6.9


nehaljwani:
xref: https://github.com/golang/go/issues/23888

### What version of pprof are you using?

a74ae6fb3cd7047c79272e3ea0814b08154a2d3c

### What operating system and processor architecture are you using?

CentOS 6.9, x86_64

### What did you do?

Compiled go from source, tests failed, came to this repo, tried running the tests, was able to reproduce the exact same failure.

### What did you expect to see?

It'd be great if the tests had passed!

### What did you see instead?

The following tests failed:
```
--- FAIL: TestObjFile (0.02s)
    --- FAIL: TestObjFile/fake_mapping (0.01s)
        binutils_test.go:252: SourceLine: unexpected error write |1: broken pipe
    --- FAIL: TestObjFile/fixed_load_address (0.01s)
        binutils_test.go:252: SourceLine: unexpected error write |1: broken pipe
    --- FAIL: TestObjFile/simulated_ASLR_address (0.01s)
        binutils_test.go:252: SourceLine: unexpected error write |1: broken pipe
```

### Why do these tests fail?

The code at https://github.com/google/pprof/blob/a74ae6fb3cd7047c79272e3ea0814b08154a2d3c/internal/binutils/addr2liner.go#L92 assumes that the `-a` flag is available to the binary `addr2line`, which is available if the binutils version is >=2.21 ; but the latest version available on CentOS 6.9 is 2.20

bsed:
ok  	cmd/pack	2.860s
ok  	cmd/trace	0.060s
--- FAIL: TestDisasm (0.01s)
	binutils_test.go:189: Disasm: unexpected error [/home/kelvin/go/pkg/tool/linux_amd64/objdump -d -C --no-show-raw-insn -l --start-address=0x0 --stop-address=0xffffffffffffffff testdata/hello]: exit status 2
--- FAIL: TestObjFile (0.09s)
	binutils_test.go:231: SourceLine: unexpected error write |1: broken pipe
FAIL
FAIL	cmd/vendor/github.com/google/pprof/internal/binutils	0.121s
ok  	cmd/vendor/github.com/google/pprof/internal/driver	12.305s
ok  	cmd/vendor/github.com/google/pprof/internal/elfexec	0.011s
ok  	cmd/vendor/github.com/google/pprof/internal/graph	0.019s
ok  	cmd/vendor/github.com/google/pprof/internal/measurement	0.024s
[...]

Issue #146, opened 2017-06-20
TestHttpsInsecure failures on certain darwin builders


ALTree:
We're getting `TestHttpsInsecure` failures on some (but not all) of the darwin builders.

OSX 10.11 and 10.12 are fine.
The OSX 10.8 and 10.10 builders fail. The darwin-arm-a1549ios builder is also failing.

This is the OSX error:

```
--- FAIL: TestHttpsInsecure (10.04s)
	fetch_test.go:305: grabProfile(https+insecure://127.0.0.1:59997/debug/pprof/profile) got PeriodType: cpu nanoseconds
		Period: 10000000
		Time: 2017-06-20 11:36:19.953808169 -0700 PDT
		Duration: 10.0
		Samples:
		samples/count cpu/nanoseconds
		        492 4920000000: 1 2 3 4 5 6 
		Locations
		     1: 0x105efd3 M=1 runtime.kevent /var/folders/6g/8k13h11n2zb8hxx99wk6f6y40000gn/T/workdir/go/src/runtime/sys_darwin_amd64.s:526 s=0
		     2: 0x102a672 M=1 runtime.netpoll /var/folders/6g/8k13h11n2zb8hxx99wk6f6y40000gn/T/workdir/go/src/runtime/netpoll_kqueue.go:79 s=0
		     3: 0x103405a M=1 runtime.findrunnable /var/folders/6g/8k13h11n2zb8hxx99wk6f6y40000gn/T/workdir/go/src/runtime/proc.go:2099 s=0
		     4: 0x10349eb M=1 runtime.schedule /var/folders/6g/8k13h11n2zb8hxx99wk6f6y40000gn/T/workdir/go/src/runtime/proc.go:2237 s=0
		     5: 0x1034d05 M=1 runtime.park_m /var/folders/6g/8k13h11n2zb8hxx99wk6f6y40000gn/T/workdir/go/src/runtime/proc.go:2300 s=0
		     6: 0x105aeda M=1 runtime.mcall /var/folders/6g/8k13h11n2zb8hxx99wk6f6y40000gn/T/workdir/go/src/runtime/asm_amd64.s:286 s=0
		Mappings
		1: 0x0/0x0/0x0   [FN]
		, want function "TestHttpsInsecure"
FAIL
FAIL	cmd/vendor/github.com/google/pprof/internal/driver	10.545s
```

[...]

PR #464, opened 2019-03-12
Drop XCode 8.3

kalyanac:
Dropping XCode 8.3 per offline discussion with @aalexand

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,15 +11,6 @@ matrix:
     - os: linux
       go: master
     - os: osx
-      osx_image: xcode8.3
-      go: 1.10.x
-    - os: osx
-      osx_image: xcode8.3
-      go: 1.11.x
-    - os: osx
-      osx_image: xcode8.3
-      go: master
-    - os: osx
       osx_image: xcode9.4
       go: 1.10.x
     - os: osx


aalexand:
@kalyanac Could you check whether the travis CI config is in sync with the XCode versions used for OSX testing by build.golang.org? And bring it in sync if there is a trivial way to do so (i.e. if needed versions are supported by Travis CI). We've been trying to be in sync with that so that vendoring pprof into golang does not bring up failing tests.

kalyanac:
it looks like they only test against
10.10 (Yosemite), 10.11 (El Capitan), and 10.12 (Sierra). 

> We only have builders for 10.10, 10.11 and 10.12 as of 2019-01-11.

Additional information here:
https://github.com/golang/go/wiki/MinimumRequirements#macos-n%C3%A9e-os-x-aka-darwin

They dont yet support 10.14. 

The support matrix is here:
https://github.com/golang/go/wiki/Darwin

We can identify failing tests and skip them on unsupported OS X versions to address the failing tests on older OS X versions. It will add maintenance burden though.

aalexand:
@kalyanac Versioning of XCode vs. OSX is a bit confusing here. With this PR we'll test xcode9.4 and xcode10.1 which is according to https://docs.travis-ci.com/user/reference/osx/#macos-version are both macOS 10.13. Excluding xcode8.3 we'll stop testing macOS 10.12. Is that OK?

It's also not clear which XCode versions the golang builders use.

I am basically cautious about two things:

1) If you want to drop support for xcode8.3 because some test can't be made passing for it, then when pprof gets vendored into golang the test will fail if golang builders use xcode 8.3.
2) Generally having too much disjoint environments with golang builders may cause some bugs to slip through and only manifest when we vendor pprof into golang tree. We had some history on that, see #146, #156, #255.

kalyanac:
It is not clear what XCode versions golang builders are using. However, it is a good estimate that the latest XCode that can be run on 10.12.6 is 9.2. ref: https://en.wikipedia.org/wiki/Xcode#Xcode_7.0_-_10.x_(since_Free_On-Device_Development). 

We can make disassembly work with 8.3 but it will introduce small but version specific code paths. Also disassembly of large binaries maybe slower due to lack of `--start-address` & `--stop-address` in objdump.

I was leaning towards minimizing version specific code paths in the code and also becasue XCode 8.3 is pretty old, but the benefits of working well within golang verndoring is important as evident from the issues you linked.

With the above caveats, I can change #449 and drop this PR.

aalexand:
The easiest for now might be to check explicitly in the test if objdump supports these flags and t.Skip() the test if not. I am not a fan of this added complexity either, just want to make sure the future vendoring of pprof goes smooth. For large binaries - that's if we try to emulate the absent flags? I wouldn't go there, I think it's reasonable to require that objdump has to be of specific version or newer and skip the test otherwise.

BTW, #338 is similar in spirit.

PR #468, opened 2019-04-18
Allow to ignore build ID when reading local files.

segfaulthunter:
I ran into a case where the Build IDs did not literally match (they had
different padding), but I knew it would be fine to use them. It would be
nice to have a escape hatch flag for these situations.

Please take a look. Thanks!

### DIFF
--- a/internal/driver/cli.go
+++ b/internal/driver/cli.go
@@ -38,6 +38,7 @@ type source struct {
 	HTTPHostport       string
 	HTTPDisableBrowser bool
 	Comment            string
+	IgnoreBuildID      bool
 }
 
 // parseFlags parses the command lines through the specified flags package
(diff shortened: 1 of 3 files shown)


aalexand:
I am slightly uncomfortable about having such escape flag since it's easy to misuse it, get incorrecty symbolized data and make wrong decisions about what needs to be optimized.

Could you describe in more details what was the padding issue? I wonder if we can update the build ID read / comparison code instead.

segfaulthunter:
The Build IDs come from simpleperf in Android. I got the following error: 605edb7a6fd07cee833856b94d99b87f00000000 != 605edb7a6fd07cee833856b94d99b87f.

I still think it is useful to have a flag to override this check, because it is frustrating if the tool prevents you from doing what you know is okay without a flag to override (what I did initially was clone the repository and remove the check manually). Would it be better to give it a more scary name and documentation? Something like force_ignore_buildid?

aalexand:
I think we should look into where these zeros are coming from and try to
fix that. I don't really want to add flags to workaround potential bugs in
pprof or other tools.

@rauls5382 for opinion

segfaulthunter:
There are other tools that generate pprof compatible protos. In Android we have at least [simpleperf](https://developer.android.com/ndk/guides/simpleperf) and [heapprofd](https://docs.perfetto.dev/#/heapprofd).

segfaulthunter:
Sorry about the close / reopen. I misclicked in the UI.

rauls5382:
I agree that we would rather avoid adding flags to pprof to workaround bugs on other tools.
Merging profiles from mismatched build ids would result on nonsensical results.

If anything I would prefer **temporarily** relaxing the build id check to ignore trailing zeroes if there's a matching prefix, while printing an informational message. That way things would work without users having to discover a new flag.

aalexand:
Closing as it doesn't look we want this change in.

PR #469, opened 2019-04-18
Correct help docs for specifying a binary

nolanmar511:
When the binary is specified after the options, it's treated like another profile.

I believe the binary needs to be specified first.


kalyanac:
I just tried this and it works fine. nocancel is the binary
`pprof  -disasm nocancel ~/pprof/pprof.cppbench_server_main.samples.cpu.001.pb.gz`

nolanmar511:
I'm mistaken. Closing this PR.

PR #471, opened 2019-05-15
Add flamegraph to list of dynamic search links

### LINKED ISSUE #431: UI filters don't seem to apply to flame graph view.
Playing with the pprof web UI, I found that I could apply filters in (for example) the graph view, and those filters would also be applied to all views, excluding the flame graph. I also wasn't apply to add filters when I was in the flame graph view.

timpalpant:
This adds `flamegraph` to the list of dynamic search links so that query parameters (esp. regex filters) are retained when navigating to/from the flamegraph view. This seems to just be an oversight since flamegraph was added later as a visualization option.

I have tested it manually by applying filters in the "Graph" view and then navigating to "Flame Graph". You can also see that, after this change, the `href` for the flamegraph link is dynamically updated with any query parameters like the other links:

fix-flamegraph

Previously, the `href` would always be fixed with `./flamegraph` (from the template), so navigating to the flamegraph would reset all filters and other query parameters.

Fixes #431

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -853,7 +853,7 @@ function viewer(baseUrl, nodes) {
     toptable.addEventListener('touchstart', handleTopClick);
   }
 
-  const ids = ['topbtn', 'graphbtn', 'peek', 'list', 'disasm',
+  const ids = ['topbtn', 'graphbtn', 'flamegraph', 'peek', 'list', 'disasm',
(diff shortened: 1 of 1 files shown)


nolanmar511:
It looks good to me; I also downloaded and manually tested this change -- it works as described.

Leaving @aalexand or @kalyanac to approve or make further comments.

aalexand:
Looks good to me.

nolanmar511:
timpalpant, thanks for this fix!!!

PR #474, opened 2019-06-07
Look for symbols in debuginfo-compatible paths

rjogrady:
Add paths compatible with gdb-style debuginfo when looking for
symbols. These are elfs containing just the symbols separated
from stripped binaries and are stored in paths like
.build-id/xx/xxxxxx.debug
where the directory is the first two characters of the build-id,
and the filename is the remaining.

On a typical Unix system these will be in
/usr/lib/debug, but also look in $PROF_BINARY_PATH as well.


internal/driver/fetch.go
```diff
@@ -401,34 +401,42 @@ mapping:
...
 		}
-
-		for _, path := range filepath.SplitList(searchPath) {
-			var fileNames []string
-			if m.BuildID != "" {
-				fileNames = []string{filepath.Join(path, m.BuildID, baseName)}
+		var fileNames []string
+		if m.BuildID != "" {
+			// gdb-style debuginfo format puts debug symbols into paths like .build-id/01/23456789abcdef.debug
+			debugInfoName := filepath.Join(".build-id", m.BuildID[:2], m.BuildID[2:]+".debug")
```
aalexand: This should probably check len(m.BuildID)?

internal/driver/fetch.go
```diff
@@ -401,34 +401,42 @@ mapping:
...
+				// Standard global gdb search path.
+				fileNames = append(fileNames, filepath.Join("/usr/lib/debug", debugInfoName))
+			}
+			for _, path := range filepath.SplitList(searchPath) {
+				fileNames = append(fileNames, filepath.Join(path, m.BuildID, baseName))
 				if matches, err := filepath.Glob(filepath.Join(path, m.BuildID, "*")); err == nil {
 					fileNames = append(fileNames, matches...)
 				}
 				fileNames = append(fileNames, filepath.Join(path, m.File, m.BuildID)) // perf path format
+				fileNames = append(fileNames, filepath.Join(path, debugInfoName))     // debuginfo format
```
aalexand: Should this be under `if runtime.GOOS != "windows" {` condition?

internal/driver/fetch.go
rjogrady: Good point.

rjogrady:
anything else to do here?

internal/driver/fetch.go
```diff
@@ -396,39 +396,54 @@ func locateBinaries(p *profile.Profile, s *source, obj plugin.ObjTool, ui plugin
...
+				// Standard global gdb search path.
+				fileNames = append(fileNames, filepath.Join("/usr/lib/debug", debugInfoName))
+			}
+			for _, path := range filepath.SplitList(searchPath) {
+				fileNames = append(fileNames, filepath.Join(path, m.BuildID, baseName))
 				if matches, err := filepath.Glob(filepath.Join(path, m.BuildID, "*")); err == nil {
 					fileNames = append(fileNames, matches...)
 				}
 				fileNames = append(fileNames, filepath.Join(path, m.File, m.BuildID)) // perf path format
+				if runtime.GOOS != "windows" && debugInfoName != "" {
```
aalexand: It might be easier to read the whole gdb search path support as a separate condition and loop:

```go
searchPaths := filepath.SplitList(searchPath)
...
if runtime.GOOS != "windows" && len(m.BuildID) == 40 {
  // gdb-style debuginfo expects a 160-bit sha-1 string as a build id.
  // Look in paths like .build-id/ab/cdef0123456.debug
  debugInfoName := filepath.Join(".build-id", m.BuildID[:2], m.BuildID[2:]+".debug")
  for _, path := range append(searchPaths, "/usr/lib/debug") {
    fileNames = append(fileNames, filepath.Join(path, debugInfoName)) // debuginfo format
  }
}
```

aalexand:
Can you add notes in the PR description on what real testing apart from unit tests was done? I'm more hesitant about accepting a PR when the code quality is not good or tests are not convincing or there is no evidence that the author thoroughly tested their change manually. So please test it as thoroughly as you can and record the trail of that in the description.

aalexand:
@rjogrady I've updated / rebased the change - is it ready to be re-reviewed? I.e. were all comments adressed? Thanks!

PR #478, opened 2019-07-22
Fix keypress focus: updateUrl is not defined

### LINKED ISSUE #463: the search doesn't work in the source view in the webview

Commit 77426154d5469500277f91e6c9d64c93b50afc6d

Linux amd64

### What did you do?

There's a problem with the search area while in the source view. With the profile attached ([foobar.pprof.zip](https://github.com/google/pprof/files/2952676/foobar.pprof.zip) and here's the source code I used to generated that profile [gen_pprof.go.txt](https://github.com/google/pprof/files/2952691/gen_pprof.go.txt)), if I go in the source view and search for "findrunnable" for example nothing is happening and I get a bunch of errors in the console:

    source:3670 Uncaught ReferenceError: updateUrl is not defined
        at HTMLInputElement.handleKey (source:3670)
    handleKey @ source:3670
    
    2source:3710 Uncaught TypeError: Cannot read property 'length' of null
        at selectMatching (source:3710)
    selectMatching @ source:3710
    setTimeout (async)
    handleSearch @ source:3680

    source:3670 Uncaught ReferenceError: updateUrl is not defined
        at HTMLInputElement.handleKey (source:3670)
    handleKey @ source:3670

    source:3710 Uncaught TypeError: Cannot read property 'length' of null
        at selectMatching (source:3710)
    selectMatching @ source:3710
    setTimeout (async)
    handleSearch @ source:3680

However the function does exist, if I go directly to http://localhost:5600/ui/source?f=findrunnable then it's fine. Same if I'm in the flame graph, click on a box and then go to View -> Source.

I expect the search to work.

### What did you see instead?

Search doesn't do anything.

timpalpant:
`updateUrl` was refactored in away in #414, leading to `Uncaught ReferenceError: updateUrl is not defined` (and the keypress no longer works).

For the `sourcelisting` page, [`nodes===null`](https://github.com/google/pprof/blob/master/internal/driver/webhtml.go#L1019), leading to: `Uncaught TypeError: Cannot read property 'length' of null`.

These fixes were validated through manual testing. I believe given the scope of the changes they are unlikely to cause distant regressions.

Fixes #463.

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -610,8 +610,9 @@ function viewer(baseUrl, nodes) {
 
   function handleKey(e) {
     if (e.keyCode != 13) return;
-    window.location.href =
-        updateUrl(new URL(window.location.href), 'f');
+    setHrefParams(window.location, function (params) {
+      params.set('f', search.value);
+    });
     e.preventDefault();
   }
 
@@ -650,9 +651,11 @@ function viewer(baseUrl, nodes) {
     })
 
     // add matching items that are not currently selected.
-    for (let n = 0; n < nodes.length; n++) {
-      if (!selected.has(n) && match(nodes[n])) {
-        select(n, document.getElementById('node' + n));
+    if (nodes) {
+      for (let n = 0; n < nodes.length; n++) {
+        if (!selected.has(n) && match(nodes[n])) {
+          select(n, document.getElementById('node' + n));
+        }
       }
     }
 


timpalpant:
The Travis build looks broken for unrelated reasons, will try to fix it in #479.

aalexand:
The CI failed with what seems a fluke so I've restarted it.

PR #479, opened 2019-07-22
Fix Travis build call to staticcheck

timpalpant:
`staticcheck`'s `-ignore` flag [has been removed](http://staticcheck.io/changes/2019.1) in favor of [Linter directives](http://staticcheck.io/docs/#ignoring-problems), and the `unused` tool was merged into `staticcheck`, breaking the Travis build. These lint ignores may not be necessary anymore anyway due to improvements in the static checker.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -23,13 +23,7 @@ echo "mode: $MODE" > coverage.txt
 # All packages.
 PKG=$(go list ./...)
 
-staticcheck -ignore '
-github.com/google/pprof/internal/binutils/binutils_test.go:SA6004
-github.com/google/pprof/internal/driver/svg.go:SA6004
-github.com/google/pprof/internal/report/source_test.go:SA6004
-github.com/google/pprof/profile/filter_test.go:SA6004
-' $PKG
-unused $PKG
+staticcheck $PKG
 
 # Packages that have any tests.
 PKG=$(go list -f '{{if .TestGoFiles}} {{.ImportPath}} {{end}}' ./...)


internal/driver/webhtml.go
```diff
@@ -610,8 +610,9 @@ function viewer(baseUrl, nodes) {
 
   function handleKey(e) {
     if (e.keyCode != 13) return;
-    window.location.href =
-        updateUrl(new URL(window.location.href), 'f');
+    setHrefParams(window.location, function (params) {
```
aalexand: This change seems unrelated.

test.sh
```diff
@@ -23,13 +23,7 @@ echo "mode: $MODE" > coverage.txt
...
 PKG=$(go list ./...)
 
-staticcheck -ignore '
-github.com/google/pprof/internal/binutils/binutils_test.go:SA6004
-github.com/google/pprof/internal/driver/svg.go:SA6004
-github.com/google/pprof/internal/report/source_test.go:SA6004
-github.com/google/pprof/profile/filter_test.go:SA6004
-' $PKG
-unused $PKG
+staticcheck $PKG
```
aalexand: Could you check that this command still does the staticcheck? E.g. make some change in the sources which should clearly violate the check and show that it's found by running test.sh.

Thank you!

internal/driver/webhtml.go
timpalpant: Sorry, I rebased incorrectly. Fixed.

test.sh
timpalpant: Good idea :)

Here's an example where I added code to violate [SA5000](https://staticcheck.io/docs/checks#SA5000) (assignment to nil map): 

```
diff --git a/internal/driver/svg.go b/internal/driver/svg.go
index 62767e7..c5b9155 100644
--- a/internal/driver/svg.go
+++ b/internal/driver/svg.go
@@ -35,6 +35,9 @@ func massageSVG(svg string) string {
        // resulting on unparsable SVG.
        svg = strings.Replace(svg, "&;", "&amp;;", -1)
 
+       var testStatic map[string]string = nil
+       testStatic["hello"] = "world"
+
        // Dot's SVG output is
        //
        //    <svg width="___" height="___"
```

And running test.sh locally catches it:

```
palpant-mbp:pprof palpant$ ./test.sh 
++ MODE=atomic
++ echo 'mode: atomic'
+++ go list ./...
++ PKG='github.com/google/pprof
github.com/google/pprof/driver
github.com/google/pprof/fuzz
github.com/google/pprof/internal/binutils
github.com/google/pprof/internal/driver
github.com/google/pprof/internal/elfexec
github.com/google/pprof/internal/graph
github.com/google/pprof/internal/measurement
github.com/google/pprof/internal/plugin
github.com/google/pprof/internal/proftest
github.com/google/pprof/internal/report
github.com/google/pprof/internal/symbolizer
github.com/google/pprof/internal/symbolz
github.com/google/pprof/internal/transport
github.com/google/pprof/profile
github.com/google/pprof/third_party/d3
github.com/google/pprof/third_party/d3flamegraph
github.com/google/pprof/third_party/svgpan'
++ staticcheck github.com/google/pprof github.com/google/pprof/driver github.com/google/pprof/fuzz github.com/google/pprof/internal/binutils github.com/google/pprof/internal/driver github.com/google/pprof/internal/elfexec github.com/google/pprof/internal/graph github.com/google/pprof/internal/measurement github.com/google/pprof/internal/plugin github.com/google/pprof/internal/proftest github.com/google/pprof/internal/report github.com/google/pprof/internal/symbolizer github.com/google/pprof/internal/symbolz github.com/google/pprof/internal/transport github.com/google/pprof/profile github.com/google/pprof/third_party/d3 github.com/google/pprof/third_party/d3flamegraph github.com/google/pprof/third_party/svgpan
einternal/driver/svg.go:39:12: assignment to nil map (SA5000)

palpant-mbp:pprof palpant$ echo $?
1
```

I can push it to verify on Travis and then undo, if you'd like.

aalexand: One thing that I am unclear about is why SA6004 had to be disabled for some files before, but not now. Any theory for that?

timpalpant: It looks like that check was noisy and removed from the staticcheck tool in: https://github.com/dominikh/go-tools/commit/dc1aef752582c5c22b80d0dc3b4f2180155f37f8

It no longer appears in the [documentation](https://staticcheck.io/docs/checks) or in the registry of checks in the code: https://github.com/dominikh/go-tools/blob/9e9c9d3692a44f108ece3d6978d347e5a6c29ada/staticcheck/analysis.go#L278-L285

I'm not sure exactly what it tried to validate, but it was called `CheckSillyRegex`, and based on the previous list of exclusions here, I would guess it was incorrectly firing on regexes in the pprof code.

aalexand: I see, thanks!

PR #481, opened 2019-08-14
Continuous Fuzzing Integration via Fuzzit

yevgenypats:
Hi Team,

I'm Yevgeny (Founder of [Fuzzit](https://fuzzit.dev))

This PR adds a continuous fuzzing integration to pprof travis pipeline via [Fuzzit](https://fuzzit.dev) service.

This means the following:

- Every time new code is pushed to master new fuzzers are built and push to fuzzit where they run continuously generate new test-cases and alerts if new crashes are found
- Every pull-request the fuzzers runs through a quick regression tests with the generated test-cases from the previous step. If something new/old bugs are introduced you will see this immediately in the Travis check.

To take ownership of the organisation, please login to https://app.fuzzit.dev and let me know your account.

Please review and feel free to comment/ask questions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ matrix:
       go: 1.12.x
     - os: linux
       go: master
+      env: WITH_FUZZIT=1
     - os: osx
       osx_image: xcode8.3
       go: 1.11.x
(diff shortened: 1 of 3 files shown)


yevgenypats:
@kjk I think you need to sign the CLA as you did the last commit (I already signed).

CCing @nolanmar511 as you worked on the fuzz targets.

aalexand:
@yevgenypats Thanks for the PR - we did some fuzzing in the past, we are done for now. We'll consider the integration in the future.

yevgenypats:
@aalexand no worries. ping me again when you are interested. Just interesting - what do you mean  by: "we did some fuzzing in the past, we are done for now"? I mean if you already have the fuzz target written why not run them continuously find new bugs and prevent from old bugs propagating again? (Not trying to pursue you using the service it's free anyway. I'm just trying to understand the use-case)

aalexand:
We did some fuzzing in https://github.com/google/pprof/tree/master/fuzz, there is some corpus there. The bugs we found there were quite trivial, much more trivial than expected. The reason is that the input to the API being fuzzed is a protobuf, I believe those cannot be fuzzed very well without specialized mutators like https://github.com/google/libprotobuf-mutator, but that doesn't exist for Go. So, before enabling any automated tooling, I'd like us or someone spend a bit more time proving that fuzzing is useful for pprof. Without that I'd prefer not to enable yet any continuous integration - any tool has its cost of ownership so I'd like to know benefits outweigh future time spending.

yevgenypats:
@aalexand Thank you very much for the detailed explanation. That makes a lot of sense, I think people are working in that direction with go-fuzz. I'll keep an eye on that as well.

PR #484, opened 2019-09-24
webhtml: fix search input svg icon

narqo:
I have noticed, that in the web ui, the icon of magnifier glass of the search input isn't rendered. Of all browsers that I've checked Safari is the only one that renders the icon, while in Firefox and Chrome the icon is missing:

Screenshot 2019-09-24 at 12 00 43

This PR fixes, what I believe was a typo, by escaping the hex value of icon's fill colour.

**Update** After the fix:
Screenshot 2019-09-24 at 20 16 27

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -14,10 +14,12 @@
 
 package driver
 
-import "html/template"
+import (
+	"html/template"
 
-import "github.com/google/pprof/third_party/d3"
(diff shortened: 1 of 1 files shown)


aalexand:
@narqo Could you add a screenshot after the fix?

narqo:
Updated the description, PHAL

narqo:
@aalexand Anything also I can do to move the PR forward?

aalexand:
@narqo merged

PR #488, opened 2019-10-14
Added capacity to slice creation, when capacity is known

vivekv96:
Simple optimisation to reduce memory allocations and copying(while appending).

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -386,7 +386,7 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	symNodes := nodesPerSymbol(g.Nodes, symbols)
 
 	// Sort for printing.
-	var syms []*objSymbol
+	syms := make([]*objSymbol, 0, len(symNodes))
 	for s := range symNodes {
 		syms = append(syms, s)
 	}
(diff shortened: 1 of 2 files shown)


vivekv96:
@googlebot I signed it!

internal/report/report.go
```diff
@@ -386,7 +386,7 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	symNodes := nodesPerSymbol(g.Nodes, symbols)
 
 	// Sort for printing.
-	var syms []*objSymbol
+	syms := make([]*objSymbol, 0, len(symNodes))
```
nolanmar511: I believe using creates an array of the specified size filled with empty values, then appending will add a non-empty value to the end (rather filling in an empty spot).
For example
```
arr := make([]int, 2)
arr = append(arr, 1)
```
produces the array [0, 0, 1].

vivekv96: @nolanmar511, regarding the example that you have provided, yes that would produce [0, 0, 1].
But, the make() that I have used has length=0 and capacity=len(symNodes), so it's different from giving it a length of 2.

So, the equivalent example = 
```
arr := make([]int, 0, 5)
arr = append(arr, 1)
```

which produces the array [1]. Playground link: https://play.golang.org/p/nSUHnfqubiJ

aalexand:
Also, does this change solve a specific performance problem? If not, the
code is fine as is, let's not optimize prematurely.

vivekv96:
@aalexand No, I came across this while just glancing over the code, haven't faced any performance issues per se. 
While I agree with you that this is a premature optimisation of sorts, and the code works as is, this does reduce allocs/op and the other metrics.
If you believe that this is unnecessary, go ahead and close this PR.
Thanks. No worries.

PR #491, opened 2019-10-21
webhtml: reenable sort option for flamegraphs

bbrks:
This sorts flamegraph elements lexicographically such that two similar profiles can be compared side-by-side to eyeball differences in flamegraph profile shapes and sizes.

Flamegraph elements prior to #367 were sorted, but since then, it has been difficult to eyeball 2 similar profiles side-by-side, as the elements appear in a somewhat random order.

## Before sorting profile comparison:
Screenshot 2019-10-21 at 22 50 25

## After sorting profile comparison:
Screenshot 2019-10-21 at 22 50 18

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -1094,6 +1094,7 @@ function viewer(baseUrl, nodes) {
       .transitionDuration(750)
       .transitionEase(d3.easeCubic)
       .inverted(true)
+      .sort(true)
       .title('')
       .tooltip(false)
       .details(document.getElementById('flamegraphdetails'));


nolanmar511:
I know there are differing opinions for whether the flame graph should be sorted by total value for each node, or whether it should be sorted by function name (mainly depending on the use case), but it looks like the order of nodes now is random. This change strikes me as positive (especially given that profile comparison is not supported in pprof's flame graph view). 

aalexand@ or kchadalavada@ -- This change LGTM, but I'll let either of you decide if this should be approved and merged,

aalexand:
@nolanmar511 Yes, I think the change is an improvement. I am not sure whether this implementation of the flame graph is capable of sorting by the total value, but whatever sorting is better than no sorting.

Something is up with the CI tests though: looks like Travis CI is going to fail with

```
+++staticcheck github.com/google/pprof github.com/google/pprof/driver github.com/google/pprof/fuzz github.com/google/pprof/internal/binutils github.com/google/pprof/internal/driver github.com/google/pprof/internal/elfexec github.com/google/pprof/internal/graph github.com/google/pprof/internal/measurement github.com/google/pprof/internal/plugin github.com/google/pprof/internal/proftest github.com/google/pprof/internal/report github.com/google/pprof/internal/symbolizer github.com/google/pprof/internal/symbolz github.com/google/pprof/internal/transport github.com/google/pprof/profile github.com/google/pprof/third_party/d3 github.com/google/pprof/third_party/d3flamegraph github.com/google/pprof/third_party/svgpan
internal/driver/interactive_test.go:308:42: possible nil pointer dereference (SA5011)
The command "./test.sh" exited with 1.
```

which we need to fix and looks like AppVeyor build didn't start at all.

PR #498, opened 2019-11-04
template functions should not be simplified

nolanmar511:


internal/graph/graph.go
```diff
@@ -429,11 +430,13 @@ func newTree(prof *profile.Profile, o *Options) (g *Graph) {
 
 // ShortenFunctionName returns a shortened version of a function's name.
 func ShortenFunctionName(f string) string {
+	f = cppAnonymousPrefixRegExp.ReplaceAllString(f, "")
 	for _, re := range []*regexp.Regexp{goRegExp, javaRegExp, cppRegExp} {
 		if matches := re.FindStringSubmatch(f); len(matches) >= 2 {
 			return strings.Join(matches[1:], "")
 		}
 	}
+
```
kalyanac: empty line intended?

nolanmar511: Done.

aalexand:
@nolanmar511 It would be good to explain the motivation / rationale for such changes in the PR description.

internal/graph/graph_test.go
```diff
@@ -475,6 +475,10 @@ func TestShortenFunctionName(t *testing.T) {
 			"foo_bar::Foo::bar",
 			"Foo::bar",
 		},
+		{
+			"Foo::bar::baz<float, long, int>::operator()",
```
aalexand: This example is unclear.

1. What is "Foo::bar::baz"? I would expect that "Foo::bar" is a namespace and "baz" is a method? But namespaces are traditionally lowercase-named in C++, so "Foo" doesn't look like a namespace, but then what is it? The other test case above makes that clear by saying "cpp::namespace::Class::method" so that it's clear what element is what, but the added test case is unclear on that.

2. In other C++ test cases here the simplification seems to be around stripping the namespace. Why the namespace is not stripped here?

In other words, I could see a test case like

```
"cpp::namespace::Class::method<float, long, int>()",
"Class::method<float, long, int>()",
```

but the current one I don't quite get.

PR #504, opened 2019-12-16
Apply the --divide_by option to the profile before saving it.

rauls5382:
This allows averaging profiles by using:
pprof -proto output --divide_by <n> profile1.pb.gz ... profilen.pb.gz


aalexand:
@rauls5382 A test would be good, but I can merge as is as well - please LMK.

rauls5382:
There are currently no tests for the proto output. I'd rather add some tests on a separate PR.

PR #507, opened 2020-02-09
Add csv format

aalexand:
This commit adds support for outputting the `top` flat output format as
a CSV. This is useful when the output is intended to be parsed by
machine, such as when pprof is used as part of a test suite.

The values are still printed as formatted values. This is arguably not
the best thing to do, but printing them as unformatted values would be
worse since those values come straight from the profile and can be very
hard to understand without knowing lower-level details of how the
profile is collected (e.g. CPU profile sample values can be in
nanoseconds or they can be in microseconds, I have seen both).

The commit also changes slightly the behavior of measurement.Percentage
function - it won't add whitespace on the left anymore. I think this
change is the right thing to do as whether the whitespace is needed
depends on the context where the value is used.


aalexand:
@rauls5382 I did not consider topproto, it's probably enough for what I need.

PR #517, opened 2020-04-03
profile.CheckValid should return an error if location has a line with empty function

Labutin:
CheckValid should check that all lines in locations have non nil function pointer.

### DIFF
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -402,6 +402,8 @@ func (p *Profile) CheckValid() error {
 				if f.ID == 0 || functions[f.ID] != f {
 					return fmt.Errorf("inconsistent function %p: %d", f, f.ID)
 				}
+			} else {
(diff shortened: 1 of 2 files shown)


profile/profile.go
```diff
@@ -402,6 +402,8 @@ func (p *Profile) CheckValid() error {
 				if f.ID == 0 || functions[f.ID] != f {
 					return fmt.Errorf("inconsistent function %p: %d", f, f.ID)
 				}
+			} else {
+				return fmt.Errorf("profile has nil function in location line")
```
aalexand, line 406: Can we include any context in the error message that will help triage the error?

aalexand, line 406: Do the nil check first and unindent the main path.

Labutin: Done

Labutin: Done

Issue #400, opened 2018-07-17
Support Intel syntax in the assembly reports


aalexand:
A user asked for this. Filing so that other users can upvote.

There is a workaround of creating a script, say ~/bin/objdump with content

```
#!/bin/bash
/usr/bin/objdump -M intel $*
```

and then setting

```
export PPROF_TOOLS=objdump:/home/user/bin/
```

but this is somewhat cumbersome.

wyk9787:
Should this be added as an independent option that only applies to "-disasm" and "-weblist" or as a parameter taken by those two commands?

aalexand:
Independent option I think. It's not special / important enough to be a dedicated argument for those commands.

PR #520, opened 2020-04-10
Support Intel syntax in the assembly reports

wyk9787:
Fixes: #400

Add an option to support Intel syntax in the assembly report, which is used in both `disasm` and `weblist` formats.

Tested on both Mac OS and Linux.

```
(pprof) disasm absl::*
   ...
         .          .    28aa620: movl  %ebx, %r11d
   ...
(pprof) intel_syntax
(pprof) disasm absl::*
   ...
         .          .    28aa620: mov   r11d, ebx
   ...

```

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -142,7 +142,7 @@ type ObjTool interface {
 
 	// Disasm disassembles the named object file, starting at
 	// the start address and stopping at (before) the end address.
-	Disasm(file string, start, end uint64) ([]Inst, error)
+	Disasm(file string, start, end uint64, intelSyntax bool) ([]Inst, error)
 }
 
 // An Inst is a single instruction in an assembly listing.
@@ -269,8 +269,8 @@ func (f *internalObjFile) Symbols(r *regexp.Regexp, addr uint64) ([]*plugin.Sym,
 	return pluginSyms, nil
 }
 
-func (o *internalObjTool) Disasm(file string, start, end uint64) ([]plugin.Inst, error) {
-	insts, err := o.ObjTool.Disasm(file, start, end)
+func (o *internalObjTool) Disasm(file string, start, end uint64, intelSyntax bool) ([]plugin.Inst, error) {
+	insts, err := o.ObjTool.Disasm(file, start, end, intelSyntax)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/binutils/binutils.go
+++ b/internal/binutils/binutils.go
@@ -160,24 +160,19 @@ func findExe(cmd string, paths []string) (string, bool) {
 func (bu *Binutils) Disasm(file string, start, end uint64, intelSyntax bool) ([]plugin.Inst, error) {
 	b := bu.get()
 	var cmd *exec.Cmd
+	args := []string{"-d", "-C", "--no-show-raw-insn", "-l",
+		fmt.Sprintf("--start-address=%#x", start),
+		fmt.Sprintf("--stop-address=%#x", end),
+		file}
+
 	if intelSyntax {
 		if runtime.GOOS == "darwin" {
-			cmd = exec.Command(b.objdump, "-x86-asm-syntax=intel", "-d", "-C", "--no-show-raw-insn", "-l",
-				fmt.Sprintf("--start-address=%#x", start),
-				fmt.Sprintf("--stop-address=%#x", end),
-				file)
+			args = append([]string{"-x86-asm-syntax=intel"}, args...)
 		} else {
-			cmd = exec.Command(b.objdump, "-M", "intel", "-d", "-C", "--no-show-raw-insn", "-l",
-				fmt.Sprintf("--start-address=%#x", start),
-				fmt.Sprintf("--stop-address=%#x", end),
-				file)
+			args = append([]string{"-M", "intel"}, args...)
 		}
-	} else {
-		cmd = exec.Command(b.objdump, "-d", "-C", "--no-show-raw-insn", "-l",
-			fmt.Sprintf("--start-address=%#x", start),
-			fmt.Sprintf("--stop-address=%#x", end),
-			file)
 	}
+	cmd = exec.Command(b.objdump, args...)
 	out, err := cmd.Output()
 	if err != nil {
 		return nil, fmt.Errorf("%v: %v", cmd.Args, err)
--- a/internal/binutils/binutils.go
+++ b/internal/binutils/binutils.go
@@ -157,12 +157,27 @@ func findExe(cmd string, paths []string) (string, bool) {
 
 // Disasm returns the assembly instructions for the specified address range
 // of a binary.
-func (bu *Binutils) Disasm(file string, start, end uint64) ([]plugin.Inst, error) {
+func (bu *Binutils) Disasm(file string, start, end uint64, intelSyntax bool) ([]plugin.Inst, error) {
 	b := bu.get()
-	cmd := exec.Command(b.objdump, "-d", "-C", "--no-show-raw-insn", "-l",
-		fmt.Sprintf("--start-address=%#x", start),
-		fmt.Sprintf("--stop-address=%#x", end),
-		file)
+	var cmd *exec.Cmd
+	if intelSyntax {
+		if runtime.GOOS == "darwin" {
+			cmd = exec.Command(b.objdump, "-x86-asm-syntax=intel", "-d", "-C", "--no-show-raw-insn", "-l",
+				fmt.Sprintf("--start-address=%#x", start),
+				fmt.Sprintf("--stop-address=%#x", end),
+				file)
+		} else {
+			cmd = exec.Command(b.objdump, "-M", "intel", "-d", "-C", "--no-show-raw-insn", "-l",
+				fmt.Sprintf("--start-address=%#x", start),
+				fmt.Sprintf("--stop-address=%#x", end),
+				file)
+		}
+	} else {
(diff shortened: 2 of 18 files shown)


kalyanac:
Can you include test outputs?

internal/binutils/binutils_test.go
```diff
@@ -188,10 +188,9 @@ func skipUnlessDarwinAmd64(t *testing.T) {
 	}
 }
 
-func TestDisasm(t *testing.T) {
-	skipUnlessLinuxAmd64(t)
+func DisasmTestingInternal(t *testing.T, intelSyntax bool) {
 	bu := &Binutils{}
-	insts, err := bu.Disasm(filepath.Join("testdata", "exe_linux_64"), 0, math.MaxUint64)
+	insts, err := bu.Disasm(filepath.Join("testdata", "exe_linux_64"), 0, math.MaxUint64, false)
```
kalyanac, line 193: 'false' should be intelSyntax

internal/driver/commands.go
```diff
@@ -154,6 +154,9 @@ var pprofVariables = variables{
 	"compact_labels": &variable{boolKind, "f", "", "Show minimal headers"},
 	"source_path":    &variable{stringKind, "", "", "Search path for source files"},
 	"trim_path":      &variable{stringKind, "", "", "Path to trim from source paths before search"},
+	"intel_syntax": &variable{boolKind, "f", "", helpText(
+		"Show assembly in Intel syntax",
+		"Only useful for command `disasm` and `weblist`")},
```
kalyanac, line 159: Only applicable to `disasm` and `weblist`

internal/binutils/binutils_test.go
wyk9787: Do you mean I should switch the behavior for `Disasm`, i.e. passing `false` to `Disasm` to indicate users want Intel syntax?

driver/driver.go
```diff
@@ -142,7 +142,7 @@ type ObjTool interface {
 
 	// Disasm disassembles the named object file, starting at
 	// the start address and stopping at (before) the end address.
-	Disasm(file string, start, end uint64) ([]Inst, error)
+	Disasm(file string, start, end uint64, intelSyntax bool) ([]Inst, error)
```
kalyanac, line 145: I wonder if accepting string is better to accommodate future changes in supported syntax. ex: asm_syntax=intel

internal/binutils/binutils_test.go
```diff
@@ -188,10 +188,9 @@ func skipUnlessDarwinAmd64(t *testing.T) {
 	}
 }
 
-func TestDisasm(t *testing.T) {
-	skipUnlessLinuxAmd64(t)
+func DisasmTestingInternal(t *testing.T, intelSyntax bool) {
```
aalexand, line 191: disasmTesting

aalexand: Actually, testDisasm

internal/binutils/binutils.go
```diff
@@ -157,12 +157,27 @@ func findExe(cmd string, paths []string) (string, bool) {
...
+				fmt.Sprintf("--start-address=%#x", start),
+				fmt.Sprintf("--stop-address=%#x", end),
+				file)
+		}
+	} else {
+		cmd = exec.Command(b.objdump, "-d", "-C", "--no-show-raw-insn", "-l",
+			fmt.Sprintf("--start-address=%#x", start),
+			fmt.Sprintf("--stop-address=%#x", end),
+			file)
+	}
```
aalexand, line 180: Can we modify the arguments instead of duplicating the whole call? Make them an array, massage as needed, use ellipsis syntax to invoke exec.Command.

internal/driver/commands.go
```diff
@@ -215,7 +218,7 @@ var pprofVariables = variables{
 		"For memory profiles, report average memory per allocation.",
 		"For time-based profiles, report average time per event.")},
 	"sample_index": &variable{stringKind, "", "", helpText(
-		"Sample value to report (0-based index or name)",
+		"Sample value to report (<x> is 0-based index or name)",
```
aalexand, line 221: This doesn't seem related to this change.

internal/report/source.go
```diff
@@ -239,7 +239,8 @@ func sourceCoordinates(asm map[int][]assemblyInstruction) (start, end int) {
 // assemblyPerSourceLine disassembles the binary containing a symbol
 // and classifies the assembly instructions according to its
 // corresponding source line, annotating them with a set of samples.
-func assemblyPerSourceLine(objSyms []*objSymbol, rs graph.Nodes, src string, obj plugin.ObjTool) map[int][]assemblyInstruction {
+func assemblyPerSourceLine(objSyms []*objSymbol, rs graph.Nodes,
+	src string, obj plugin.ObjTool, intelSyntax bool) map[int][]assemblyInstruction {
```
aalexand, line 243: Please avoid unrelated formatting changes. Keep this one line, long lines are fine in Go.

aalexand:
This PR seems to contain unrelated commits.

driver/driver.go
aalexand: I wouldn't do it - this makes code harder to refactor. We can revise the current approach when we have more options. E.g. this could take an option structure of something. But just an argument is fine for now.

internal/driver/commands.go
wyk9787: Done.

driver/driver.go
wyk9787: Ack.

internal/binutils/binutils_test.go
wyk9787: Done.

internal/binutils/binutils.go
wyk9787: Done.

internal/driver/commands.go
wyk9787: Reverted.

internal/report/source.go
wyk9787: Ack. Now fixed.

wyk9787:
> This PR seems to contain unrelated commits.

Should now be reverted.

internal/report/report.go
```diff
@@ -79,6 +79,8 @@ type Options struct {
 	Symbol     *regexp.Regexp // Symbols to include on disassembly report.
 	SourcePath string         // Search path for source files.
 	TrimPath   string         // Paths to trim from source file paths.
+
+	IntelSyntax bool // Whether or not to print assembly in Intel syntax
```
aalexand, line 83: nit: missing trailing full stop.

wyk9787: Done.

wyk9787:
> to commands

Done.

internal/binutils/binutils_test.go
wyk9787: Now fixed.

aalexand:
> Are you referring to "--start_address" and "--stop-address"? Do you mean they should be put after `file`?

I am referring to the file path argument (it's the only positional one: `--start-address` and `--stop-address` are not, these are options / flags).

wyk9787:
> > Are you referring to "--start_address" and "--stop-address"? Do you mean they should be put after `file`?
> 
> I am referring to the file path argument (it's the only positional one: `--start-address` and `--stop-address` are not, these are options / flags).

I see. Right now, I am actually inserting the "intel flags" at the beginning of the `args`, so the file path argument is indeed the last argument.

aalexand:
> I see. Right now, I am actually inserting the "intel flags" at the beginning of the `args`, so the file path argument is indeed the last argument.

Ah, I missed that, sorry.

internal/binutils/binutils.go
```diff
@@ -157,12 +157,22 @@ func findExe(cmd string, paths []string) (string, bool) {
 
 // Disasm returns the assembly instructions for the specified address range
 // of a binary.
-func (bu *Binutils) Disasm(file string, start, end uint64) ([]plugin.Inst, error) {
+func (bu *Binutils) Disasm(file string, start, end uint64, intelSyntax bool) ([]plugin.Inst, error) {
 	b := bu.get()
-	cmd := exec.Command(b.objdump, "-d", "-C", "--no-show-raw-insn", "-l",
+	var cmd *exec.Cmd
```
aalexand, line 162: Remove this var, and use "cmd :=" below, as prior to to this change.

internal/binutils/binutils.go
```diff
@@ -157,12 +157,22 @@ func findExe(cmd string, paths []string) (string, bool) {
...
+	args := []string{"-d", "-C", "--no-show-raw-insn", "-l",
 		fmt.Sprintf("--start-address=%#x", start),
 		fmt.Sprintf("--stop-address=%#x", end),
-		file)
+		file}
+
+	if intelSyntax {
+		if runtime.GOOS == "darwin" {
+			args = append([]string{"-x86-asm-syntax=intel"}, args...)
```
aalexand, line 170: nit: I personally would append this flag at the end, and move appending file to after this code. It doesn't really matter much in this case, but appending "args..." is O(N), unnecessarily. Again, this doesn't really matter here since the list is fixed len and tiny. So feel free to leave this as is.

internal/binutils/binutils.go
wyk9787: Done.

internal/binutils/binutils.go
wyk9787: It makes sense. Now changed.

PR #521, opened 2020-04-15
Filter out null characters in tags.

wyk9787:
Null characters in tags will lead to null characters in labels, and `dot` is not able to parse such labels.

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -686,6 +686,8 @@ func defaultLabelFormat(v int64, key string) string {
 func (m TagMap) findOrAddTag(label, unit string, value int64) *Tag {
 	l := m[label]
 	if l == nil {
+		// Filter out all null characters since dot is not able to parse them
+		label = strings.ReplaceAll(label, "\000", "")
 		l = &Tag{
 			Name:  label,
 			Unit:  unit,


internal/graph/graph.go
```diff
@@ -686,6 +686,8 @@ func defaultLabelFormat(v int64, key string) string {
 func (m TagMap) findOrAddTag(label, unit string, value int64) *Tag {
 	l := m[label]
 	if l == nil {
+		// Filter out all null characters since dot is not able to parse them
```
aalexand, line 689: nit: missing trailing dot.

internal/graph/graph.go
```diff
@@ -686,6 +686,8 @@ func defaultLabelFormat(v int64, key string) string {
...
+		label = strings.ReplaceAll(label, "\000", "")
```
aalexand, line 690: Is it possible to add a test case?

It would also be good to mention why this can happen. Also, are null characters the only ones that need to be filtered out?

aalexand: I also think escaping rather than filtering out these characters would be a more correct approach.

wyk9787:
Closing this PR as there is no need to address this issue in pprof.

PR #522, opened 2020-04-16
Remove support for go 1.12 and add support for go 1.14 in Travis.

wyk9787:
Since the release of go 1.14, go 1.12 is not supported any more: https://golang.org/doc/devel/release.html#policy.


wyk9787:
> Could you clarify in the PR title that this removes 1.12 support and adds 1.14?

Done.

Issue #526, opened 2020-04-22
internal/driver fails at Go tip due to changes in pprof


andybons:
See https://travis-ci.org/github/google/pprof/jobs/678277764 for an example failure.

This is due to https://go.googlesource.com/go/+/2ff1e3e, specifically the behavior change mentioned in the commit message:

> Providing the "seconds" parameter to other types of profiles is an error.

@hyangah

hyangah:
Why is the test querying the "/debug/pprof/goroutine" endpoint with "seconds" parameter?
Can the test be fixed?

hyangah:
Actually, I am thinking to fix from go to expand the list of endpoints that support "seconds" param.

nolanmar511:
I think this test could also be changed, if that's the action we should take. 

@wyk9787 -- Could you take a look?

wyk9787:
> I think this test could also be changed, if that's the action we should take.
> 
> @wyk9787 -- Could you take a look?

Will do.

andybons:
Thanks for the quick responses, everyone.

hyangah:
FYI - https://go-review.googlesource.com/c/go/+/229537 is merged. Not sure if the current test will pass though because the 'seconds' parameter is now not ignored.

PR #527, opened 2020-04-22
Remove the "seconds" parameter to fix tests.

wyk9787:
Fixes: #526

Due to https://go.googlesource.com/go/+/2ff1e3e, the behavior changes to
> Providing the "seconds" parameter to other types of profiles is an error.


nolanmar511:
It looks like this is being fixed on the Golang side; we may want to merge this, but plan to revert it once the golang fix lands sufficiently.

nolanmar511:
@kalyanac -- Thoughts?

Should we merge this fix permanently, temporarily, or not at all given that there is a fix in golang?

aalexand:
What is the fix on the Golang side? The goroutine request is a point request, it doesn't need the "seconds" parameter. It should have been removed in https://github.com/google/pprof/pull/350/files, but it wasn't. So this PR seems correct regardless.

Issue #493, opened 2019-10-24
[doc] how to read pprof dot graph?


AlmogBaku:
How to read the graph? Explanations and legend of the signs in the graph can be very useful!

I.e. what is dotted line? what is dashed line? what are the number means? what are the boxes(n the end of the tree) means?

Also- improving the labels/captions(title when hovering a dom item) can be very useful

nolanmar511:
Having an in-graph legend would definitely be useful!!

Just for reference, I'm going to note some of the features which we might consider documenting in a legend here:
* Dotted/dashed lines indicated that that intervening nodes have been removed. Nodes are removed to keep graphs small enough for visualization.
* The wider an arrow the more of the metric being measured is used along that path.
* Nodes and edges are colored according to their total value (total being the amount of a metric used by a node and all of its children); large positive values are red; large negative values are green.

And some of the text which could be better documented:
* Within each Node, there is text reading "1024kB (5%) of 2048kB(10%)". That indicates that 1024kB were allocated in that location (a function or line number), 2048kB were allocated by that location and all functions it calls, and that the 1024kB allocated in that location account for 5% of all allocations in the profile and the 2048kB allocated by that location and all of its children accoount for 10% of all allocations in the profile.
* The numbers next to the arrows indicated how much of a metric is used going down that path.

And, in heap profiles, there are those cube-like boxes coming from leaf nodes, like the following:
![image](https://user-images.githubusercontent.com/14046562/67496741-c0346380-f631-11e9-96f8-e06f1f48916a.png)
Those show the size of the allocation (with the number next to the arrow indicated how much memory is allocated in allocations of that size.

Please point out any features which I missed!

nolanmar511:
I'm thinking that we maybe should add a legend which shows the meaning of node and line colors, line widths, and dashed/solid lines. 

Other features that I mentioned, we've thought about adding a graph with labels showing the meaning of all features to our documentation. I hope that might help.

AlmogBaku:
In the meanwhile, maybe you can explain it here and let's add a ref in the docs?

nolanmar511:
@AlmogBaku -- I'd added an explanation of how to read the graph view in https://github.com/google/pprof/issues/493#issuecomment-545952479. Is any information missing there?

erdemtuna:
Hello @nolanmar511, how do we interpret the two images below? I could not figure it out following your explanation.

[...]

PR #530, opened 2020-04-30
Add url to "Interpret the call graph" section in pprof legend.

wyk9787:
Fixes: #493

Example:
Screen Shot 2020-04-29 at 6 10 06 PM


internal/report/report.go
```diff
@@ -1165,7 +1165,7 @@ func ProfileLabels(rpt *Report) []string {
 
 // reportLabels returns printable labels for a report. Includes
 // profileLabels.
-func reportLabels(rpt *Report, g *graph.Graph, origCount, droppedNodes, droppedEdges int, fullHeaders bool) []string {
+func reportLabels(rpt *Report, g *graph.Graph, origCount, droppedNodes, droppedEdges int, fullHeaders bool, helperURL bool) []string {
```
nolanmar511, line 1168: It looks like fullHeaders and helperURL always have the same value. And it makes sense to me that the helper URL might be required only when full headers are required. 

Perhaps no argument needs to be added to this function, and fullHeaders can be used in place of helperURL?

wyk9787: Done.

wyk9787:
Done with changes.

PR #531, opened 2020-04-30
Update README to reflect what green nodes/edges mean.

wyk9787:


doc/README.md
```diff
@@ -213,6 +213,10 @@ the graph, controlled by the *nodecount* option.
 
 * **"(inline)" Edge Marker**: the call has been inlined into the caller.
 
+* **Green Nodes/Edges**: when [comparing two profiles](#comparing-profiles), the
```
nolanmar511, line 216: I think a larger change should be considered (basically, everywhere "red" is mentioned, "green" should also be mentioned). To minimize the number of top-level bullets, it may also make sense to replace the "Thicker & Redder Edges" and "Thinner & Greyer Edges" bullets with a "Edge color and weight bullet".

Also note that edge weight and font size both indicate a large absolute value.

wyk9787: Should I have a separate bullet point for "comparing two profiles"? It will be difficult to describe those properties combining both the normal situation and the "comparison mode". For example, a red edge means larger value in either situation, but "more resources were used along that path" only applies to the normal situation and does not apply to the "comparison mode".

nolanmar511: Comparison is one way in which negative values can appear in the profiles which we typically work with. But pprof is a general tool; there's no reason that profiles with negative values cannot be viewed with the pprof CLI, so green can show up when one is not using comparison mode if the input profile contains negative values.

wyk9787: Done.

wyk9787:
Updated top-level bullets to always describe the properties of nodes and edges.

doc/README.md
```diff
@@ -190,27 +190,29 @@ the graph, controlled by the *nodecount* option.
...
 ### Interpreting the Callgraph
 
-* **Flat Value**: the value of a location itself. This is indicated by the font 
-  size.
-    *   Nodes with larger font sizes have larger flat values.
-    *   Nodes with smaller font sizes have smaller flat values.
+* **Node Color**:
+  * large positive cum values are red.
+  * large negative cum values are green.
+  * values close to zero are grey.
```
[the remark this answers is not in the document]
wyk9787: Done.

doc/README.md
```diff
@@ -190,27 +190,29 @@ the graph, controlled by the *nodecount* option.
...
 
-* **Cum Value**: the value of a location plus all of its descendants. This is
-  indicated by the node's color.
-    *   Redder nodes have greater cum values.
-    *   Greyer nodes have smaller cum values.
+* **Node Font Size**:
+  * larger font size means larger flat values.
+  * smaller font size means smaller flat values.
```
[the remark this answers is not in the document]
wyk9787: Done.

wyk9787:
Done with changes.

PR #532, opened 2020-05-05
Fixed demangling fail on mac.

mbbill:
the demangle library doesn't like the double underscore names. This patch will trim one underscore to make it behave the same as mac's c++filt tool.

### DIFF
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -223,7 +223,9 @@ func Demangle(prof *profile.Profile, force bool, demanglerMode string) {
 		if fn.Name != "" && fn.SystemName != fn.Name {
 			continue // Already demangled.
 		}
-		copy(o, options)
+		if(strings.HasPrefix(fn.SystemName, "__")) {
+			fn.SystemName = fn.SystemName[1:]
+		}
 		if demangled := demangle.Filter(fn.SystemName, o...); demangled != fn.SystemName {
 			fn.Name = demangled
 			continue


internal/symbolizer/symbolizer.go
```diff
@@ -223,7 +223,9 @@ func Demangle(prof *profile.Profile, force bool, demanglerMode string) {
 		if fn.Name != "" && fn.SystemName != fn.Name {
 			continue // Already demangled.
 		}
-		copy(o, options)
+		if(strings.HasPrefix(fn.SystemName, "__")) {
```
aalexand, line 226: I am not sure this is a safe change. An already-demangled symbol could legitimately start with double-underscore and this will modify that name which is not expected.

Please also update the PR description with specific example of where the problem occurs, including a reproducer if possible. Please also add a test that fails before the fix and passes after.

mbbill: I agree it's a bit hack and I hope you guys knows how to fix it in a better way. I actually find a previous PR also trying to fix the same issue but that PR has been sitting there for more than a year.
I guess nobody is using this on Mac since we don't need a repro I believe, it's just won't work. I'm fine if you guys willing to keep this broken for longer. On my side even a hacky fix is better then keep this broken for so long.

mbbill:
@googlebot I signed it!

internal/symbolizer/symbolizer.go
kalyanac: I will revive #449 and try to get it submitted. That should address thie need.

aalexand: I'm not opposed to this PR either (it's smaller than #449 so should be easier to review), but there are no tests and no specific examples in the PR description and I'd need those to make sure the PR does what (and only what) it intends to do.

mbbill: agreed. I'll add description and tests. I'm also thinking maybe it's better land this path on `github.com/ianlancetaylor/demangle`, let's see what I can do here.

mbbill: @kalyanac if your patch can get merged that would be great! looking forward for that!

nolanmar511:
Based on comments, it seems that the plan was for this PR to be superseded by https://github.com/google/pprof/pull/449. I'm closing this PR (and following up on #449).

PR #534, opened 2020-05-06
Add support for LLVM objdump

kalyanac:
Adds support for LLVM objdump


kalyanac:
Failures unrelated. Restarted the jobs

kalyanac:
gentle ping.

aalexand:
It feels that the PR description doesn't tell the full story. There is no binutils version 9.0, that seems LLVM objdump. It seems that what we are trying to dispatch on here is binutils vs. LLVM objdump, so the PR description and the code change should be centered around that IMO.

internal/binutils/binutils.go
```diff
@@ -120,6 +121,29 @@ func (bu *Binutils) SetTools(config string) {
...
+		return objdump, false
+	}
+
+	re := regexp.MustCompile(`.*version (([0-9]*)\.([0-9]*)\.([0-9]*)).*$`)
+	fields := re.FindStringSubmatch(strings.Split(string(cmdOut), "\n")[0])
+	if len(fields) != 5 {
+		return objdump, false
+	}
+
+	if ver, _ := strconv.Atoi(fields[2]); ver < 9 {
```
aalexand, line 141: Don't ignore errors.

internal/binutils/binutils.go
```diff
@@ -120,6 +121,29 @@ func (bu *Binutils) SetTools(config string) {
 	bu.update(func(r *binrep) { initTools(r, config) })
 }
 
+func ensureObjdumpVersion(objdump string, objdumpFound bool) (string, bool) {
```
aalexand, line 124: Prefer top-down function order in Go - place this function after its use, i.e. after initTools.

internal/binutils/binutils.go
```diff
@@ -159,10 +183,19 @@ func findExe(cmd string, paths []string) (string, bool) {
...
 	b := bu.get()
+	if !b.objdumpFound {
+		return nil, fmt.Errorf("couldn't find objdump")
+	}
 	args := []string{"-d", "-C", "--no-show-raw-insn", "-l",
 		fmt.Sprintf("--start-address=%#x", start),
 		fmt.Sprintf("--stop-address=%#x", end)}
 
+	if runtime.GOOS == "darwin" {
+		args = []string{"-disassemble-all", "-no-show-raw-insn",
```
aalexand, line 194: Avoid re-assignment, this makes code more difficult to understand. Initialize once:
```
var foo ...
if (...) {
  foo = ...
} else {
  foo = ...
}
```

internal/binutils/binutils_test.go
```diff
@@ -189,14 +191,39 @@ func skipUnlessDarwinAmd64(t *testing.T) {
 }
 
 func testDisasm(t *testing.T, intelSyntax bool) {
+	if runtime.GOOS == "darwin" {
+		// get objdump version and skip test if < 9.0
```
aalexand, line 195: nit: get -> Get (proper sentences for comments).

MaskRay:
Thanks for the patch! As, Alexey noted, this patch with a few tweaks can work with llvm-objdump on other (non-Mac) platforms. Both rupprecht and I have made some UI improvements to llvm-objdump 10. llvm-objdump 10 may be closer to GNU objdump. (The function:line pattern just looks like GNU objdump's output)

Consider using `--version` (instead of Sun-style `-version`). GNU objdump doesn't accept `-version`. Latest llvm-objdump does not accept `-version`, either.

internal/binutils/binutils.go
MaskRay: Two-dash `--no-show-raw-insn` should work as well, for all llvm-objdump versions I know. Is it not the case for macOS objdump?

internal/binutils/binutils.go
kalyanac: done.

internal/binutils/binutils.go
kalyanac: done.

internal/binutils/binutils_test.go
kalyanac: done.

kalyanac:
> It feels that the PR description doesn't tell the full story. There is no binutils version 9.0, that seems LLVM objdump. It seems that what we are trying to dispatch on here is binutils vs. LLVM objdump, so the PR description and the code change should be centered around that IMO.

The PR description states `objdump` version should be at least 9.0, not `binutils`. This is from OS X core (`pkgid: com.apple.pkg.Core`) and not the homebrew binutils, which  packages GNU tools. 

Yes, this is (Apple) LLVM objdump. This PR is only to add support for the `objdump` installed with XCode. 
```
objdump -version
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
```
LLVM objdump support for Linux should be a separate, probably small, PR.

kalyanac:
> Thanks for the patch! As Alexey noted, this patch with a few tweaks can work with llvm-objdump on other (non-Mac) platforms. Both rupprecht and I have made some UI improvements to llvm-objdump 10. llvm-objdump 10 may be closer to GNU objdump. (The function:line pattern just looks like GNU objdump's output)

A separate PR will add support for LLVM objdump for non-Mac platforms. 

> Consider using `--version` (instead of Sun-style `-version`). GNU objdump doesn't accept `-version`. Latest llvm-objdump does not accept `-version`, either.

`objdump` on Mac seems to support the `--version` format but does not document it. `--help` only uses the single dash format. 

I am not too familiar with how Apple changes their code from LLVM, so I figured it is safe to stay with documented flags.

aalexand:
> The PR description states `objdump` version should be at least 9.0, not `binutils`. This is from OS X core (`pkgid: com.apple.pkg.Core`) and not the homebrew binutils, which packages GNU tools.

`objdump` is part of binutils package on Linux. On a Mac machine `objdump` can come either from LLVM or from binutils. We should check for whether objdump being used is from LLVM or from binutils, not for the version.

MaskRay:
> > Thanks for the patch! As Alexey noted, this patch with a few tweaks can work with llvm-objdump on other (non-Mac) platforms. Both rupprecht and I have made some UI improvements to llvm-objdump 10. llvm-objdump 10 may be closer to GNU objdump. (The function:line pattern just looks like GNU objdump's output)
> 
> A separate PR will add support for LLVM objdump for non-Mac platforms.
> 
> > Consider using `--version` (instead of Sun-style `-version`). GNU objdump doesn't accept `-version`. Latest llvm-objdump does not accept `-version`, either.
> 
> `objdump` on Mac seems to support the `--version` format but does not document it. `--help` only uses the single dash format.
> 
> I am not too familiar with how Apple changes their code from LLVM, so I figured it is safe to stay with documented flags.

Before LLVM 10, various binary utilities's `--help` do print `-foo`. However, in LLVM 10, many use `--foo` now. It is possible that when `/usr/bin/objdump` on macOS upgrades to LLVM 10 (they contain some local patches but I am unsure what patches they carry), its `--help` may print `--foo`.

For GNU objdump, `objdump -version` is a bit problematic:
```
% objdump -version                                                                                                                                    
objdump: invalid option -- 'o'
```

I think detecting `--version` will be safe.

kalyanac:
I am going to repurpose this PR to add LLVM objdump support for both Mac & Linux. I noticed that the objdump on my Mac is dated (Apple 10.0.0 is LLVM 6.0.0). I updated to Apple LLVM 11.0.3 in XCode 11.4, which is LLVM 9.0.0. gLinux also uses LLVM 9.0.1. 

The output formats seem to match. We will still include binutils objdump support for Linux, but I wont add support for Mac. We can add if users request it.

aalexand:
> The output formats seem to match. We will still include binutils objdump support for Linux, but I wont add support for Mac. We can add if users request it.

I think Linux vs. OSX shouldn't matter. If we detect objdump as LLVM vs. binutils and use appropriate flags then things should work regardless whether it's Linux or OSX.

kalyanac:
`objdump` on Mac is Apple LLVM objdump. `binutils` installs one with same name but not in path, and adds to path `gobjdump`. We have to look for both, parse the version string to see if it is Apple LLVM or binutils objdump and handle any variations in flags, or output formats.

This adds a lot of code to initTools, disassembly and tests for a feature that currently not available on Mac.

aalexand:
> `objdump` on Mac is Apple LLVM objdump. `binutils` installs one with same name but not in path, and adds to path `gobjdump`. We have to look for both, parse the version string to see if it is Apple LLVM or binutils objdump and handle any variations in flags, or output formats.
> 
> This adds a lot of code to initTools, disassembly and tests for a feature that currently not available on Mac.

In this PR we already parse the version string and set the flags based on that. I propose that instead of dispatching on the version number (which I find a weird way of deciding whether we got binutils' or LLVM's objdump), we dispatch on whether "binutils" or "LLVM" string is present in the "--version" output. The rest is pretty much the same - except 1) also remove the "if OSX" condition and 2) look up the objdump tool by both llvm-objdump and objdump names (the former is needed for b/141775016, the latter - for this PR).

Maybe we should work on b/141775016 first. Once that is in place, I think this PR is simply adding the ability to locate the objdump tool by a different name (since it's installed as objdump on Mac, not as llvm-objdump).

kalyanac:
Please review once the tests pass. This should work for all the following combinations. I tested with
* gLinux objdump 2.34
* gLinux llvm-objdump-9 9.0.1
* Macos binutils/objdump 2.34 (for binaries named objdump only, not gobjdump)
* Macos objdump Apple LLVM 11.0.3

aalexand:
CI tests fail FWIW

kalyanac:
Style check failed with a recommendation, all other tests passed. Updating with the recommended style.

kalyanac:
The failures are all on MacOS images with old XCode versions that do not have the correct objdump. The rest of the tests pass. 

Will add back the code to skip test with unsupported objdump on Mac OS.

kalyanac:
The latest failures are not related. Sometimes homebrew takes a while to setup and the test times out. I am restarting them but they seem to keep having the same problem.

wyk9787:
> The latest failures are not related. Sometimes homebrew takes a while to setup and the test times out. I am restarting them but they seem to keep having the same problem.

I have had the same issue for my previous PR as well but couldn't figure out a way to increase Travis' timeout. One possible alternative is to use Github Action instead of Travis.

kalyanac:
@nolanmar511 recently added Github actions on one of the other repos. We should consider adding here as well. These unrelated failures cause significant overhead.

kalyanac:
I can't restart the AppVeyor run without another commit. The failure is unrelated.

internal/binutils/binutils.go
```diff
@@ -39,6 +40,8 @@ type Binutils struct {
 	rep *binrep
 }
 
+var objdumpLLVMVerRE = regexp.MustCompile(`LLVM\sversion (([0-9]*)\.([0-9]*)\.([0-9]*)).*\s`)
```
MaskRay, line 43: An llvm-objdump compiled from llvm-project source: `  LLVM version 11.0.0git` (10.0.0 has been released, so the git HEAD version is now 11.0.0git)

Internal llvm-objdump may use other version scheme like `  llvm version google3-trunk 111111111111111111111111111111111commit-id` (I guess the lowercase `llvm` is not intentional. I can fix it.)

internal/binutils/binutils.go
```diff
@@ -140,7 +144,87 @@ func initTools(b *binrep, config string) {
...
+// it picks objdump with the latest major version. If no LLVM objdump is found, returns the path to
+// any binary named objdump, if found.
+func findObjdump(paths []string) (string, bool, bool) {
+	allObjdump := findAllObjdump("llvm-objdump*", paths)
+
+	if objdump, objdumpFound := findExe("llvm-objdump", paths); objdumpFound {
+		allObjdump = append(allObjdump, objdump)
+	}
+
+	if objdump, objdumpFound := findExe("llvm-objdump-9", paths); objdumpFound {
```
MaskRay, line 161: Not very sure we want to support Debian-flavored `/usr/bin/llvm-objdump-9`

Users can run update-alternatives if they do want `/usr/bin/llvm-objdump`

internal/binutils/binutils.go
```diff
@@ -140,7 +144,87 @@ func initTools(b *binrep, config string) {
...
+		allObjdump = append(allObjdump, objdump)
+	}
+
+	buObjdumpFound := false
+	var buObjdump string
+	if objdump, objdumpFound := findExe("objdump", paths); objdumpFound {
+		cmdOut, _ := exec.Command(objdump, "--version").Output()
+		if strings.Contains(string(cmdOut), "GNU") && strings.Contains(string(cmdOut), "Binutils") {
```
MaskRay, line 169: It may be slightly more robust to check just the first line (`GNU objdump (GNU Binutils for Debian) 2.34`)

internal/binutils/binutils.go
```diff
@@ -159,12 +243,24 @@ func findExe(cmd string, paths []string) (string, bool) {
...
-		fmt.Sprintf("--start-address=%#x", start),
-		fmt.Sprintf("--stop-address=%#x", end)}
+	if !b.objdumpFound && !b.llvmObjdumpFound {
+		return nil, fmt.Errorf("couldn't find objdump")
+	}
+	var args []string
+
+	if b.llvmObjdumpFound {
+		args = []string{"--disassemble-all", "--no-show-raw-insn",
+			"--line-numbers", fmt.Sprintf("--start-address=%#x", start),
```
MaskRay, line 253: Either --line-numbers is -l is accepted by all these objdump programs. Unify the option names.

MaskRay: llvm-objdump (Apple LLVM objdump) should support `-C`. Is it not the case on your machine?

internal/binutils/binutils.go
kalyanac: I can make the regex to match case insensitive but google3-trunk is troublesome. The code will look for a major version and use that to pick the latest one. Other option is to skip the search and just look for 
llvm-objdump, objdump, gobjdump 
in that order

This will also skip looking for llvm-objdump-ver

kalyanac: I will also simplify the regex to only match the major version and ignore the rest.

internal/binutils/binutils.go
kalyanac: gLinux has llvm-ovjdump-7, 8 & 9 in path.

Looks like this might be gLinux specific. They are just symlinks to llvm-objdump is various version locations.

internal/binutils/binutils.go
kalyanac: Yes, looks like they can be unified. I will verify the flags and combine.

internal/binutils/binutils.go
MaskRay: It is now `LLVM version google3-trunk`. If the regex parses a version (`google3-trunk`), the code can just assume the version is new enough.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,87 @@ func initTools(b *binrep, config string) {
...
-	b.objdump, b.objdumpFound = findExe("objdump", append(paths["objdump"], defaultPath...))
+	// MacOS binary is named objdump, Linux binary is named llvm-objdump-version
+	b.objdump, b.llvmObjdumpFound, b.objdumpFound = findObjdump(append(paths["llvm-objdump"], append(paths["objdump"], defaultPath...)...))
+}
+
+// findObjdump finds all binaries named objdump, llvm-objdump*. If multiple LLVM objdump are found,
+// it picks objdump with the latest major version. If no LLVM objdump is found, returns the path to
+// any binary named objdump, if found.
+func findObjdump(paths []string) (string, bool, bool) {
+	allObjdump := findAllObjdump("llvm-objdump*", paths)
```
aalexand, line 155: findExe uses exec.LookPath and findAllObjdump does not, which changes the semantics of how the paths are used here. To avoid having to document that subtle difference, can we avoid having to do the globbing and instead look up the objdump tools by their exact known names? We can list as many of them as needed, but I'm thinking just "llvm-objdump", "objdump", plus "gobjdump" on OSX should be enough.

internal/binutils/binutils.go
aalexand: I am not sure we want to manage the selection of the versioned binary names. I think it's reasonable to assume that there is unversioned symlink (llvm-objdump) available and if not then user should make sure that there is.

internal/binutils/binutils.go
aalexand, line 43: Nit: why " " vs. "\s" in a couple places? Can these strings have tab characters in them?

"[0-9]*" should be "[0-9]+" I think. Also "\d+" is shorter and more common.

Outer brackets around the version string are not needed I think.

".*\s" at the end shouldn't be needed I think, since regexps are greedy.

Putting this together: `LLVM version (\d+)\.(\d+)\.(\d+)`

internal/binutils/binutils_test.go
```diff
@@ -188,15 +190,51 @@ func skipUnlessDarwinAmd64(t *testing.T) {
 	}
 }
 
+// ensureObjdump checks if a supported version of objdump is availabe. Keep the test simple
+// and check for LLVM objdump v11.0+ or a GNU Binutils objdump.
+func ensureObjdump() bool {
+	cmdOut, _ := exec.Command("objdump", "--version").Output()
+	// Keep this regexp in sync with the one used in binutils.go
+	//	re := regexp.MustCompile(`LLVM\sversion (([0-9]*)\.([0-9]*)\.([0-9]*)).*\s`)
```
aalexand, line 198: This line and the comment above can be removed, as the code reuses the regular expression so there is nothing to keep in sync?

internal/binutils/binutils_test.go
```diff
@@ -188,15 +190,51 @@ func skipUnlessDarwinAmd64(t *testing.T) {
 	}
 }
 
+// ensureObjdump checks if a supported version of objdump is availabe. Keep the test simple
+// and check for LLVM objdump v11.0+ or a GNU Binutils objdump.
+func ensureObjdump() bool {
+	cmdOut, _ := exec.Command("objdump", "--version").Output()
```
aalexand, line 196: Do not ignore erorrs.

internal/binutils/binutils_test.go
```diff
@@ -188,15 +190,51 @@ func skipUnlessDarwinAmd64(t *testing.T) {
...
+	// Keep this regexp in sync with the one used in binutils.go
+	//	re := regexp.MustCompile(`LLVM\sversion (([0-9]*)\.([0-9]*)\.([0-9]*)).*\s`)
+	fields := objdumpLLVMVerRE.FindStringSubmatch(string(cmdOut))
+	if len(fields) == 5 {
+		if ver, _ := strconv.Atoi(fields[2]); ver >= 11.0 {
+			return true
+		}
+	}
+
+	cmdOut, _ = exec.Command("objdump", "--version").Output()
```
aalexand, line 206: Do not ignore errors.

internal/binutils/binutils_test.go
```diff
@@ -188,15 +190,51 @@ func skipUnlessDarwinAmd64(t *testing.T) {
...
+	if strings.Contains(string(cmdOut), "GNU") && strings.Contains(string(cmdOut), "Binutils") {
+		return true
+	}
+
+	cmdOut, _ = exec.Command("gobjdump", "--version").Output()
```
aalexand, line 211: Do not ignore errors.

internal/binutils/binutils_test.go
```diff
@@ -188,15 +190,51 @@ func skipUnlessDarwinAmd64(t *testing.T) {
...
+		return true
+	}
+	return false
+}
+
 func testDisasm(t *testing.T, intelSyntax bool) {
+	// The check is only performed on MacOS. Linux usually has a supported
+	// version of objdump available.
+	if runtime.GOOS == "darwin" && !ensureObjdump() {
+		t.Skip("cannot find usable objdump")
```
aalexand, line 222: nit: Cannot

(use sentence case in t.{Log*,Skip} methods)

internal/binutils/binutils.go
kalyanac: @aalexand Ok to do this check or should we do it in the internal version only?

internal/binutils/binutils.go
kalyanac: I'll remove this check.

internal/binutils/binutils.go
kalyanac: I removed this code since we will require llvm-objdump to be available.

internal/binutils/binutils.go
kalyanac: The regex is also not searching only on the first line, so functionally these two seem similar. We also do not want to search only first line. For example, `<objdump> --version` prints version in different lines on mac & linux

internal/binutils/binutils.go
```diff
@@ -140,7 +144,87 @@ func initTools(b *binrep, config string) {
...
+			finalObjdump = objdump
+		}
+	}
+
+	if llvmObjdumpFound {
+		if runtime.GOOS == "darwin" && finalVersion >= 11 {
+			// Ensure objdump is at least version 11.0 on MacOS.
+			return finalObjdump, llvmObjdumpFound, false
+		}
+		if runtime.GOOS != "darwin" && finalVersion >= 9 {
```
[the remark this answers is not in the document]
kalyanac: changed

internal/binutils/binutils_test.go
kalyanac: Forgot to remove. Removed now

internal/binutils/binutils_test.go
kalyanac: done.

internal/binutils/binutils.go
kalyanac: I used a similar expression now.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,87 @@ func initTools(b *binrep, config string) {
...
+		if ver, _ := strconv.Atoi(fields[2]); ver > finalVersion {
+			finalVersion = ver
+			llvmObjdumpFound = true
+			finalObjdump = objdump
+		}
+	}
+
+	if llvmObjdumpFound {
+		if runtime.GOOS == "darwin" && finalVersion >= 11 {
+			// Ensure objdump is at least version 11.0 on MacOS.
```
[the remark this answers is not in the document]
kalyanac: The flags displayed by --help were different. I tested 6, 8, 9 (those available on gLinux) and settled on a minimum version of 8 for Linux and 10.0 for Mac.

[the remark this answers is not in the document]
kalyanac: XCode 10.0.0 is LLVM 6.0.1

internal/binutils/binutils.go
kalyanac: Done.

kalyanac:
Some Mac combinations which were not tested in CI

```
$ ./objdump --version
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
  Optimized build.
  Default target: x86_64-apple-darwin19.4.0
  Host CPU: skylake
....

$> PATH=./ /usr/local/go/bin/go test -v
...
=== RUN   TestDisasm
--- PASS: TestDisasm (0.03s)
=== RUN   TestObjFile
    TestObjFile: binutils_test.go:181: This test only works on x86-64 Linux
--- SKIP: TestObjFile (0.00s)
```


```
$ objdump --version
Apple LLVM version 11.0.3 (clang-1103.0.32.62)
  Optimized build.
  Default target: x86_64-apple-darwin19.4.0
  Host CPU: skylake
...

$ /usr/local/go/bin/go test -v
...
=== RUN   TestDisasm
--- PASS: TestDisasm (0.08s)
=== RUN   TestObjFile
    TestObjFile: binutils_test.go:181: This test only works on x86-64 Linux
--- SKIP: TestObjFile (0.00s)
....
```

kalyanac:
Failures are due to timeouts. PR can be reviewed

internal/binutils/binutils.go
```diff
@@ -39,6 +40,8 @@ type Binutils struct {
 	rep *binrep
 }
 
+var objdumpLLVMVerRE = regexp.MustCompile(`((?i)LLVM) ((?i)version) (\d*)\.(\d*)\.(\d*)`)
```
aalexand, line 43: Can this be simply `(?i)LLVM) version (\d*)\.(\d*)\.(\d*)`?

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
 	}
 	b.nm, b.nmFound = findExe("nm", append(paths["nm"], defaultPath...))
-	b.objdump, b.objdumpFound = findExe("objdump", append(paths["objdump"], defaultPath...))
+	// Look for llvm-objdump and objdump
+	b.objdump, b.llvmObjdumpFound, b.objdumpFound = FindObjdump(append(paths["llvm-objdump"], append(paths["objdump"], append(paths["gobjdump"], defaultPath...)...)...))
+}
+
+// FindObjdump finds and returns preferred objdump binary. Order of preference is: llvm-objdump, objdump.
+// On MacOS only, also looks for gobjdump with least preference.
+func FindObjdump(paths []string) (string, bool, bool) {
```
aalexand, line 153: Does this function need to be public?

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
 		b.addr2line, b.addr2lineFound = findExe("gaddr2line", append(paths["addr2line"], defaultPath...))
 	}
 	b.nm, b.nmFound = findExe("nm", append(paths["nm"], defaultPath...))
-	b.objdump, b.objdumpFound = findExe("objdump", append(paths["objdump"], defaultPath...))
+	// Look for llvm-objdump and objdump
+	b.objdump, b.llvmObjdumpFound, b.objdumpFound = FindObjdump(append(paths["llvm-objdump"], append(paths["objdump"], append(paths["gobjdump"], defaultPath...)...)...))
```
aalexand, line 148: For addr2line above we use paths[addr2line] for both addr2line and gaddr2line cases. We should probably be consistent with that here.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
 		b.addr2line, b.addr2lineFound = findExe("gaddr2line", append(paths["addr2line"], defaultPath...))
 	}
 	b.nm, b.nmFound = findExe("nm", append(paths["nm"], defaultPath...))
-	b.objdump, b.objdumpFound = findExe("objdump", append(paths["objdump"], defaultPath...))
+	// Look for llvm-objdump and objdump
```
aalexand, line 147: nit nit: trailing dot (keep comments full sentences)

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
+	b.objdump, b.llvmObjdumpFound, b.objdumpFound = FindObjdump(append(paths["llvm-objdump"], append(paths["objdump"], append(paths["gobjdump"], defaultPath...)...)...))
+}
+
+// FindObjdump finds and returns preferred objdump binary. Order of preference is: llvm-objdump, objdump.
```
aalexand, line 151: Keep comments in 80 cols.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
+func FindObjdump(paths []string) (string, bool, bool) {
+	objdumpNames := []string{"llvm-objdump", "objdump"}
+	if runtime.GOOS == "darwin" {
+		objdumpNames = append(objdumpNames, "gobjdump")
+	}
+
+	for _, objdumpName := range objdumpNames {
+		if objdump, objdumpFound := findExe(objdumpName, paths); objdumpFound {
+			cmdOut, err := exec.Command(objdump, "--version").Output()
+			if err != nil {
```
aalexand, line 162: Should we fail if there is an error? If this command fails it is probably a bug we need to fix, not user error.

https://github.com/golang/go/wiki/CodeReviewComments#handle-errors

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
+				continue
+			}
+			if llvmObjdumpFound := isLlvmObjdump(string(cmdOut)); llvmObjdumpFound {
```
aalexand, line 165: Llvm -> LLVM

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
+				return objdump, false, buObjdumpFound
+			}
+		}
+	}
+	return "", false, false
+}
+
+func isLlvmObjdump(output string) bool {
+	fields := objdumpLLVMVerRE.FindStringSubmatch(output)
+	if len(fields) == 6 {
```
aalexand, line 178:
```
if len(fields) != 6 {
  return false
}
```

and unindent.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
+		}
+		if runtime.GOOS == "darwin" {
+			if verMajor > 10 {
+				return true
+			}
+			// Ensure LLVM objdump is at least version 10.0.1 on MacOS.
+			if verMajor == 10 && verPatch >= 1 {
+				return true
+			}
+			return false
```
aalexand, line 201: return verMajor > 10 || (verMajor == 10 && verPatch >= 1)

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
+		}
+	}
+	return false
+}
+
+func isBuObjdump(output string) bool {
+	if strings.Contains(output, "GNU objdump") && strings.Contains(output, "Binutils") {
```
aalexand, line 208: return strings.Contains(output, "GNU objdump") && strings.Contains(output, "Binutils")

internal/binutils/binutils.go
```diff
@@ -159,12 +228,15 @@ func findExe(cmd string, paths []string) (string, bool) {
 // of a binary.
 func (bu *Binutils) Disasm(file string, start, end uint64, intelSyntax bool) ([]plugin.Inst, error) {
 	b := bu.get()
-	args := []string{"-d", "-C", "--no-show-raw-insn", "-l",
-		fmt.Sprintf("--start-address=%#x", start),
+	if !(b.objdumpFound || b.llvmObjdumpFound) {
```
aalexand, line 231: I think `if !b.objdumpFound && !b.llvmObjdumpFound` would be easier to read, but maybe it's just me.

internal/binutils/disasm.go
```diff
@@ -25,10 +25,11 @@ import (
...
 var (
-	nmOutputRE            = regexp.MustCompile(`^\s*([[:xdigit:]]+)\s+(.)\s+(.*)`)
-	objdumpAsmOutputRE    = regexp.MustCompile(`^\s*([[:xdigit:]]+):\s+(.*)`)
-	objdumpOutputFileLine = regexp.MustCompile(`^(.*):([0-9]+)`)
-	objdumpOutputFunction = regexp.MustCompile(`^(\S.*)\(\):`)
+	nmOutputRE                = regexp.MustCompile(`^\s*([[:xdigit:]]+)\s+(.)\s+(.*)`)
+	objdumpAsmOutputRE        = regexp.MustCompile(`^\s*([[:xdigit:]]+):\s+(.*)`)
+	objdumpOutputFileLine     = regexp.MustCompile(`^(.*):([0-9]+)`)
+	objdumpOutputFunction     = regexp.MustCompile(`^(\S.*)\(\):`)
+	objdumpOutputFunctionLLVM = regexp.MustCompile(`^([[:xdigit:]]+)?\s?(.*):`)
```
aalexand, line 32: Can we update disasm_test.go to cover this case?

internal/binutils/binutils_test.go
```diff
@@ -189,14 +189,25 @@ func skipUnlessDarwinAmd64(t *testing.T) {
...
+	}
+
+	insts, err := bu.Disasm(filepath.Join("testdata", testexe), 0, math.MaxUint64, intelSyntax)
 	if err != nil {
 		t.Fatalf("Disasm: unexpected error %v", err)
 	}
 	mainCount := 0
 	for _, x := range insts {
-		if x.Function == "main" {
+		if x.Function == "main" || x.Function == "_main" {
```
aalexand, line 210: A comment would be good on main vs. _main.

internal/binutils/binutils.go
aalexand: This would be even simpler if instead of objdumpFound and llvmObjdumpFound we'd have objdumpFound and isLLVMObjdump where the first flag is for whether the tool is found and the second for what flavor it is.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,72 @@ func initTools(b *binrep, config string) {
...
+			}
+			if buObjdumpFound := isBuObjdump(string(cmdOut)); buObjdumpFound {
+				return objdump, false, buObjdumpFound
+			}
+		}
+	}
+	return "", false, false
+}
+
+func isLlvmObjdump(output string) bool {
```
aalexand, line 176: I'd consider adding unit tests for isLLVMObjdump and isBinutilsObjdump to check the pattern maching.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Done. Done.

internal/binutils/binutils.go
kalyanac: Done. 
I implemented a version check at one point and dropped it later. Thanks for catching it.

internal/binutils/binutils.go
kalyanac: Done. Done.

internal/binutils/disasm.go
kalyanac: It will fail when the code attempts to use objdump in Disasm().

internal/binutils/binutils_test.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: It will fail when the code attempts to use objdump in Disasm().

internal/binutils/binutils.go
kalyanac: We'll check for `trunk` and accept it.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/disasm.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: No, made private.

internal/binutils/binutils.go
kalyanac: Leaving the current expression as is.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils_test.go
```diff
@@ -189,14 +189,25 @@ func skipUnlessDarwinAmd64(t *testing.T) {
...
+			// Mac binaries have a leading underscore.
```
kalyanac, line 210: symbols, not binaries

kalyanac:
Failures due to timeout, restarted. Code ok to review.

kalyanac:
gentle ping.

internal/binutils/binutils.go
aalexand: `(?i)LLVM) version (\d*)\.(\d*)\.(\d*)` does the same but is shorter.

https://play.golang.org/p/MuS-HtOiBo4

internal/binutils/binutils.go
```diff
@@ -140,7 +144,65 @@ func initTools(b *binrep, config string) {
 		b.addr2line, b.addr2lineFound = findExe("gaddr2line", append(paths["addr2line"], defaultPath...))
 	}
 	b.nm, b.nmFound = findExe("nm", append(paths["nm"], defaultPath...))
-	b.objdump, b.objdumpFound = findExe("objdump", append(paths["objdump"], defaultPath...))
+	// Look for llvm-objdump and objdump.
```
aalexand, line 147: nit: I'd remove this comment. The doc to findObjdump function is sufficient. And this comment is incomplete since it doesn't mention gobjdump. It's better to keep the comment in one place, at findObjdump.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,65 @@ func initTools(b *binrep, config string) {
...
+	b.objdump, b.isLLVMObjdump, b.objdumpFound = findObjdump(append(paths["objdump"], defaultPath...))
+}
+
+// findObjdump finds and returns preferred objdump binary.
```
aalexand, line 151: Document return values.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,65 @@ func initTools(b *binrep, config string) {
...
+	}
+	return "", false, false
+}
+
+func isLLVMObjdump(output string) bool {
+	fields := objdumpLLVMVerRE.FindStringSubmatch(output)
+	if len(fields) != 6 {
+		// Check if built from trunk.
+		output = strings.ToLower(output)
+		return strings.Contains(output, "trunk") && strings.Contains(output, "llvm version")
```
aalexand, line 182: I would change the regular expression to something like `(?i)LLVM) version (?:(\d*)\.(\d*)\.(\d*)|.*(trunk).*)`. This would simplify this code:
* No need for ToLower and string.Contains(llvm output) check.
* If capture group 4 is "trunk" then it's trunk.
* Otherwise capture group 1, 2 and 3 must be numbers and is the version number.

internal/binutils/binutils.go
MaskRay: You can drop `(?i)` for `LLVM`. I have fixed the internal issue (lowercase `llvm`).

internal/binutils/binutils.go
MaskRay: You can drop `ToLower` or `(?i)` for `llvm`. I have fixed the internal issue.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: done.

internal/binutils/binutils.go
kalyanac: done.

internal/binutils/binutils_test.go
```diff
@@ -189,14 +189,25 @@ func skipUnlessDarwinAmd64(t *testing.T) {
...
+
+	insts, err := bu.Disasm(filepath.Join("testdata", testexe), 0, math.MaxUint64, intelSyntax)
 	if err != nil {
 		t.Fatalf("Disasm: unexpected error %v", err)
 	}
 	mainCount := 0
 	for _, x := range insts {
-		if x.Function == "main" {
+		if x.Function == "main" || x.Function == "_main" {
+			// Mac symbols have a leading underscore.
```
aalexand, line 210: nit: I think it's more common to put such comment before the if statement rather than in its body. As is now, I'd read the code as if the comment applies to "mainCount++" which looks confusing.

internal/binutils/binutils_test.go
```diff
@@ -401,3 +414,105 @@ func TestOpenMalformedMachO(t *testing.T) {
...
+	// Test that the objdump version strings are parsed properly.
+
+	var version string
+	if runtime.GOOS == "darwin" {
+		// Valid Apple LLVM version string with usable version.
+		version = `Apple LLVM version 11.0.3 (clang-1103.0.32.62)
+		Optimized build.
+		Default target: x86_64-apple-darwin19.4.0
+		Host CPU: skylake`
+		if runtime.GOOS == "darwin" && !isLLVMObjdump(version) {
```
aalexand, line 428: runtime.GOOS == "darwin" is already checked above, this check is redundant.

internal/binutils/binutils_test.go
```diff
@@ -401,3 +414,105 @@ func TestOpenMalformedMachO(t *testing.T) {
 		t.Errorf("Open: got %v, want error containing 'Mach-O'", err)
 	}
 }
+
+func TestObjdumpVersionChecks(t *testing.T) {
+	// Test that the objdump version strings are parsed properly.
+
+	var version string
+	if runtime.GOOS == "darwin" {
```
aalexand, line 422: Consider making this test a table-driven test, which is common in Go? https://github.com/golang/go/wiki/TableDrivenTests

internal/binutils/binutils.go
```diff
@@ -140,7 +144,76 @@ func initTools(b *binrep, config string) {
...
+	}
+
+	for _, objdumpName := range objdumpNames {
+		if objdump, objdumpFound := findExe(objdumpName, paths); objdumpFound {
+			cmdOut, err := exec.Command(objdump, "--version").Output()
+			if err != nil {
+				continue
+			}
+			if llvmObjdumpFound := isLLVMObjdump(string(cmdOut)); llvmObjdumpFound {
+				return objdump, llvmObjdumpFound, llvmObjdumpFound
```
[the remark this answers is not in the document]
kalyanac: Done. 
Is this a style preference or is there a specific reason? 
Since there are two bools, specifying the variable makes it more readable I think.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,76 @@ func initTools(b *binrep, config string) {
...
+			}
+			if buObjdumpFound := isBuObjdump(string(cmdOut)); buObjdumpFound {
+				return objdump, false, buObjdumpFound
```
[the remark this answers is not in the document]
kalyanac: Done.

internal/binutils/binutils_test.go
kalyanac: Done.

internal/binutils/binutils_test.go
kalyanac: Done.

internal/binutils/binutils.go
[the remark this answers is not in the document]
aalexand: If a variable is returned, I'd assume its value can be different at this point. Understanding that its value is actually the same is additional mental load which is nice to avoid when reading code.

internal/binutils/binutils_test.go
kalyanac: Done.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,76 @@ func initTools(b *binrep, config string) {
 		b.addr2line, b.addr2lineFound = findExe("gaddr2line", append(paths["addr2line"], defaultPath...))
 	}
 	b.nm, b.nmFound = findExe("nm", append(paths["nm"], defaultPath...))
-	b.objdump, b.objdumpFound = findExe("objdump", append(paths["objdump"], defaultPath...))
+	b.objdump, b.isLLVMObjdump, b.objdumpFound = findObjdump(append(paths["objdump"], defaultPath...))
```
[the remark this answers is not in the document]
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Great, thanks. Done.

kalyanac:
Restarted timedout tests.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,76 @@ func initTools(b *binrep, config string) {
...
+		objdumpNames = append(objdumpNames, "gobjdump")
+	}
+
+	for _, objdumpName := range objdumpNames {
+		if objdump, objdumpFound := findExe(objdumpName, paths); objdumpFound {
+			cmdOut, err := exec.Command(objdump, "--version").Output()
+			if err != nil {
+				continue
+			}
+			if llvmObjdumpFound := isLLVMObjdump(string(cmdOut)); llvmObjdumpFound {
```
aalexand, line 170: the var is not needed now:

```
if isLLVMObjdump(string(cmdOut)) {
```

internal/binutils/binutils.go
```diff
@@ -140,7 +144,76 @@ func initTools(b *binrep, config string) {
...
+				return objdump, true, true
+			}
+			if buObjdumpFound := isBuObjdump(string(cmdOut)); buObjdumpFound {
```
aalexand, line 173: Same here.

internal/binutils/binutils.go
```diff
@@ -140,7 +144,76 @@ func initTools(b *binrep, config string) {
...
+	if err != nil {
+		return false
+	}
+	if runtime.GOOS == "linux" && verMajor >= 8 {
+		// Ensure LLVM objdump is at least version 8.0 on Linux.
+		// Some flags are not supported by previous versions.
+		return true
+	}
+	if runtime.GOOS == "darwin" {
+		// Ensure LLVM objdump is at least version 10.0.1 on MacOS.
```
aalexand, line 206: Curious - why require a newer version on OSX compared to Linux?

internal/binutils/binutils.go
```diff
@@ -140,7 +144,76 @@ func initTools(b *binrep, config string) {
...
+	if err != nil {
+		return false
+	}
+	verPatch, err := strconv.Atoi(fields[3])
+	if err != nil {
+		return false
+	}
+	if runtime.GOOS == "linux" && verMajor >= 8 {
+		// Ensure LLVM objdump is at least version 8.0 on Linux.
+		// Some flags are not supported by previous versions.
```
aalexand, line 202: Would be good to be more specific and say which flag is not supported by versions < 8.0, instead of "some".

internal/binutils/binutils.go
```diff
@@ -159,13 +232,16 @@ func findExe(cmd string, paths []string) (string, bool) {
 // of a binary.
 func (bu *Binutils) Disasm(file string, start, end uint64, intelSyntax bool) ([]plugin.Inst, error) {
 	b := bu.get()
-	args := []string{"-d", "-C", "--no-show-raw-insn", "-l",
-		fmt.Sprintf("--start-address=%#x", start),
+	if !b.objdumpFound {
+		return nil, fmt.Errorf("couldn't find objdump")
```
aalexand, line 236: I think saying "couldn't find objdump" in the error here is somewhat misleading since we didn't try to search for the tool here but the error may make it look like we did. I'd use something like "cannot disasm: no objdump tool available" which might be a bit clearer.

aalexand, line 236: errors.New() as no arguments to format.

internal/binutils/binutils_test.go
```diff
@@ -401,3 +414,125 @@ func TestOpenMalformedMachO(t *testing.T) {
 		t.Errorf("Open: got %v, want error containing 'Mach-O'", err)
 	}
 }
+
+func TestObjdumpVersionChecks(t *testing.T) {
+	// Test that the objdump version strings are parsed properly.
+	type objdumpTest struct {
```
aalexand, line 420: This is commonly named "testcase".

internal/binutils/binutils_test.go
```diff
@@ -401,3 +414,125 @@ func TestOpenMalformedMachO(t *testing.T) {
...
+		desc string
+		os   string
+		ver  string
+		want bool
+	}
+
+	for _, tc := range []objdumpTest{
+		{
+			// Valid Apple LLVM version string with usable version.
```
aalexand, line 429: Here and below - no need for this comment, the `desc` field covers this.

internal/binutils/binutils_test.go
```diff
@@ -401,3 +414,125 @@ func TestOpenMalformedMachO(t *testing.T) {
...
+			ver: `Apple LLVM version 11.0.3 (clang-1103.0.32.62)
+			Optimized build.`,
+			want: true,
+		},
+		{
+			// Valid Apple LLVM version string with unusable version.
+			desc: "Unusable Apple LLVM version",
+			os:   "darwin",
+			ver: `Apple LLVM version 10.0.0 (clang-1000.11.45.5)
+			Optimized build.`,
```
aalexand, line 441: I think using usual string instead of backtick string would be more readable: `"Apple LLVM version 10.0.0 (clang-1000.11.45.5)\nOptimized build."`

It would also more precisely match the output of the tool - with backtick strings there is some leading whitespace on 2+ lines due to the indentation.

Similar below.

internal/binutils/binutils_test.go
```diff
@@ -401,3 +414,125 @@ func TestOpenMalformedMachO(t *testing.T) {
...
+			if got := isLLVMObjdump(tc.ver); got != tc.want {
+				t.Errorf("%v: got %v, want %v", tc.desc, got, tc.want)
+			}
+		}
+	}
+	for _, tc := range []objdumpTest{
+		{
+			// Valid GNU objdump version string.
+			desc: "Valid GNU objdump version string",
+			os:   runtime.GOOS,
```
aalexand, line 520: Looks like this field isn't used below, so no need to set it?

internal/binutils/disasm_test.go
```diff
@@ -132,6 +132,25 @@ func TestFunctionAssembly(t *testing.T) {
 				{Addr: 0x2001, Text: "instruction two"},
 			},
 		},
+		{
+			plugin.Sym{Name: []string{"_main"}, Start: 0x30000, End: 0x3FFF},
+			`_main:
```
aalexand, line 137: nit: here similar to a comment in another test file - a double-quote string with escaped newline would read a bit easier than the backquote string with no indentation on 2+ lines.

internal/binutils/binutils.go
kalyanac: The LLVM version is not same as the version reported by Apple objdump. 
See [this page](https://en.wikipedia.org/wiki/Xcode#Xcode_7.0_-_11.x_(since_Free_On-Device_Development)).

internal/binutils/disasm_test.go
kalyanac: Done. Exceeding 80 columns in some cases.

internal/binutils/binutils_test.go
kalyanac: Done.

internal/binutils/binutils_test.go
kalyanac: Done. Exceeding 80 columns in some cases as a result.

internal/binutils/binutils_test.go
kalyanac: Done.

internal/binutils/binutils_test.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Apple version 10.0.1 is LLVM 7.0.0

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Done.

internal/binutils/binutils.go
kalyanac: Done.

kalyanac:
Gentle ping

PR #535, opened 2020-05-06
Add support for saving the current configuration settings

ghemawat:
Add support for saving the current configuration settings as a named
configuration and reloading such a configuration later.

All of the configurations are stored in a settings file, by default:
~/.config/pprof.json

A struct named config replaces the old map[string]string which used
to store the set of configurable options. Instead, each option is
now a field in the config struct.

The UI now has a new 'Config' menu with

Save as: prompts for a name and saves current config.
Default: applies default config.
$X: for every config named X: applies settings from X.

The currently selected config is highlighted.

Every named config has a delete button to control deletion.

Some semi-related changes:
1. Both filefunctions and functions in the granularity group had an
   initial value of true, which is incorrect since these are mutually
   incompatible choices. Set filefunctions' initial value to false.
2. Renamed the group for sorting from "cumulative" to "sort".
3. Store testing.T in TestUI instead of leaving the field nil.


rsc:
Nice. Can you use os.UserConfigDir to find the config dir? That will do all the things in the code right now but play nicer on Macs (using $HOME/Library/Application Support).

ghemawat:
> Nice. Can you use os.UserConfigDir to find the config dir? That will do all the things in the code right now but play nicer on Macs (using $HOME/Library/Application Support).

Thanks for the suggestion.

I think UserConfigDir was added in Go 1.13. Alex, which older versions of Go is pprof promising to support currently. Would it be a problem to require 1.13 or newer?

aalexand:
> > Nice. Can you use os.UserConfigDir to find the config dir? That will do all the things in the code right now but play nicer on Macs (using $HOME/Library/Application Support).
> 
> Thanks for the suggestion.
> 
> I think UserConfigDir was added in Go 1.13. Alex, which older versions of Go is pprof promising to support currently. Would it be a problem to require 1.13 or newer?

@ghemawat We follow https://golang.org/doc/devel/release.html#policy, so since 1.14 is the current we support 1.13 and 1.14, so requiring 1.13+ should be fine.

ghemawat:
Switched to using os.UserConfigDir() as suggested by Russ.

PR #542, opened 2020-06-02
settings: Only call MkdirAll on save

evanj:
Previously settingsFileName() created the ~/.config/pprof directory
if it did not exist. However, this is not possible for user nobody,
since its home directory is set to "/nonexistent". Instead, only
create the directory when we actually attempt to save the file, so
the error will happen at the appropriate interaction. This als
prevents pprof from creating empty settings directories.

Fixes the following error when running pprof's web UI as user nobody:

    mkdir /nonexistent: permission denied

### DIFF
--- a/internal/driver/settings.go
+++ b/internal/driver/settings.go
@@ -61,7 +61,7 @@ func writeSettings(fname string, settings *settings) error {
 	// XDG specifies permissions 0700 when creating settings dirs:
 	// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
 	if err := os.MkdirAll(filepath.Dir(fname), 0700); err != nil {
-		return err
+		return fmt.Errorf("failed to create settings directory: %w", err)
 	}
 
 	if err := ioutil.WriteFile(fname, data, 0644); err != nil {
--- a/internal/driver/settings.go
+++ b/internal/driver/settings.go
@@ -28,11 +28,7 @@ func settingsFileName() (string, error) {
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


internal/driver/settings.go
```diff
@@ -60,6 +56,17 @@ func writeSettings(fname string, settings *settings) error {
 	if err != nil {
 		return fmt.Errorf("could not encode settings: %w", err)
 	}
+
+	// create the settings directory if it does not exist
+	fdir := filepath.Dir(fname)
+	if _, err := os.Stat(fdir); os.IsNotExist(err) {
```
aalexand, line 62: Why is this check needed? I.e. why not os.MkdirAll(fdir, 0700)?

evanj: Great suggestion. Mostly because I didn't think of it. It turns out: this stat is exactly what MkdirAll does. Oops! Removing this now.

internal/driver/settings.go
```diff
@@ -60,6 +56,14 @@ func writeSettings(fname string, settings *settings) error {
...
 	}
+
+	// create the settings directory if it does not exist
+	// XDG specifies permissions 0700 when creating settings dirs:
+	// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+	if err := os.MkdirAll(filepath.Dir(fname), 0700); err != nil {
+		return fmt.Errorf("failed to create settings directory: %w", err)
+	}
+
 	if err := ioutil.WriteFile(fname, data, 0644); err != nil {
```
aalexand, line 67: Maybe this should have been changed to 0600 too.

PR #554, opened 2020-07-08
Remove extra version check on GNU objdump.

wyk9787:
Remove the extra version check to be compatible with different versions of GNU objdump (e.g. specialized version built internally within Google).


kalyanac:
Ok to merge after tests pass.

PR #556, opened 2020-08-05
Change llvm-symbolizer to use --inlines

MaskRay:
The single dash -inlining is exotic. Change it to the double-dash and canonical
option --inlines (canonical in the sense that --inlines is also supported by
addr2line). The one-dash long option is discouraged due to its collision with
grouped short options.

-demangle=false is also exotic but not changed. It is currently
implemented as a legacy alias (llvm/llvm-project@c558c22cab9a555d2e521102b775759381e9727f).
We can change it to --no-demangle once support for llvm-symbolizer < 9.0 is
dropped (https://reviews.llvm.org/D56773).

### DIFF
--- a/internal/binutils/addr2liner_llvm.go
+++ b/internal/binutils/addr2liner_llvm.go
@@ -70,7 +70,7 @@ func newLLVMSymbolizer(cmd, file string, base uint64) (*llvmSymbolizer, error) {
 	}
 
 	j := &llvmSymbolizerJob{
-		cmd: exec.Command(cmd, "-inlining", "-demangle=false"),
+		cmd: exec.Command(cmd, "--inlines", "-demangle=false"),
 	}
 
 	var err error


aalexand:
Looks like CI on go master branch is failing:

```
--- FAIL: TestObjFile (0.10s)
    --- FAIL: TestObjFile/fake_mapping (0.08s)
    --- FAIL: TestObjFile/fixed_load_address (0.01s)
    --- FAIL: TestObjFile/simulated_ASLR_address (0.01s)
```

MaskRay:
I suspect some configurations have llvm-symbolizer<9. How do I detect llvm-symbolizer versions?

Or can I fix `-inlining` only and leave --no-demangle to a Go expert? :)

kalyanac:
Thank you for sending the PR. If the intent is to fix the flags, can you post why the tests failed? The failure is on Xenial Ubuntu  (18.04) with Go master. I see Ubuntu 18.04 passing with Go 1.13 and 1.14.

MaskRay:
> Thank you for sending the PR. If the intent is to fix the flags, can you post why the tests failed? The failure is on Xenial Ubuntu (18.04) with Go master. I see Ubuntu 18.04 passing with Go 1.13 and 1.14.

I think the likely reason it failed is because 18.04 uses llvm-symbolizer<9.0 . We have to run `llvm-symbolizer --version` to retrieve the version. However, this just causes one additional subprocess invocation. We can change it in the future when llvm-symbolizer<9.0 support is dropped.

kalyanac:
> I think the likely reason it failed is because 18.04 uses llvm-symbolizer<9.0 . 

18.04 with Go 1.13 and 1.14 passed, only the test with master failed.

MaskRay:
I can guess the failure. `--inlines` is a llvm-objdump 9.0.0 option (https://reviews.llvm.org/D57083) and the bot may be using an older one. Before, llvm-symbolizer's command line inherits the style of many other LLVM internal utilities. The command line parsing is loose and exotic in some cases (`-inlines=0` `-inlines=false` instead of `--no-inlines`). That was one reason I wanted to switch llvm-symbolizer to behave more like GNU getopt_long (https://reviews.llvm.org/D83530)

I do want pprof to move away from llvm-symbolizer's legacy aliases but I don't know how to do now.. I think improving this will require non-trivial code. I am afraid that my Go knowledge doesn't catch up:/ Maybe I will just discard this patch. If any of you thinks this is worth fixing I will appreciate it:)

PR #557, opened 2020-09-02
Escape labels when generating graph

auxiliary:
Profile comments can have various special symbols which are then inserted as labels when generating graphs. If symbols are not escaped, graph generation can fail. This PR fixes that.

### DIFF
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -127,7 +127,7 @@ func (b *builder) addLegend() {
 	}
 	title := labels[0]
 	fmt.Fprintf(b, `subgraph cluster_L { "%s" [shape=box fontsize=16`, title)
-	fmt.Fprintf(b, ` label="%s\l"`, strings.Join(labels, `\l`))
+	fmt.Fprintf(b, ` label="%s\l"`, strings.Join(escapeForDot(labels), `\l`))
 	if b.config.LegendURL != "" {
 		fmt.Fprintf(b, ` URL="%s" target="_blank"`, b.config.LegendURL)
 	}
@@ -471,4 +471,15 @@ func min64(a, b int64) int64 {
 		return a
 	}
 	return b
+}
+
+// escapeForDot escapes double quotes and backslashes, and replaces Graphviz's
+// "center" character (\n) with a left-justified character.
+// See https://graphviz.org/doc/info/attrs.html#k:escString for more info.
+func escapeForDot(in []string) []string {
+	var out = make([]string, len(in))
+	for i := range in {
+		out[i] = strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(in[i], `\`, `\\`), `"`, `\"`), "\n", `\l`)
+	}
+	return out
 }
--- a/internal/graph/dotgraph_test.go
+++ b/internal/graph/dotgraph_test.go
@@ -178,7 +178,7 @@ func baseAttrsAndConfig() (*DotAttributes, *DotConfig) {
 	}
 	c := &DotConfig{
 		Title:  "testtitle",
-		Labels: []string{"label1", "label2"},
+		Labels: []string{"label1", "label2", `label3: "foo"`},
 		Total:  100,
 		FormatValue: func(v int64) string {
 			return strconv.FormatInt(v, 10)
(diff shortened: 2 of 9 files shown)


internal/graph/dotgraph.go
```diff
@@ -127,7 +128,7 @@ func (b *builder) addLegend() {
 	}
 	title := labels[0]
 	fmt.Fprintf(b, `subgraph cluster_L { "%s" [shape=box fontsize=16`, title)
-	fmt.Fprintf(b, ` label="%s\l"`, strings.Join(labels, `\l`))
+	fmt.Fprintf(b, ` label="%s\l"`, html.EscapeString(strings.Join(labels, `\l`)))
```
aalexand, line 131: It's not obvious to me why `html.EscapeString()` is used.

This code generates a dot file, this has nothing to do with HTML. I'd expect this to adhere to whatever string quoting / special character rules for strings in dot are. And https://graphviz.org/doc/info/lang.html says:

> In quoted strings in DOT, the only escaped character is double-quote ("). That is, in quoted strings, the dyad \" is converted to "; all other characters are left unchanged. In particular, \\ remains \\. Layout engines may apply additional escape sequences.

Plus see escaping rules for escString in https://graphviz.org/doc/info/attrs.html.

So I think we should replace "\" with "\\" and `"` with `\"`, am I missing something?

Please also add a test.

auxiliary: You're right, I was initially under the impression that graphviz runs into issues with non-html characters but this is purely an issue with the dot file. Amended the commit, added a test for the escapeDoubleQuotes function, and added labels with quotes to the test files.

auxiliary: Just noticed a formatting issue in the string, will amend it momentarily.

auxiliary: ...done.

internal/graph/dotgraph_test.go
```diff
@@ -326,6 +326,46 @@ func TestTagCollapse(t *testing.T) {
...
+			want:  `label: \"foo\" and \"bar\"`,
+		},
+		{
+			desc:  "with escaped character but no doubles quotes",
+			input: `label: \$foo and bar`,
+			want:  `label: \$foo and bar`,
+		},
+		{
+			desc:  "with escaped backslash but not escaped doubles quotes",
+			input: `label: \\"foo" and bar`,
```
aalexand, line 352: To clarify this example:

I expect that whatever is specified in the input label string (that is, in the profile comment field) will get visualized verbatim in the final rendered output.

From this point of view I think I'd expect `\\` to become `\\\\` so that in the final rendered version it becomes two backslashes.

Am I missing something?

internal/graph/dotgraph.go
```diff
@@ -472,3 +473,8 @@ func min64(a, b int64) int64 {
 	}
 	return b
 }
+
+func escapeDoubleQuotes(s string) string {
+	re := regexp.MustCompile(`\\([\s\S])|(")`)
```
aalexand, line 478: The regular expression should be compiled once, not on every function call.

aalexand, line 478: Would it work to simply replace `\` with `\\` and `"` with `\"`?

This could be ``strings.ReplaceAll(strings.ReplaceAll(s, `\`, `\\`), `"`, `\"`)``?

internal/graph/dotgraph_test.go
auxiliary: That does make sense, just that the regex that I had was based on the idea of keeping the previous behavior intact. Previously if you had a \\\\ in a comment, it would show up as \\ in the graph.

internal/graph/dotgraph.go
auxiliary: I am not sure this is the best way, since there are occasionally other valid symbols that already have escaping backslashes (e.g. \n, \t), and this messes them up.

aalexand: I'm not sure I follow. If the comment string in a profile contains `\n` then it should be rendered as literal `\n` in pprof CLI output, in rendered graphs, everywhere. There are no escape sequences that are documented as supported in the profile.proto embedded comments.

auxiliary: I think I found what the issue was here. Using your replacement line was eating a new line. The reason is that the code adds \l characters _before_ feeding it to dotgraph.go (see internal/report/report.go:1210). A simple \ replacement wouldn't really work there. Other characters such as \l are supported in graphviz (See https://graphviz.org/doc/info/attrs.html under escString). So the solution is to either make the replacement more aware or make the replacement further back in the code before any postprocessing. The advantage to the regex I had was that it left other special characters intact as well as kept the previous behavior.

aalexand: I see. I would prefer to iterate over labels and do the replacement that I suggested on each element there, before joining with \l. You can avoid the Join() call then. Basically, replace

```
fmt.Fprintf(b, ` label="%s\l"`, escapeDoubleQuotes(strings.Join(labels, `\l`)))
```

with

```
fmt.Fprint(b, ` label="`)
for _, label := range labels {
  fmt.Fprint(b, strings.ReplaceAll(strings.ReplaceAll(label, `\`, `\\`), `"`, `\"`) + `\l`)
}
fmt.Fprint(b, `"`)
```

I think it's clearer. It's harder to reason about the regular expression you used.

auxiliary: Ack. The regex is no longer used.

auxiliary: Gotcha. Done, but in a separate function with separate tests for it. `\n` is also replaced with `\l` per our discussion.

internal/graph/dotgraph.go
```diff
@@ -472,3 +472,10 @@ func min64(a, b int64) int64 {
 	}
 	return b
 }
+
+func escapeChars(s []string) []string {
```
aalexand, line 476: nit: I'd name this escapeForDot and refer to https://graphviz.org/doc/info/attrs.html#k:escString

auxiliary: Done.

internal/graph/dotgraph.go
```diff
@@ -472,3 +472,13 @@ func min64(a, b int64) int64 {
 	}
 	return b
 }
+
+// escapeForDot escapes double quotes and backslashes, and replaces Graphviz's
+// "center" character (\n) with a left-justified character.
+// See https://graphviz.org/doc/info/attrs.html#k:escString for more info.
+func escapeForDot(s []string) []string {
```
aalexand, line 479: Could this function construct and return a new slice? It's current signature suggests that it does that (because it returns a slice) but in fact it changes the input slice in place. The combination of in-place modification and returning a value is imo not good. Also, I can't say just from this local context if it's OK to modify caller's labels slice content, so I'd prefer not modifying its contents.

auxiliary: Yup, makes sense. Done.

aalexand (approved):

PR #559, opened 2020-09-25
Modify Scale() and ScaleN() to round and drop all-zero samples when scaling

nolanmar511:


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/chzyer/logex v1.1.10 // indirect
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
+	github.com/google/go-cmp v0.5.2
```
aalexand, line 9: I wonder if it's OK to add new dependencies in terms of pprof being a part of golang distribution.

aalexand (dismissed):

aalexand (commented):
Looks like CI failed.

go.mod
nolanmar511: Done.

aalexand (dismissed):

profile/profile.go
```diff
@@ -712,7 +713,8 @@ func (s *Sample) DiffBaseSample() bool {
 	return s.HasLabel("pprof::base", "true")
 }
 
-// Scale multiplies all sample values in a profile by a constant.
+// Scale multiplies all sample values in a profile by a constant and keeps
```
kalyanac, line 716: and only keeps samples ...

kalyanac:
Restarted the failed test.

profile/profile.go
nolanmar511: done.

nolanmar511:
I think the two test failures are unrelated to this change; I don' t think re-running will help.

See https://github.com/google/pprof/pull/560#issuecomment-699088881.

aalexand:
@nolanmar511 Interestingly there are still CI failures.

nolanmar511:
Just looking at this. The failures are also happening on the main branch.
I've filed https://github.com/google/pprof/issues/585.

nolanmar511:
Tests are all passing now. PTAL

aalexand (approved):

PR #561, opened 2020-09-26
Test --x86-asm-syntax=intel only on amd64

kalyanac:
Some llvm-objdump installations may not include all backends and will cause failures.

### DIFF
--- a/internal/binutils/binutils_test.go
+++ b/internal/binutils/binutils_test.go
@@ -220,7 +220,10 @@ func TestDisasm(t *testing.T) {
 	if runtime.GOOS != "linux" && runtime.GOOS != "darwin" {
 		t.Skip("This test only works on Linux or Mac")
 	}
-	testDisasm(t, true)
+	
+	if runtime.GOARCH != "amd64" {
+	        testDisasm(t, true)
+	}
 	testDisasm(t, false)
 }
 
--- a/internal/binutils/binutils_test.go
+++ b/internal/binutils/binutils_test.go
@@ -494,7 +494,7 @@ func TestObjdumpVersionChecks(t *testing.T) {
 		},
 		{
 			desc: "Invalid GNU objdump version string",
-			ver:  "GNU objdump (GNU Banutils) 2.34\nCopyright (C) 2020 Free Software Foundation, Inc.",
+			ver:  "GNU nm (GNU Binutils) 2.34\nCopyright (C) 2020 Free Software Foundation, Inc.",
 			want: false,
 		},
 	} {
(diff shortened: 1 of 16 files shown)


kalyanac:
Sorry about the multiple commits. I tried to create the commit directly from Github's editor and that did not go well.

kalyanac:
Test failures are benign and unrelated.

internal/binutils/binutils_test.go
```diff
@@ -220,7 +220,9 @@ func TestDisasm(t *testing.T) {
 	if runtime.GOOS != "linux" && runtime.GOOS != "darwin" {
 		t.Skip("This test only works on Linux or Mac")
 	}
-	testDisasm(t, true)
+	if runtime.GOARCH == "amd64" {
+		testDisasm(t, true)
```
aalexand, line 224: It's preferred to skip tests using t.Skip() so that it's clear that it's skipped and why.

I would split this test into two:

```
func TestDisasm(t *testing.T) {
	if runtime.GOOS != "linux" && runtime.GOOS != "darwin" {
		t.Skip("This test only works on Linux or OSX")
	}
	testDisasm(t, false)
}

func TestDisasmIntelSyntax(t *testing.T) {
	if (runtime.GOOS != "linux" && runtime.GOOS != "darwin") || runtime.GOARCH != "amd64" {
		t.Skip("This test only works on x86_64 Linux or x86_64 OSX as it tests Intel asm syntax")
	}
	testDisasm(t, true)
}
```

kalyanac: Done.

internal/binutils/binutils_test.go
```diff
@@ -220,10 +220,16 @@ func TestDisasm(t *testing.T) {
 	if runtime.GOOS != "linux" && runtime.GOOS != "darwin" {
 		t.Skip("This test only works on Linux or Mac")
 	}
-	testDisasm(t, true)
 	testDisasm(t, false)
 }
 
+func TestDisasmIntelSyntax(t *testing.T) {
+	if (runtime.GOOS != "linux" && runtime.GOOS != "darwin") || runtime.GOARCH != "amd64" {
+		t.Skip("This test only works on x86_64 Linux or x86_64 OSX as it tests Intel asm syntax")
```
aalexand, line 228: nit: reading up a bit, it looks like the right name for the OS on Mac is "macOS" - see https://en.wikipedia.org/wiki/MacOS#macOS. So OSX -> macOS here and Mac -> macOS above.

aalexand (approved):

internal/binutils/binutils_test.go
kalyanac: Done.

kalyanac:
Failure is unrelated.

aalexand (approved):

PR #564, opened 2020-10-03
Add missing `escapeForDot()` to labels for function names

NHDaly:
In some programming languages, e.g. JuliaLang, function names can contain arbitrary characters. For example, in julia, functions with non-identifier characters are represented via the string macro `var"..."`, which allows constructing identifiers that wouldn't otherwise parse.

These names are handled correctly by `pprof` in the FlameGraph view, but before this commit, they would produce an invalid dot file.

This fixes the dot graph export for names that contain `"`.

## Review Guidelines

- All of the code changes are in the `internal/graph/dotgraph.go` file, and consist of escaping a few more callsites that produce text that will go in a string in the `.dot` file.
- The rest of the files are modifications to the tests:
    1. We added a new test to test string escaping in the newly escaped parts of the `.dot` file.
    2. We fixed a bunch of windows tests, which were previously validating incorrect escaping of `\`, and had to be updated once we started correctly escaping `"...\..."` to `"...\\..."` on Windows.

Fixes https://github.com/JuliaPerf/PProf.jl/issues/30.

### DIFF
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -474,7 +474,7 @@ func min64(a, b int64) int64 {
 	return b
 }
 
-// Applies escapeForDot to all strings in the given slice.
+// escapeAllForDot applies escapeForDot to all strings in the given slice.
 func escapeAllForDot(in []string) []string {
 	var out = make([]string, len(in))
 	for i := range in {
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -127,7 +127,7 @@ func (b *builder) addLegend() {
 	}
 	title := labels[0]
 	fmt.Fprintf(b, `subgraph cluster_L { "%s" [shape=box fontsize=16`, title)
-	fmt.Fprintf(b, ` label="%s\l"`, strings.Join(escapeForDot(labels), `\l`))
+	fmt.Fprintf(b, ` label="%s\l"`, strings.Join(escapeAllForDot(labels), `\l`))
 	if b.config.LegendURL != "" {
 		fmt.Fprintf(b, ` URL="%s" target="_blank"`, b.config.LegendURL)
 	}
@@ -187,7 +187,7 @@ func (b *builder) addNode(node *Node, nodeID int, maxFlat float64) {
 
 	// Create DOT attribute for node.
 	attr := fmt.Sprintf(`label="%s" id="node%d" fontsize=%d shape=%s tooltip="%s (%s)" color="%s" fillcolor="%s"`,
-		label, nodeID, fontSize, shape, escapeStringForDot(node.Info.PrintableName()), cumValue,
+		label, nodeID, fontSize, shape, escapeForDot(node.Info.PrintableName()), cumValue,
 		dotColor(float64(node.CumValue())/float64(abs64(b.config.Total)), false),
 		dotColor(float64(node.CumValue())/float64(abs64(b.config.Total)), true))
 
@@ -305,8 +305,8 @@ func (b *builder) addEdge(edge *Edge, from, to int, hasNodelets bool) {
 		arrow = "..."
 	}
 	tooltip := fmt.Sprintf(`"%s %s %s (%s)"`,
-		escapeStringForDot(edge.Src.Info.PrintableName()), arrow,
-		escapeStringForDot(edge.Dest.Info.PrintableName()), w)
+		escapeForDot(edge.Src.Info.PrintableName()), arrow,
+		escapeForDot(edge.Dest.Info.PrintableName()), w)
 	attr = fmt.Sprintf(`%s tooltip=%s labeltooltip=%s`, attr, tooltip, tooltip)
 
 	if edge.Residual {
@@ -383,7 +383,7 @@ func dotColor(score float64, isBackground bool) string {
 
 func multilinePrintableName(info *NodeInfo) string {
 	infoCopy := *info
-	infoCopy.Name = escapeStringForDot(ShortenFunctionName(infoCopy.Name))
+	infoCopy.Name = escapeForDot(ShortenFunctionName(infoCopy.Name))
 	infoCopy.Name = strings.Replace(infoCopy.Name, "::", `\n`, -1)
 	infoCopy.Name = strings.Replace(infoCopy.Name, ".", `\n`, -1)
 	if infoCopy.File != "" {
@@ -474,16 +474,18 @@ func min64(a, b int64) int64 {
 	return b
 }
 
-// escapeForDot escapes double quotes and backslashes, and replaces Graphviz's
-// "center" character (\n) with a left-justified character.
-// See https://graphviz.org/doc/info/attrs.html#k:escString for more info.
-func escapeForDot(in []string) []string {
+// Applies escapeForDot to all strings in the given slice.
+func escapeAllForDot(in []string) []string {
 	var out = make([]string, len(in))
 	for i := range in {
-		out[i] = escapeStringForDot(in[i])
+		out[i] = escapeForDot(in[i])
 	}
 	return out
 }
-func escapeStringForDot(str string) string {
+
+// escapeForDot escapes double quotes and backslashes, and replaces Graphviz's
+// "center" character (\n) with a left-justified character.
+// See https://graphviz.org/doc/info/attrs.html#k:escString for more info.
+func escapeForDot(str string) string {
 	return strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(str, `\`, `\\`), `"`, `\"`), "\n", `\l`)
 }
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -187,7 +187,7 @@ func (b *builder) addNode(node *Node, nodeID int, maxFlat float64) {
 
 	// Create DOT attribute for node.
 	attr := fmt.Sprintf(`label="%s" id="node%d" fontsize=%d shape=%s tooltip="%s (%s)" color="%s" fillcolor="%s"`,
-		label, nodeID, fontSize, shape, node.Info.PrintableName(), cumValue,
+		escapeForDot(label), nodeID, fontSize, shape, node.Info.PrintableName(), cumValue,
 		dotColor(float64(node.CumValue())/float64(abs64(b.config.Total)), false),
 		dotColor(float64(node.CumValue())/float64(abs64(b.config.Total)), true))
 
@@ -247,7 +247,7 @@ func (b *builder) addNodelets(node *Node, nodeID int) bool {
 			continue
 		}
 		weight := b.config.FormatValue(w)
-		nodelets += fmt.Sprintf(`N%d_%d [label = "%s" id="N%d_%d" fontsize=8 shape=box3d tooltip="%s"]`+"\n", nodeID, i, t.Name, nodeID, i, weight)
+		nodelets += fmt.Sprintf(`N%d_%d [label = "%s" id="N%d_%d" fontsize=8 shape=box3d tooltip="%s"]`+"\n", nodeID, i, escapeForDot(t.Name), nodeID, i, weight)
 		nodelets += fmt.Sprintf(`N%d -> N%d_%d [label=" %s" weight=100 tooltip="%s" labeltooltip="%s"]`+"\n", nodeID, nodeID, i, weight, weight, weight)
 		if nts := lnts[t.Name]; nts != nil {
 			nodelets += b.numericNodelets(nts, maxNodelets, flatTags, fmt.Sprintf(`N%d_%d`, nodeID, i))
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -149,7 +149,6 @@ func (b *builder) addNode(node *Node, nodeID int, maxFlat float64) {
 	} else {
 		label = multilinePrintableName(&node.Info)
 	}
-	// label = strings.Replace(label, `"`, `\"`, -1)
 
 	flatValue := b.config.FormatValue(flat)
 	if flat != 0 {
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -149,6 +149,7 @@ func (b *builder) addNode(node *Node, nodeID int, maxFlat float64) {
 	} else {
 		label = multilinePrintableName(&node.Info)
 	}
+	// label = strings.Replace(label, `"`, `\"`, -1)
 
 	flatValue := b.config.FormatValue(flat)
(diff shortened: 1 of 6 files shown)


NHDaly:
Hello! 👋 I believe this is my first contribution to `pprof`. Please excuse my sloppy `go` abilities. I learned a small bit of Go when I used to work at google a few years ago, but I never learned much.

I am one of the maintainers of [PProf.jl](https://github.com/JuliaPerf/PProf.jl), the `julia` bindings to `pprof`. We love this package, and make heavy use of it at my company, and in the wider julia community.

However, as described above, julia function names can be quite messy, and I encountered a case where `pprof` is failing to escape names that contain a string.

I've tested the above fix locally, and the tests pass, but it appears that some of the build tests have failed. Help here would be very much appreciated.

Thanks much!
Best,
~Nathan

NHDaly:
Based on the ci output, it looks like the `test-mac` job is failing to build for unrelated reasons?

Otherwise, I think the tests should be passing now. :) Please take a look when you're available. Thanks!

internal/graph/dotgraph.go
```diff
@@ -479,7 +480,10 @@ func min64(a, b int64) int64 {
...
 	var out = make([]string, len(in))
 	for i := range in {
-		out[i] = strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(in[i], `\`, `\\`), `"`, `\"`), "\n", `\l`)
+		out[i] = escapeStringForDot(in[i])
 	}
 	return out
 }
+func escapeStringForDot(str string) string {
+	return strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(str, `\`, `\\`), `"`, `\"`), "\n", `\l`)
+}
```
NHDaly, line 489: I wasn't quite sure what to call this function. I needed another version of this which works on a single string at a time.

Very open to suggestions/requests if you have any other ideas for what to call this! :) Thanks.

internal/graph/dotgraph.go
```diff
@@ -479,7 +480,10 @@ func min64(a, b int64) int64 {
 func escapeForDot(in []string) []string {
 	var out = make([]string, len(in))
 	for i := range in {
-		out[i] = strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(in[i], `\`, `\\`), `"`, `\"`), "\n", `\l`)
+		out[i] = escapeStringForDot(in[i])
 	}
 	return out
 }
+func escapeStringForDot(str string) string {
```
aalexand, line 487: I would call this function `escapeForDot` and rename the one above to `escapeAllForDot`.

Also, the comment from the function above should move to here since this is where it applies. And for the function above add a comment along the lines "applies escapeForDot to all strings in the given slice".

aalexand (commented):
I am curious on what the code coverage dropped so substantially.

internal/graph/dotgraph.go
NHDaly: Thanks @aalexand that's a good suggestion. I'll do that tomorrow! :)

NHDaly:
> I am curious on what the code coverage dropped so substantially.

Yeah, it's really weird, right? I saw that too, and it led me to believe the coverage tool must be flaky. Do you find it to be usually quite reliable?

I added a test case that I believe covers the cases I've modified the code for. But maybe I messed something else up somehow? I would appreciate guidance here since I'm not a regular go programmer. 🙈 Thanks for your help so far! :)

internal/graph/dotgraph.go
aalexand: Thanks - please ping the PR once that is done.

NHDaly: Thanks @aalexand - this is done! :)

NHDaly (commented):
This is again ready for review.
I have applied the renaming we suggested.

Also, as long as I was looking at this, I've also added escaping for Tags, which actually proved to be a bit more difficult, since `graph.go` was already doing some manual escaping of `\n`, which wound up double escaped.

I've sorted that all out, but since it's a fairly sizeable PR as well, I've opened it as a separate PR against this branch, here:
https://github.com/NHDaly/pprof/pull/1

Please take a look at that PR as well if you get a chance, and we can either merge it directly into this branch first if you prefer merging a single PR, or we can take them one at a time. It's up to you. Thanks! 

(I tried opening it directly against this repo, but then i couldn't base it on top of this PR. GitHub still has some rough edges...)

internal/graph/dotgraph_test.go
```diff
@@ -138,6 +138,22 @@ func TestComposeWithStandardGraphAndURL(t *testing.T) {
 	compareGraphs(t, buf.Bytes(), "compose6.dot")
 }
 
+func TestComposeWithNamesThatNeedEscaping(t *testing.T) {
+	g := baseGraph()
+	a, c := baseAttrsAndConfig()
+
+	g.Nodes[0].Info = NodeInfo{Name: "var\"src\""}
```
aalexand, line 145: Use backtick strings.

internal/graph/dotgraph_test.go
```diff
@@ -138,6 +138,22 @@ func TestComposeWithStandardGraphAndURL(t *testing.T) {
...
+	g.Nodes[1].Info = NodeInfo{Name: "var\"#dest#\""}
+
+	// Set edge to be Residual.
```
aalexand, line 148: Comment on the "why", not about the "what". It's clear that the line below sets the residual to true, it's unclear why.

internal/graph/dotgraph_test.go
```diff
@@ -138,6 +138,22 @@ func TestComposeWithStandardGraphAndURL(t *testing.T) {
 	compareGraphs(t, buf.Bytes(), "compose6.dot")
 }
 
+func TestComposeWithNamesThatNeedEscaping(t *testing.T) {
+	g := baseGraph()
+	a, c := baseAttrsAndConfig()
+
```
aalexand, line 144: Remove this empty line.

internal/graph/dotgraph_test.go
```diff
@@ -138,6 +138,22 @@ func TestComposeWithStandardGraphAndURL(t *testing.T) {
...
+	g.Nodes[0].Info = NodeInfo{Name: "var\"src\""}
+	g.Nodes[1].Info = NodeInfo{Name: "var\"#dest#\""}
+
```
aalexand, line 147: Remove this empty line.

internal/graph/dotgraph_test.go
NHDaly: Thanks! I'm not sure "why" either, unfortunately! 😭 

I copied this test verbatim from here, and then adapted it for what i wanted to test:
https://github.com/google/pprof/blob/dff1afcd898c3bc6799f02ca45da17ba49dea69c/internal/graph/dotgraph_test.go#L67-L93

Do you know why they set the edge to be Residual? do you know what it means and/or what it does? 😬 

I think i tried deleting it but the test failed, but i don't remember why.

internal/graph/dotgraph_test.go
NHDaly:
```suggestion
```

internal/graph/dotgraph_test.go
NHDaly:
```suggestion
	g.Nodes[0].Info = NodeInfo{Name: `var"src"`}
```

internal/graph/dotgraph_test.go
```diff
@@ -138,6 +138,22 @@ func TestComposeWithStandardGraphAndURL(t *testing.T) {
 	compareGraphs(t, buf.Bytes(), "compose6.dot")
 }
 
+func TestComposeWithNamesThatNeedEscaping(t *testing.T) {
+	g := baseGraph()
+	a, c := baseAttrsAndConfig()
+
+	g.Nodes[0].Info = NodeInfo{Name: "var\"src\""}
+	g.Nodes[1].Info = NodeInfo{Name: "var\"#dest#\""}
```
NHDaly, line 146:
```suggestion
	g.Nodes[1].Info = NodeInfo{Name: `var"#dest#"`}
```

internal/graph/dotgraph_test.go
NHDaly:
```suggestion
```

aalexand:
Two Windows tests are failing.

internal/graph/dotgraph_test.go
aalexand: No, I don't know why off top of my head. What was the failure without it?

NHDaly:
> Two Windows tests are failing.

Ugh, thanks.

It appears that the failure is because now we're escaping the backslashes in windows paths, e.g. from  `testdata\source1` to `testdata\\source1`.

I'm very unsure what to do about this.
Is the original `\` even correct? Is it okay to have a single standalone `\` in a string like that, even on windows? Maybe this is actually fixing a bug?

If so, I guess we can just fix the test to expect the new behavior instead of finding some way to disable it on windows.

---------------------

Also, it looks like the coverage report is back to normal, so that's good! :)

internal/graph/dotgraph_test.go
NHDaly: Hmm, i dunno! It just changes what the edge prints like:
```diff
        -N1 -> N2 [label=" 10" weight=11 color="#b28559" tooltip="var\"src\" -> var\"#dest#\" (10)" labeltooltip="var\"src\" -> var\"#dest#\" (10)"]
        +N1 -> N2 [label=" 10" weight=11 color="#b28559" tooltip="var\"src\" ... var\"#dest#\" (10)" labeltooltip="var\"src\" ... var\"#dest#\" (10)" style="dotted"]
```

But hmm, yeah, i don't think this is important to our test at all, so maybe i'll just delete it, and change the expected graph to match whatever it's printing? :)

Or i can keep it and add a comment along the lines of `// We mark this edge residual to follow the same behavior as the other tests. We don't know if it's needed, but this test is only focused on the string escaping.` 😅 

I'm supportive of deleting the line and changing the expected result if that works for you!

aalexand:
> > Two Windows tests are failing.
> 
> Ugh, thanks.
> 
> It appears that the failure is because now we're escaping the backslashes in windows paths, e.g. from `testdata\source1` to `testdata\\source1`.
> 
> I'm very unsure what to do about this.
> Is the original `\` even correct? Is it okay to have a single standalone `\` in a string like that, even on windows? Maybe this is actually fixing a bug?
> 
> If so, I guess we can just fix the test to expect the new behavior instead of finding some way to disable it on windows.
> 
> Also, it looks like the coverage report is back to normal, so that's good! :)

I would expect that double slashes should be escaped in the *.dot file strings, yes.

internal/graph/dotgraph_test.go
aalexand: Unfortunately, I won't have time to figure this out but I'm also a bit hesitant to either put partial or "we do this but we don't know why" comment. So you can either wait till I have time to look at it or you need to dive a bit more into this and figure this out. It is a bit unfair to ask you to block on this since you are using a copy of a pre-existing comment, but I prefer to use every PR to improve the code a tiny bit. Sorry about this!

NHDaly:
> I would expect that double slashes should be escaped in the *.dot file strings, yes.


Awesome, thanks!

NHDaly:
Okay! :) I believe I have fixed the Windows test, by adjusting the test to expect the properly escaped `\` (i.e. `\\`) in the output, in 4182efa1c7d4dc8dda5aab546fd87114f273090b.

We'll see what the tests say.

I'm sorry for my delay! Also, thanks for merging in master into the branch a few times! :)

internal/graph/dotgraph_test.go
NHDaly: Done. Fixed in 060ba7359877a1faf326f1637ab8f802ecb95f6d.

It turns out that this was just because I was copying from a test that was _specifically testing_ that Residual edges work as expected. This was unnecessary info for my test, so i've deleted that line and adjusted the expected output accordingly. 👍 Thanks! :)

NHDaly:
@aalexand - I have (finally) fixed all the remaining windows test failures! :)

If you could, would you please take another look when you get the chance? Thanks very much!
Best,
~Nathan

NHDaly:
Friendly ping: this is now ready for final review :) Thanks again for all the help so far!

aalexand:
@NHDaly I'll take a look at it by EOW, sorry for the delay.

NHDaly:
:) Thanks, no worries at all! :) I just wanted to be sure this stays on your radar 👍

internal/graph/dotgraph.go
```diff
@@ -473,13 +474,18 @@ func min64(a, b int64) int64 {
 	return b
 }
 
-// escapeForDot escapes double quotes and backslashes, and replaces Graphviz's
-// "center" character (\n) with a left-justified character.
-// See https://graphviz.org/doc/info/attrs.html#k:escString for more info.
-func escapeForDot(in []string) []string {
+// Applies escapeForDot to all strings in the given slice.
+func escapeAllForDot(in []string) []string {
```
Segflow, line 478: You don’t have to comment unexported methods, but when you do, it’s better if you follow the convention: start the comment with the name of thing it describes (https://blog.golang.org/godoc)
So the above code can be improved by changing the comment to:
```
escapeAllForDot applies escapeForDot to all strings in the given slice.
```

NHDaly, line 478:
```suggestion
// escapeAllForDot applies escapeForDot to all strings in the given slice.
func escapeAllForDot(in []string) []string {
```

NHDaly: Fixed in fc32837. Thanks! :)

Also, I marked that commit as `Co-Authored-By: @Segflow`, which seems to have triggered this Google authorship bot.. 😅 Sorry about that. lol

NHDaly:
@googlebot I fixed it.

internal/graph/dotgraph.go
NHDaly: Thanks. Done! :)

Segflow (approved):

NHDaly:
Thanks for the review, @Segflow! :)

Friendly ping @aalexand - when you get the chance this should be all set to merge. I think it's a simple review now; we're just escaping a couple new strings that weren't being escaped before 😁 

Thanks!

NHDaly:
(oh, whoops, sorry, I hadn't re-requested a Review from you through the GitHub UI. I've done that now! 👍)

NHDaly:
i've also updated the top-level comment with Review Guidelines that explain the changes in this PR and will hopefully make it easier to review. Thanks!! :)

aalexand (approved):

NHDaly:
Awesome. Thank you very much for your help and your time! :)

PR #565, opened 2020-10-04
Fix a race in temp file creation.

### LINKED ISSUE #500: With multiple input perf.data files, some files are recognized as either empty input files or unrecognized profile format
When I run pprof with multiple perf.data files (e.g. pprof -top *.data), some files will be randomly recognized as empty input files or in unrecognized profile format. 
However, when I run pprof with those "wrong" files, it can successfully generate report. 

[pprof.log](https://github.com/google/pprof/files/3866898/pprof.log)

aalexand:
The added test fails before the fix and passes after.

Fixes #500.

### DIFF
--- a/internal/driver/tempfile.go
+++ b/internal/driver/tempfile.go
@@ -24,9 +24,11 @@ import (
 // newTempFile returns a new output file in dir with the provided prefix and suffix.
 func newTempFile(dir, prefix, suffix string) (*os.File, error) {
 	for index := 1; index < 10000; index++ {
-		path := filepath.Join(dir, fmt.Sprintf("%s%03d%s", prefix, index, suffix))
-		if _, err := os.Stat(path); err != nil {
-			return os.Create(path)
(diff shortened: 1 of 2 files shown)


nolanmar511 (approved):

internal/driver/tempfile.go
```diff
@@ -25,9 +25,21 @@ import (
...
-		if _, err := os.Stat(path); err != nil {
-			return os.Create(path)
+		_, err := os.Stat(path)
+		if err == nil {
+			continue
 		}
+		if !os.IsNotExist(err) {
+			return nil, err
+		}
+		f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
```
kalyanac, line 35: I dont think Stat() is needed. Just call OpenFile with `os.O_RDWR|os.O_CREATE|os.O_EXCL` directly and handle err as needed. If this is racy, checking Stat() prior to OpenFile does not guarantee success in all cases.

aalexand: Done. I thought about this and left the `os.Stat()` call to avoid spurious write accesses - somehow it felt better to only probe with a read rather than with a write. But it shouldn't matter as far as I can think of it, and less code is good.

kalyanac (approved):

PR #568, opened 2020-10-05
Escape Nodelet/tag names for Dot as well

NHDaly:
This escaping requirement was previously leaking into the implementation
of `graph.go`, by writing `\n` directly instead of "\n". This meant that
if it was displayed in a different program besides `dot`, it would have
been potentially incorrectly escaped.

Now, this commit moves all the escaping directly into dotgraph, which is
more encapsulated, and also allows us to escape other characters in the
name which need escaping.

This required adding a justify option to escapeForDot to preserve old behavior.
This allows tag newlines to stay center justified, and legend label
newlines to stay left justified, as they were before we started escaping
them for Dot.

--

This was meant to be opened as a follow-up PR to https://github.com/google/pprof/pull/564, but apparently I can't adjust it to use a fork's branch as a diffbase. that's disappointing. I will close this in favor of https://github.com/NHDaly/pprof/pull/1 for now.

### DIFF
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -127,7 +127,7 @@ func (b *builder) addLegend() {
 	}
 	title := labels[0]
 	fmt.Fprintf(b, `subgraph cluster_L { "%s" [shape=box fontsize=16`, title)
-	fmt.Fprintf(b, ` label="%s\l"`, strings.Join(escapeAllForDot(labels, LeftJustify), `\l`))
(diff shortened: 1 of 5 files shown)


NHDaly:
This was meant to be opened as a follow-up PR to #564, but apparently I can't adjust it to use a fork's branch as a diffbase. that's disappointing. I will close this in favor of NHDaly#1 for now.

NHDaly:
i've just realized that i left this PR unmerged and forgot about it after merging https://github.com/google/pprof/pull/564...

I'm reopening it as a reminder to myself to look into it, and see whether there's still anything here we need to do.. thanks

PR #571, opened 2020-10-09
Remove potential Go module versions from shortened names

### LINKED ISSUE #515: Go module version suffix shown in graphs instead of package name
Moved from golang/go#37825.

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/jake/.cache/go-build"
GOENV="/home/jake/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/jake/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/jake/zikaeroh/hortbot/hortbot/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build332050882=/tmp/go-build -gno-record-gcc-switches"
GOROOT/bin/go version: go version go1.14 linux/amd64
GOROOT/bin/go tool compile -V: compile version go1.14
uname -sr: Linux 5.5.7-zen1-1-zen
/usr/lib/libc.so.6: GNU C Library (GNU libc) stable release version 2.31.
gdb --version: GNU gdb (GDB) 9.1
</pre></details>

Linux, amd64.

### What did you do?

Created a profile, and opened it with `go tool pprof` to view a graph of memory allocations.

Nodes with package names.

### What did you see instead?

Packages which are at the root of a non-v1 module show their version suffix rather than their actual package name, making it more difficult to determine what they are.

For example, jackc's pgx/pgconn, which are at v4 and v2 respectively:

image

It looks like `pprof` just takes the last element of the import path, which used to be somewhat accurate (ignoring path/package name differences), but is less so now.

zikaeroh:
Fixes #515.

Remove potential module path versions (v2, v3, v4, etc) from the input string before extracting a shortened name.  This makes it easier to tell which packages are which if the versions happen to match.

An example similar to my report:

image

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -35,7 +35,7 @@ var (
 	// See tests for examples.
 	goRegExp = regexp.MustCompile(`^(?:[\w\-\.]+\/)+(.+)`)
 	// Checks for a package name that could be a module version.
-	goVerRegExp = regexp.MustCompile(`^v[2-9]+\.`)
+	goVerRegExp = regexp.MustCompile(`^v([2-9]|[1-9][0-9]+)\.`)
 	// Strips C++ namespace prefix from a C++ function / method name.
 	// NOTE: Make sure to keep the template parameters in the name. Normally,
 	// template parameters are stripped from the C++ names but when
@@ -467,7 +467,6 @@ func shortenGoFunc(f string, name string) string {
 			end = idx
 		}
 	}
-
 	return f[end+1:]
 }
 
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -35,7 +35,7 @@ var (
 	// See tests for examples.
 	goRegExp = regexp.MustCompile(`^(?:[\w\-\.]+\/)+(.+)`)
 	// Removes potential module versions in a package path.
-	goVerRegExp = regexp.MustCompile(`^(.*)/v(?:[2-9]|[1-9][0-9]+)([./].*)$`)
+	goVerRegExp = regexp.MustCompile(`^(.*?)/v(?:[2-9]|[1-9][0-9]+)([./].*)$`)
 	// Strips C++ namespace prefix from a C++ function / method name.
 	// NOTE: Make sure to keep the template parameters in the name. Normally,
 	// template parameters are stripped from the C++ names but when
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -34,6 +34,8 @@ var (
 	// Removes package name and method arugments for Go function names.
 	// See tests for examples.
 	goRegExp = regexp.MustCompile(`^(?:[\w\-\.]+\/)+(.+)`)
+	// Checks for a package name that could be a module version.
+	goVerRegExp = regexp.MustCompile(`^v[2-9]+\.`)
 	// Strips C++ namespace prefix from a C++ function / method name.
 	// NOTE: Make sure to keep the template parameters in the name. Normally,
 	// template parameters are stripped from the C++ names but when
@@ -442,10 +444,31 @@ func ShortenFunctionName(f string) string {
 	f = cppAnonymousPrefixRegExp.ReplaceAllString(f, "")
 	for _, re := range []*regexp.Regexp{goRegExp, javaRegExp, cppRegExp} {
 		if matches := re.FindStringSubmatch(f); len(matches) >= 2 {
-			return strings.Join(matches[1:], "")
+			name := strings.Join(matches[1:], "")
+			if re == goRegExp {
+				return shortenGoFunc(f, name)
+			}
+			return name
 		}
 	}
 	return f
+}
+
+func shortenGoFunc(f string, name string) string {
+	if !goVerRegExp.MatchString(name) {
+		return name
+	}
+
+	// The shortened name could start with a module version (like "v2"). Go back one slash.
+	end := len(f) - len(name) - 1
+	if end >= 0 {
+		prefix := f[:end]
+		if idx := strings.LastIndex(prefix, "/"); idx >= 0 {
+			end = idx
+		}
+	}
+
+	return f[end+1:]
 }
 
 // TrimTree trims a Graph in forest form, keeping only the nodes in kept. This
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -459,7 +459,8 @@ func shortenGoFunc(f string, name string) string {
 		return name
 	}
 
-	// The shortened name could start with a module version (like "v2"). Go back one slash.
+	// The shortened name could start with a module version (like "v2"). Go
+	// back one slash.
 	end := len(f) - len(name) - 1
 	if end >= 0 {
 		prefix := f[:end]
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -34,8 +34,8 @@ var (
 	// Removes package name and method arugments for Go function names.
 	// See tests for examples.
 	goRegExp = regexp.MustCompile(`^(?:[\w\-\.]+\/)+(.+)`)
-	// Checks for a package name that could be a module version.
-	goVerRegExp = regexp.MustCompile(`^v([2-9]|[1-9][0-9]+)\.`)
+	// Removes potential module versions in a package path.
+	goVerRegExp = regexp.MustCompile(`^(.*)/v(?:[2-9]|[1-9][0-9]+)([./].*)$`)
 	// Strips C++ namespace prefix from a C++ function / method name.
 	// NOTE: Make sure to keep the template parameters in the name. Normally,
 	// template parameters are stripped from the C++ names but when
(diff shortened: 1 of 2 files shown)


internal/graph/graph.go
```diff
@@ -442,12 +444,33 @@ func ShortenFunctionName(f string) string {
...
+
+	// The shortened name could start with a module version (like "v2"). Go back one slash.
+	end := len(f) - len(name) - 1
+	if end >= 0 {
+		prefix := f[:end]
+		if idx := strings.LastIndex(prefix, "/"); idx >= 0 {
+			end = idx
+		}
+	}
+
```
nolanmar511, line 470: Perhaps remove this line break.

nolanmar511 (commented):
Thank you for this PR!

internal/graph/graph.go
```diff
@@ -34,6 +34,8 @@ var (
 	// Removes package name and method arugments for Go function names.
 	// See tests for examples.
 	goRegExp = regexp.MustCompile(`^(?:[\w\-\.]+\/)+(.+)`)
+	// Checks for a package name that could be a module version.
+	goVerRegExp = regexp.MustCompile(`^v[2-9]+\.`)
```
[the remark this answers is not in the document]
zikaeroh: Oh, you're right, I brainfarted on that one. Should have been `^v([2-9]|[0-9]+)\.`.

internal/graph/graph_test.go
```diff
@@ -451,6 +451,18 @@ func TestShortenFunctionName(t *testing.T) {
 			"github.com/blah/blah/vendor/gopkg.in/redis.v3.(*baseClient).(github.com/blah/blah/vendor/gopkg.in/redis.v3.process)-fm",
 			"redis.v3.(*baseClient).(github.com/blah/blah/vendor/gopkg.in/redis.v3.process)-fm",
 		},
+		{
+			"github.com/jackc/pgx/v4.(*Conn).Query",
```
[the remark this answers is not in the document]
zikaeroh: Sure thing.

internal/graph/graph.go
zikaeroh: Sure, if you want. I view the above block as calculating `end`, so that was my thinking.

internal/graph/graph.go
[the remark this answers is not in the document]
zikaeroh: I switched it to a corrected regex, but if there's some more complicated one that works better it can be used. I know my brain shuts down trying to read that long one... :slightly_smiling_face:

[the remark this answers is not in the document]
zikaeroh: The key thing is to allow v2, v3, v10, v1234, etc, but not v0 or v1.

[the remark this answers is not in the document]
zikaeroh:
```
(?:^(?:[\w\-\.]+\/)+((?:[\w\-\.]+\/v(?:[2-9]|[1-9][0-9]+)+)(?:\.[^.\n]+){2})$)|(?:^(?:[\w\-\.]+\/)+(.+))
```

Is the long one with the restricted version select; that'd be a one line change and it appears to pass the tests. I'm happy to use it instead.

internal/graph/graph_test.go
```diff
@@ -451,6 +451,30 @@ func TestShortenFunctionName(t *testing.T) {
 			"github.com/blah/blah/vendor/gopkg.in/redis.v3.(*baseClient).(github.com/blah/blah/vendor/gopkg.in/redis.v3.process)-fm",
 			"redis.v3.(*baseClient).(github.com/blah/blah/vendor/gopkg.in/redis.v3.process)-fm",
 		},
+		{
+			"github.com/foo/bar/v4.(*Foo).Bar",
```
aalexand, line 455: Curious - why is version string sometimes a separate subdirectory and sometimes a prefix of the package name? Is this something that the package owners choose? Are these options restricted at these two, or are there more?

Oh, I guess it's a function of how deep below the versioning level the actual symbol is?

internal/graph/graph.go
```diff
@@ -442,12 +444,32 @@ func ShortenFunctionName(f string) string {
...
 	}
 	return f
 }
 
+func shortenGoFunc(f string, name string) string {
+	if !goVerRegExp.MatchString(name) {
+		return name
+	}
+
+	// The shortened name could start with a module version (like "v2"). Go back one slash.
```
aalexand, line 462: Keep comments in 80 columns please.

internal/graph/graph.go
```diff
@@ -442,12 +444,32 @@ func ShortenFunctionName(f string) string {
 	f = cppAnonymousPrefixRegExp.ReplaceAllString(f, "")
 	for _, re := range []*regexp.Regexp{goRegExp, javaRegExp, cppRegExp} {
 		if matches := re.FindStringSubmatch(f); len(matches) >= 2 {
-			return strings.Join(matches[1:], "")
+			name := strings.Join(matches[1:], "")
+			if re == goRegExp {
+				return shortenGoFunc(f, name)
```
aalexand, line 449: I feel it might be simpler to first remove the version substring from the name, and then handle it just like before.

internal/graph/graph_test.go
zikaeroh: I think you're referring to the tests; some of the tests I've added are where the "version" isn't a version at all. The only valid version in paths are "v2", "v3", ... "v1234", etc. So you'd have `github.com/foo/bar`, `github.com/foo/bar/v2`, `github.com/foo/bar/v3`, and so on, then subpackages like `github.com/foo/bar/v3/baz`. Custom domains mean you can have things like `gotest.tools/assert`, `gotest.tools/v3/assert`. But a package can be at the level where the version appears, so when `github.com/jackc/pgx` was bumped to `github.com/jackc/pgx/v4`, it's still referred to as `pgx` in the code.

But if it isn't a valid version part, then I don't want to treat it as one naively (i.e. "something.com/hello/v123xyz" isn't versioned, "something.com/hello/v123/xyz" is because the version is its own element).

internal/graph/graph.go
zikaeroh: Not sure if you saw the previous review comments, but if preferred this all can be removed and replaced with a single regex change.

internal/graph/graph.go
zikaeroh: Sure.

internal/graph/graph.go
zikaeroh: > I feel it might be simpler to first remove the version substring from the name, and then handle it just like before.

I'm not sure how this is possible; the name here is extracted from the regex directly. If we remove the version suffix, you get the empty string.

zikaeroh: Oh, I see, you mean that if the name matches a version, remove the suffix and then try again. It wouldn't distinguish two versions of the same module, but I guess it's no worse than any other name aliasing within the same graph.

zikaeroh: I think I misinterpreted again (sorry!), so I'll wait for clarification.

I think you meant just starting with something like `github.com/jackc/pgx/v4/foo.bar`, then replacing the first instance of `/v4/` with `/`, then running the regex again. Not quite a suffix, but functional enough.

aalexand: Yes, I meant first remove all occurrences of `/[1-9][0-9]*[./]`, then business as usual.

aalexand: Sorry, `/v[1-9][0-9]*[./]`.

aalexand: And not all occurrences but at most one occurrence (assuming there can't be two version substrings in the name).

zikaeroh: All instances wouldn't work, because it's legal for me to write `github.com/foo/bar/v4/something/v8` or similar. First instance I believe would work as intended, though.

zikaeroh: I'll give it a try. Would be `v([2-9]|[1-9][0-9]+)\.`, though, as `v0` and `v1` don't exist.

zikaeroh: Done, and retitled to match the new fix. `regexp` doesn't have a nice "just replace once", so I used the replace all with two capture groups method.

aalexand (dismissed):

aalexand (approved):

PR #576, opened 2020-10-23
feat: reduce map access

orisano:
I reduced map access to improve newGraph's performance.

```
name        old time/op    new time/op    delta
NewGraph-8     1.37s ± 4%     1.17s ± 5%  -14.89%  (p=0.000 n=10+10)

name        old alloc/op   new alloc/op   delta
NewGraph-8    47.8MB ± 3%    46.9MB ± 2%     ~     (p=0.075 n=10+10)

name        old allocs/op  new allocs/op  delta
NewGraph-8      555k ± 3%      548k ± 3%     ~     (p=0.190 n=10+10)
```

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -126,11 +126,11 @@ func (n *Node) AddToEdge(to *Node, v int64, residual, inline bool) {
 // AddToEdgeDiv increases the weight of an edge between two nodes. If
 // there isn't such an edge one is created.
 func (n *Node) AddToEdgeDiv(to *Node, dv, v int64, residual, inline bool) {
-	if n.Out[to] != to.In[n] {
+	e := n.Out[to]
+	if e != to.In[n] {
 		panic(fmt.Errorf("asymmetric edges %v %v", *n, *to))
 	}
-
-	if e := n.Out[to]; e != nil {
+	if e != nil {
 		e.WeightDiv += dv
 		e.Weight += v
 		if residual {
@@ -319,8 +319,8 @@ func New(prof *profile.Profile, o *Options) *Graph {
 // nodes.
 func newGraph(prof *profile.Profile, o *Options) (*Graph, map[uint64]Nodes) {
 	nodes, locationMap := CreateNodes(prof, o)
-	seenNode := make(map[*Node]bool)
-	seenEdge := make(map[nodePair]bool)
+	seenNode := make(map[*Node]struct{})
+	seenEdge := make(map[nodePair]struct{})
 	for _, sample := range prof.Sample {
 		var w, dw int64
 		w = o.SampleValue(sample.Value)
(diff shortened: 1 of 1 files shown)


internal/graph/graph.go
```diff
@@ -352,14 +352,16 @@ func newGraph(prof *profile.Profile, o *Options) (*Graph, map[uint64]Nodes) {
 					continue
 				}
 				// Add cum weight to all nodes in stack, avoiding double counting.
-				if _, ok := seenNode[n]; !ok {
-					seenNode[n] = true
+				seenNodeLen := len(seenNode)
+				if seenNode[n] = struct{}{}; seenNodeLen != len(seenNode) {
```
aalexand, line 356: What is the speedup from this particular change? This change modifies the code from a very obvious pattern to a rather subtle trick IMO - so I'd rather leave the code as is if the effect is not significant. Same about the edge map.

orisano: This change is an important part of this PR.

In the original code, `runtime.mapaccess2_fast64` and `runtime.mapassign_fast64ptr` are called.
Uses `runtime.mapaccess2_fast64` to check if an element is already present, but it is expensive.
It can find out if an element is already present by getting the size of the map in advance and checking the size of the map after `runtime.mapassign`.
This avoids unnecessary `runtime.mapaccess2_fast64` and improves performance.

aalexand: Can you include numbers to clarify by how much each change improves performance?

I'd prefer code to stay clearer if the performance win is marginal.

It would also be good to add a benchmark for the graph creation and use https://godoc.org/golang.org/x/perf/cmd/benchstat to demonstrate the net effect.

orisano: Benchmark is below (in driver_test.go).
```go
func BenchmarkNewGraph(b *testing.B) {
	f, err := os.Open("testdata/a.pb.gz")
	if err != nil {
		b.Fatal(err)
	}
	defer f.Close()
	p, err := profile.Parse(f)
	if err != nil {
		b.Fatal(err)
	}
	cfg := defaultConfig()
	options := &plugin.Options{
		UI:  &proftest.TestUI{},
		Obj: fakeObjTool{},
	}
	_, r, err := generateRawReport(p, []string{"svg"}, cfg, options)
	if err != nil {
		b.Fatal(err)
	}
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		report.GetDOT(r)
	}
}
```

benchstat's result (excludes the change of AddToEdgeDiv):
```
name        old time/op    new time/op    delta
NewGraph-8     1.41s ± 3%     1.26s ± 3%  -10.73%  (p=0.000 n=10+10)

name        old alloc/op   new alloc/op   delta
NewGraph-8    47.6MB ± 3%    47.1MB ± 2%     ~     (p=0.280 n=10+10)

name        old allocs/op  new allocs/op  delta
NewGraph-8      552k ± 3%      550k ± 3%     ~     (p=0.739 n=10+10)
```

benchstat's result:
```
name        old time/op    new time/op    delta
NewGraph-8     1.41s ± 3%     1.20s ± 2%  -14.96%  (p=0.000 n=10+9)

name        old alloc/op   new alloc/op   delta
NewGraph-8    47.6MB ± 3%    46.7MB ± 2%   -1.84%  (p=0.028 n=10+9)

name        old allocs/op  new allocs/op  delta
NewGraph-8      552k ± 3%      545k ± 2%     ~     (p=0.113 n=10+9)
```

but, `testdata/a.pb.gz` is a private profile(includes 40000 samples) from our production environment.
it is hard to publish.

I close this PR.

PR #579, opened 2020-11-13
Update supported Xcode versions to fix Github Action failures.

wyk9787:
More specifically, we drop support for Xcode 11.0, 11.1 and add support for Xcode 12.1

Github Action supports the following Xcode versions: https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#xcode

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         go: ['1.14', '1.15']
         os: ['macos-10.15']
-        xcode-version: ['12.1', '12.0.1', '11.7', '11.6', '11.5', '11.4.1', '11.3.1', '11.2.1']
+        xcode-version: ['12.1', '12.0', '11.7', '11.6', '11.5', '11.4.1', '11.3.1', '11.2.1']
     steps:
       - name: Update go version
         uses: actions/setup-go@v2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         go: ['1.14', '1.15']
         os: ['macos-10.15']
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         go: ['1.14', '1.15']
         os: ['macos-10.15']
-        xcode-version: ['12.0', '11.7', '11.6', '11.5', '11.4.1', '11.3.1', '11.2.1', '11.1', '11.0']
+        xcode-version: ['12.1', '12.0', '11.7', '11.6', '11.5', '11.4.1', '11.3.1', '11.2.1']
```
kalyanac, line 23: Should it be

12.0.1 instead of 12.0?
Can we also include 12.2 as a non-mandatory check?

wyk9787: I tried 12.0.1, but it failed somehow: https://github.com/google/pprof/runs/1398239582?check_suite_focus=true

Error: Could not find Xcode version that satisfied version spec: '12.0.1'

Added 12.2. Could you update the check requirements in the settings? Thanks!

kalyanac: I marked 12.1 as required. It shows 12.0.1 in the list FWIW.

kalyanac:
Marked 12.1 as required check in repo settings.

wyk9787:
> Marked 12.1 as required check in repo settings.

Could you also disable these 4 checks for both 11.0 and 11.1 so that this PR can be unblocked?

kalyanac:
Done.

wyk9787:
@kalyanac All checks have passed. Could you approve the PR and merge it to unblock other PRs?

kalyanac (approved):

Issue #558, opened 2020-09-25
Support visualizing tags as nodes


nolanmar511:
The `pprof` CLI has some support for visualizing profiles with tags (the `-tags` output format allows one to see all tags in the profile, [tag filtering](https://github.com/google/pprof/tree/master/doc#tag-filtering) allows one to view only or exclude portions of the profile without tags, and the graph view(s) show some tags as nodeletes), but it's not always possible to visualize which tags are associated with which stacks when using `pprof`.

On a few occasions now, internal users have hacked `pprof` to show certain tags as nodes at the root of the call stack, so it's something that has been useful. For example, if samples are tagged such that they can be associated with particular chunks of work (for example, work done for a particular URL endpoint), it can be useful to add a pseudo-node for a particular tag toward the root of the call stack to group stacks associated with a particular chunk of work in the flame graph view.

On possibility for doing this would be to introduce the following options to `pprof`:
* `-tagroot`: takes in a comma-separated list of label keys and adds the associated key/value pairs  as pseudo-nodes on the stack for samples which have the listed labels. For each sample and label key, if the sample has a label with the specified key, a pseudo-node will be added. Pseudo-nodes will be added at the top of the call stack (closest to the root), with pseudo-node for the label that is first in the list being highest in the callstack and the pseudo-node for the label that is last in the list appearing just before frames in the original profile.
* `-tagleaf`: takes in a comma-separated list of label keys and adds the associated labels as pseudo-nodes on the stack. For each sample and label key, if the sample has a label with the specified key, a pseudo-node will be added. Pseudo-nodes will be added at the bottom of the call stack (after leaf nodes), with pseudo-node for the label that is first in the list being the child of what was the leaf in the original profile and the pseudo-node for the label that is last in the list appearing as the new leaf node.

A key-value pair with key “<key>” and value “value” could be represented as a Function with a function name “<key>:<value>”, and a filename of “<key>”.

mhansen:
Turns out @felixge may have already implemented and documented a variant of this: https://github.com/felixge/pprofutils#labelframes. Linking these up as prior art on this issue.

PR #583, opened 2020-11-30
Add a StringList type to be used for commandline flags.

nilayvaish:
StringList flags would used for developing the feature: add tags as
pseudo locations.
#558


internal/stringlist/stringlist.go
```diff
@@ -0,0 +1,35 @@
...
+// Package stringlist provides type StringList that can be used as a type for
+// commandline flags for providing a slice of strings as input.
+package stringlist
+
+import (
+	"fmt"
+)
+
+// StringList is an alias for slice of strings.
+type StringList []string
```
nolanmar511, line 24: Could []string be used directly?

I'm not immediately seeing the additional utility of having a wrapper class here, though could be missing something.

driver/driver.go
```diff
@@ -94,7 +95,7 @@ type FlagSet interface {
 
 	// StringList is similar to String but allows multiple values for a
 	// single flag
-	StringList(name string, def string, usage string) *[]*string
+	StringList(name string, def string, usage string) *stringlist.StringList
```
nolanmar511, line 98: This is part of pprof's public interface (one passes in `plugin.Options` to `driver.PProf`; we use this internally to, among other things, change the profile fetcher / symbolizer for internal pprof). 

It would be nice if this interface did not have to change. Since we use this interface internally, it will definitely make this PR simpler to land if there aren't changes to pprof's public interface.

nolanmar511 (commented):
Thanks for this change!

Left a few initial comments about the StringList type.

driver/driver.go
aalexand: Also, this public interface shouldn't use an internal type (`github.com/google/pprof/internal/stringlist`) in its signature.

nilayvaish: I think the interface needs to change.  Currently there is no difference between a StringList flag and a String flag.  A StringList flag can only hold a single string for now.  To support a list of strings, we need to provide a type that adds strings to a list.  The new type implements the [flag.Value interface](https://golang.org/src/flag/flag.go?s=32508:32557#L291) to achieve this.

> Also, this public interface shouldn't use an internal type (github.com/google/pprof/internal/stringlist) in its signature.

I have moved the new type to a public directory.

internal/stringlist/stringlist.go
nilayvaish: The new type implements [the flag.Value interface](https://golang.org/src/flag/flag.go?s=32508:32557#L291).  I think we cannot retain the []string type and provide the functions required by the flag package.  I am open to considering other ways to achieve this, but there does not seem to be any.

driver/driver.go
nolanmar511: I think I'd need to see a design document considering this more carefully proposed change before I'd be comfortable changing the public interface. I'm not currently persuaded that the interface change is the best or only way forward; I'd really need to be convinced of that before I'd feel comfortable with a change to the interface. 

As additional context, changing the interface is a really big deal. First, we don't currently version pprof. Now that golang does support versioning a library, if we change the public interface, this becomes something we ought to look into. IMO, it'd be necessary to at least fully consider versioning pprof  using go modules prior to making any breaking changes to the public interface. Next, changes to the public interface can make it really hard to sync our internal copy of pprof, since it becomes possible that code used for our internal version of pprof will break. If it is concluded that changes to the public interface are necessary, I would prefer to first see changes made to our internal copy of pprof to know that nothing internal will break (a PR would still need to be sent to change code, and any changes that need to be made to our dependencies on this library would need to be done when importing pprof from GitHub).

flagset/flagset.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2021 Google Inc. All Rights Reserved.
```
aalexand, line 1: I like pulling this interface out to avoid duplication between driver/driver.go and internal/plugin/plugin.go. Can we make this a separate PR? I think this can go in regardless of other discussions. One thing I'm not sure about is whether it's worth a separate top-level directory - seems too shallow for that. Maybe have it as driver/flagset.go? (and "package driver").

nilayvaish: > Can we make this a separate PR?

Yes.  We just need to decide on the package for flagset.go.

> Maybe have it as driver/flagset.go?

I had tried adding to driver package before, but ran into an import cycle.
package github.com/google/pprof
	imports github.com/google/pprof/driver
	imports github.com/google/pprof/internal/driver
	imports github.com/google/pprof/internal/binutils
	imports github.com/google/pprof/internal/plugin
	imports github.com/google/pprof/driver: import cycle not allowed
package github.com/google/pprof
	imports github.com/google/pprof/driver
	imports github.com/google/pprof/internal/driver
	imports github.com/google/pprof/internal/binutils
	imports github.com/google/pprof/internal/plugin
	imports github.com/google/pprof/driver: import cycle not allowed

Are you OK going with a new package?  I could not think of any other reasonable place for this.

aalexand: I'd prefer not to have a new top-level package. Actually, reading go/go-style/decisions#interfaces I wonder if we should leave things as is with two copies of the interface for now. It's duplication, but it's rather unrelated issue to this PR, so I'd tackle it separately if we ever need to.

driver/driver.go
nilayvaish: I have updated the PR with a different implementation (suggested by lmilewski) that does not change the public interface.  Please take another look.

internal/driver/flags.go
```diff
@@ -44,9 +45,33 @@ func (*GoFlags) String(o, d, c string) *string {
...
+func (sl *stringList) Set(v string) error {
+	*sl = append(*sl, &v)
+	return nil
+}
+
+func (sl *stringList) Get() interface{} {
+	return sl
+}
+
+func (sl *stringList) String() string {
```
nolanmar511, line 59: For my understanding, where (other than tests) is this method used?

internal/driver/flags_test.go
```diff
@@ -0,0 +1,65 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package driver
+
+import (
+	"testing"
+)
+
+func TestString(t *testing.T) {
```
nolanmar511, line 21: It seems a bit different (at least to the best of my knowledge and in the context of golang) to test functions on a private type in golang. Testing the public interface would probably be preferable.

But, I know that there are no tests for flag.go right now; will leave it up to @aalexand  to decide what sort of testing is sufficient.

internal/driver/flags.go
nilayvaish: This function is required to implement the flag.Value interface.  Link: https://golang.org/pkg/flag/#Value.

internal/driver/flags_test.go
nolanmar511: Ping to @nilayvaish and @aalexand.

I'm not sure about this testing. It may be better than not having tests, but I'm a bit skeptical about testing only a private type, rather than the publicly available functions.

nilayvaish: AFAICT, it is not possible to test the public interface without creating a separate process.  Would you be OK with the _test.go file creating a new process with flags?  Otherwise, I think we need to drop the tests all together.

aalexand:
@nilayvaish @nolanmar511 Looking at this again, I'm now not sure why flag changes are needed to support `-tagroot` / `-tagleaf`. Those shouldn't be flags, should be vars in internal/driver/config.go like `-tagfocus` / `-tagignore` and friends.

Also, one thing to watch for with list-like flags is append vs. reset semantics. E.g. in the pprof interactive shell we need a way to specify the full set of values for these flags, without ad-infinitum append behavior.

nilayvaish:
> @nilayvaish @nolanmar511 Looking at this again, I'm now not sure why flag changes are needed to support `-tagroot` / `-tagleaf`. I believe those shouldn't be flags, I think they should be vars in internal/driver/config.go like `-tagfocus` / `-tagignore` and friends.

We are treating `tagroot` and `tagleaf` like `tagfocus and `tagignore`, only the types are different.  I have added another commit to show this.

> 
> Also, one thing to watch for with list-like flags is append vs. reset semantics. E.g. in the pprof interactive shell we need a way to specify the full set of values for these flags, without ad-infinitum append behavior.

This is an interesting point I had not considered before.  I checked [`internal/drive/interactive.go`](https://github.com/google/pprof/blob/master/internal/driver/interactive.go) and it seems like the command line is parsed again.  `TagIgnore`and `TagFocus` have been treated specially in the parse routing: [`parseCommandLine`](https://github.com/google/pprof/blob/master/internal/driver/interactive.go#L219).  We could do something similar for `TagRoot` and `TagLeaf`.  Let me know what you think.

internal/driver/commands.go
```diff
@@ -199,6 +199,14 @@ var configHelp = map[string]string{
...
 		"Numeric tag filter examples: 1kb, 1kb:10kb, memory=32mb:",
 		"String tag filter examples: foo, foo.*bar, mytag=foo.*bar"),
+	"tagleaf": helpText(
+		"Adds a pseudo leaf node for samples with tags matched by regexp",
+		"Use name=value1,value2 syntax to specify the tag and the values",
+		"for which nodes would be added.  Example: latency=1s,2s,3s."),
+	"tagroot": helpText(
+		"Adds a pseudo root node for samples with tags matched by regexp",
+		"Use name=value1,value2 syntax to specify the tag and the values",
+		"for which nodes would be added.  Example: latency=1s,2s,3s."),
```
aalexand, line 209: Couple of questions:

1) The option / flag is of type []string / multi-value. The examples are of the form "foo=val1,val2" What exactly the multi-value here? Is it the comma-separated multi-values as in example, i.e. "foo=val1" and "val2" as two values which is weird? Or is that there is ability to have multiple occurrences of each of these like "-flag=foo=val1,val2 -flag=bar=val3,val4" and the documentation needs to explain this?

2) How are numeric tags handled, including with units like memory? Can we reuse the same value format as in tagfocus / taghide etc.? I feel that we are introducing a new way to specify a set of values and I don't see why different tag* flags can't unify on a single way to do that.

PR #584, opened 2020-12-01
Fixing comment in dotColor function.

yarchi:
Providing score -1 actually returns green not red. Conversely, 1 returns red.
See lines 375 and 378. r = 1 + saturation*score is r = 1 - saturation if score == -1. Thus, r decreases and g < r.

### DIFF
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -322,8 +322,8 @@ func (b *builder) addEdge(edge *Edge, from, to int, hasNodelets bool) {
 }
 
 // dotColor returns a color for the given score (between -1.0 and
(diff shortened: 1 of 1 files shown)


yarchi:
@googlebot I signed it!

kalyanac:
@wyk9787 Can you check why the Windows tests are failing?

yarchi:
It says:
`
ok  	github.com/google/pprof/profile	0.073s
?   	github.com/google/pprof/third_party/d3	[no test files]
?   	github.com/google/pprof/third_party/d3flamegraph	[no test files]
?   	github.com/google/pprof/third_party/svgpan	[no test files]
FAIL
`
Looks like the reason that this three folders don't have corresponding unit tests.

wyk9787 (approved):
Thanks for the fix!

PR #588, opened 2020-12-08
Add GCUs as a unit type

nolanmar511:
This change also refactors measurement.go to allow more unit conversion code to be shared. 

Here are a few reasonable concerns about this change:
* 'cycles' is no longer treated as a time unit.
* Using for loops,  rather than switch statements may increase computation complexity. I didn't notice any difference between `go tool pprof` and `pprof` built with this change when viewing profiles that were relatively small (1-2 MiB profile.pb.gz's; using -`top` output format); but pprof might be slower for larger profiles.
* I did not use "github.com/google/go-cmp/cmp" and "github.com/google/go-cmp/cmp/cmpopts" for float comparison in testing. I wrote a `floatEqual` function because we've generally avoided adding dependencies in the past and I was having a hard time making unit tests on GitHub pass when I added these imports.


internal/measurement/measurement.go
```diff
@@ -172,69 +176,55 @@ func Percentage(value, total int64) string {
 	}
 }
 
+// unit includes a list of names representing a specific unit and a factor
+// which one can multiple a value in the specified unit by to get the value
+// in terms of the base unit.
+type unit struct {
+	preferredName string
+	names         []string
```
aalexand, line 184: Thought on naming: maybe `aliases`?

internal/measurement/measurement.go
```diff
@@ -172,69 +176,55 @@ func Percentage(value, total int64) string {
...
+	{"GB", []string{"gb", "gbyte", "gigabyte"}, 1024 * 1024 * 1024},
+	{"TB", []string{"tb", "tbyte", "terabyte"}, 1024 * 1024 * 1024 * 1024},
+	{"PB", []string{"pb", "pbyte", "petabyte"}, 1024 * 1024 * 1024 * 1024 * 1024},
+}
+
 // isMemoryUnit returns whether a name is recognized as a memory size
 // unit.
 func isMemoryUnit(unit string) bool {
-	switch strings.TrimSuffix(strings.ToLower(unit), "s") {
-	case "byte", "b", "kilobyte", "kb", "megabyte", "mb", "gigabyte", "gb":
```
aalexand, line 179: Interesting that terabyte wasn't recognized as a memory unit?

internal/measurement/measurement.go
```diff
@@ -172,69 +176,55 @@ func Percentage(value, total int64) string {
...
+	return convertUnit(value, fromUnit, toUnit, unit{"B", []string{}, 1.0}, uf, as)
+}
+
+var timeUnits = []unit{
+	{"ns", []string{"ns", "nanosecond"}, float64(time.Nanosecond)},
+	{"μs", []string{"μs", "us", "microsecond"}, float64(time.Microsecond)},
+	{"ms", []string{"ms", "millisecond"}, float64(time.Millisecond)},
+	{"s", []string{"s", "sec", "second"}, float64(time.Second)},
+	{"hrs", []string{"hour", "hr"}, float64(time.Hour)},
+	{"days", []string{"day"}, 24 * float64(time.Hour)},
```
aalexand, line 225: Like Go time package, I'd prefer not to go above time.Hour because DST. A day is not really 24 hours - can be 23, 24 or 25, routinely.

internal/measurement/measurement.go
```diff
@@ -172,69 +176,55 @@ func Percentage(value, total int64) string {
...
-	case "pb", "pbyte", "petabyte":
-		output, toUnit = float64(value)/(1024*1024*1024*1024*1024), "PB"
-	}
-	return output, toUnit, true
+	return convertUnit(value, fromUnit, toUnit, unit{"B", []string{}, 1.0}, uf, as)
+}
+
+var timeUnits = []unit{
+	{"ns", []string{"ns", "nanosecond"}, float64(time.Nanosecond)},
+	{"μs", []string{"μs", "us", "microsecond"}, float64(time.Microsecond)},
```
aalexand, line 221: FWIW, I'm a bit scared to have the non-ASCII character in the canonical name.

internal/measurement/measurement.go
```diff
@@ -172,69 +176,55 @@ func Percentage(value, total int64) string {
 	}
 }
 
+// unit includes a list of names representing a specific unit and a factor
+// which one can multiple a value in the specified unit by to get the value
+// in terms of the base unit.
+type unit struct {
+	preferredName string
```
aalexand, line 183: Maybe "canonical" instead of "preferred"? Not a strong opinion.

internal/measurement/measurement.go
```diff
@@ -172,69 +176,55 @@ func Percentage(value, total int64) string {
...
+	names         []string
+	factor        float64
+}
+
+var memoryUnits = []unit{
+	{"B", []string{"b", "byte"}, 1},
+	{"kB", []string{"kb", "kbyte", "kilobyte"}, 1024},
```
aalexand, line 190: FWIW / nit: I like expressing IEC constants as shifts of 2: 1 << 10, 1 << 20, 1 << 30, 1 << 40, 1 << 50.

aalexand, line 190: It's a bit sad that pprof confuses kB and KiB, MB and MiB and so on. What we have here is IEC binary prefixes, not SI decimal ones...

aalexand: But we are not fixing it now :)

internal/measurement/measurement.go
```diff
@@ -262,67 +249,125 @@ func timeLabel(value int64, fromUnit, toUnit string) (v float64, u string, ok bo
...
+			}
+		}
+	}
+	return unit, 0, false
+}
+
+// autoscale takes in the value with units of base unit and returns
+// that value scaled to a reasonable unit if a reasonable unit is
+// found.
+func autoscale(value float64, units []unit) (float64, string, bool) {
```
aalexand, line 360: autoScale?

internal/measurement/measurement.go
```diff
@@ -262,67 +249,125 @@ func timeLabel(value int64, fromUnit, toUnit string) (v float64, u string, ok bo
...
+	{"k·GCU", []string{"kilogcu"}, 1e3},
+	{"M·GCU", []string{"megagcu"}, 1e6},
+	{"G·GCU", []string{"gigagcu"}, 1e9},
+	{"T·GCU", []string{"teragcu"}, 1e12},
+	{"P·GCU", []string{"petagcu"}, 1e15},
+}
+
+var caseSensitiveGCUUnits = []unit{
+	{"n·GCU", []string{"n·GCU", "n*GCU"}, 1e-9},
+	{"μ·GCU", []string{"μ·GCU", "μ*GCU", "u·GCU", "u*GCU"}, 1e-6},
```
aalexand, line 279: I'm again a bit cautious about unicode characters in a CLI tool. Is it safe to assume that everybody's terminal is capable of printing these characters?

internal/measurement/measurement.go
nolanmar511: `aliases` is much better!

internal/measurement/measurement.go
nolanmar511: I think terabyte was a valid unit for the purposes of display, but a unit that probably wouldn't have been used in the profile.proto (since that represents data in integers, and 1TiB as the smallest possible value is pretty coarse). 

But this also did surprise me.

Marking as resolved.

internal/measurement/measurement.go
nolanmar511: Done.

internal/measurement/measurement.go
nolanmar511: Done (here and elsewhere)

internal/measurement/measurement.go
nolanmar511: Done and thanks! Canonical reads much better to me.

internal/measurement/measurement.go
nolanmar511: Done.

nolanmar511: Do agree with this!

I was considering if a generic approach to unit conversion could work; but kilo meaning both 10^3 and 2^10 made that less feasible (though, using IEC prefixes in addition to SI prefixes doesn't resolve all issues with the approach I was considering).

I'm not marking this as resolved, because I'd like to switch the canonicalName for these units to  using IEC prefixes as a follow-up (this isn't a complete fix, though).

internal/measurement/measurement.go
nolanmar511: Done.

internal/measurement/measurement.go
nolanmar511: Done (here and elsewhere)

nolanmar511:
PTAL

internal/measurement/measurement.go
```diff
@@ -262,67 +246,124 @@ func timeLabel(value int64, fromUnit, toUnit string) (v float64, u string, ok bo
...
+	{"m*GCU", []string{"milligcu"}, 1e-3},
+	{"GCU", []string{"gcu"}, 1},
+	{"k*GCU", []string{"kilogcu"}, 1e3},
+	{"M*GCU", []string{"megagcu"}, 1e6},
+	{"G*GCU", []string{"gigagcu"}, 1e9},
+	{"T*GCU", []string{"teragcu"}, 1e12},
+	{"P*GCU", []string{"petagcu"}, 1e15},
+}
+
+var caseSensitiveGCUUnits = []unit{
```
aalexand, line 274: Do we need these now? Trying to understand if this is a YAGNI that we can postpone.

nolanmar511: Removed this.

internal/measurement/measurement.go
```diff
@@ -262,67 +242,104 @@ func timeLabel(value int64, fromUnit, toUnit string) (v float64, u string, ok bo
...
+		return v / defaultUnit.factor, defaultUnit.canonicalName, true
+	}
+	return v / toUnitFactor, toUnit, true
+}
+
+// unitFactor returns the canonical name for the unit, the factor by which one
+// must multiply a value specified in terms of unit in order to get the value
+// specified in terms of the base unit, and a boolean to indicated if a unit
+// was identified for the specified unit within the slice units.
+func unitFactor(unit string, units []unit) (string, float64, bool) {
```
aalexand, line 318: This is optional, but I would consider a couple of things API-wise, see WDYT.

We could make []unit a type so that it can be a receiver. I'd also name unitFactor as findByAlias and return *unit from it - the reason is that it returns more than just the factor, it pretty much searches for a unit by alias.

Having a knownUnit function could simplify isTimeUnit / isMemoryUnit / isGCUUnit a bit.

So, something like:

```
type unitList []unit
...
// findByAlias .... Returns nil if the unit with such alias is not found.
func (us unitList) findByAlias(alias string) *unit {
  ...
}

func (us unitList) autoScale(...) ... {
  ...
}

func (us unitList) sniffUnit(unit string) *unit {
  return us.findByAlias(strings.TrimSuffix(strings.ToLower(unit), "s"))
}
```

With this, I think you could remove isTimeUnit and its friends as it simply becomes `if timeUnits.sniffUnit(s) != nil`.

I think `convertUnit` would also simplify, since now it can also accept a receiver and it doesn't need to accept the function pointers. I could be wrong on this, but I feel that this can simplify.

nolanmar511: Thanks! This is much simpler. 

One thing that happened with this change: 'cycle' is no longer treated as a time unit. There weren't any tests which failed because of this, but wanted to call it out in case this could be a problem.

nolanmar511:
PTAL

internal/measurement/measurement.go
```diff
@@ -172,157 +176,120 @@ func Percentage(value, total int64) string {
...
+		{"ms", []string{"ms", "millisecond"}, float64(time.Millisecond)},
+		{"s", []string{"s", "sec", "second"}, float64(time.Second)},
+		{"hrs", []string{"hour", "hr"}, float64(time.Hour)},
+	},
+	defaultUnit: unit{"s", []string{}, float64(time.Second)},
+}
+
+var gcuUnits = unitType{
+	units: []unit{
+		{"n*GCU", []string{"nanogcu"}, 1e-9},
```
wyk9787, line 284: Just curious: why do we use "n*GCU" here instead of "nGCU"?

internal/measurement/measurement.go
```diff
@@ -172,157 +176,120 @@ func Percentage(value, total int64) string {
...
-		d = time.Duration(value) * time.Millisecond
-	case "second", "sec", "s":
-		d = time.Duration(value) * time.Second
-	case "cycle":
-		return float64(value), "", true
-	default:
+// convertUnit converts a value from the fromUnit to the toUnit, autoscaling
+// the value if the toUnit is "minimum" or "auto". If the fromUnit is not
+// included in units, then a false boolean will be returned. If the toUnit
+// is not in units, the value will be returned in terms of the default unit.
```
wyk9787, line 239: Maybe "unitType" is better here? I think you are referring to "ut unitType", right?

wyk9787 (dismissed):

aalexand (dismissed):
This does look simpler - nice! Thanks!

internal/measurement/measurement.go
nolanmar511: It's consistent w/ changes made to our internal frontend. So, this is just for consistency.

internal/measurement/measurement.go
nolanmar511: Thanks for catching this!

aalexand (approved):

PR #593, opened 2020-12-21
Support disasm on Windows

### LINKED ISSUE #592: Windows Disasm support
I've been annoyed of not having Windows disasm support for Go profiling a while now and was thinking of adding it. The thing that seems to be missing is PE support in `(*BinUtils).Open`, I'm sure I'll find more once I dig deeper. Is there something I should watch out for?

egonelbre:
This change allows go profiles to be disassembled in pprof. Currently the profiled program must be compiled with `go build -buildmode=exe` to disable ASLR.

Fixes #592

# Manual testing part

This requires having either GCC or LLVM installed, the best way I found to install them is:
* TDM-GCC: https://jmeubank.github.io/tdm-gcc/articles/2020-03/9.2.0-release
* LLVM using `choco install llvm`

For testing I used go benchmarks. A minimal benchmark:

```
package example

import "testing"

func BenchmarkArray(t *testing.B) {
	x := V{}
	for i := 0; i < t.N; i++ {
		x = x.Add(V{X: float32(i), Y: float32(i)})
	}
	g = x
}

var g V

type V struct{ X, Y float32 }

//go:noinline
func (v V) Add(b V) V { return V{X: v.X + b.X, Y: v.Y + b.Y} }
```

Then ran the benchmark with:
```
go test -buildmode=exe -bench . -cpuprofile cpu.prof
pprof --disasm V.Add  test.test.exe cpu.prof
```

Also tested with cgo benchmarks https://github.com/golang/go/blob/master/misc/cgo/test/test.go#L1075.


egonelbre:
It seems that Windows doesn't have code coverage setup.

internal/binutils/addr2liner.go
```diff
@@ -70,7 +70,11 @@ func (a *addr2LinerJob) write(s string) error {
 }
 
 func (a *addr2LinerJob) readLine() (string, error) {
-	return a.out.ReadString('\n')
+	s, err := a.out.ReadString('\n')
```
kalyanac, line 73: What is the motivation for moving this from readString() to readLine()?

kalyanac (commented):
Overall, looks good to me. 

Please also include doc changes to describe how to use on Windows, supported configuratoins, packages to be installed (llvm), etc.

internal/binutils/addr2liner.go
egonelbre: This it's more obvious that every `ReadString("\n")` from the buffer has been properly trimmed. Otherwise it's easy to lose track of the location. But, I don't think it's strictly necessary.

egonelbre:
> Please also include doc changes to describe how to use on Windows, supported configuratoins, packages to be installed (llvm), etc.

Where should the docs be updated?

egonelbre:
But, I tested the code with llvm and tdm-gcc (9.2.0) in combination with `go build -buildmode=exe`.

kalyanac:
> But, I tested the code with llvm and tdm-gcc (9.2.0) in combination with `go build -buildmode=exe`.

Can you describe the test setup in the PR description?

kalyanac:
> > Please also include doc changes to describe how to use on Windows, supported configuratoins, packages to be installed (llvm), etc.
> 
> Where should the docs be updated?

How about adding it to Readme.md?

kalyanac:
@wyk9787 Can you also review?

egonelbre:
> Can you describe the test setup in the PR description?

Added the test description, let me know whether something needs clarification.

> How about adding it to Readme.md?

It felt weird to include a section just for Windows there, since there's no equivalent section for Linux. Or at least, there is already a paragraph:

```
If the profile samples contain machine addresses, pprof can symbolize them through the use of the native binutils tools (addr2line and nm).
```

I could extend this to mention llvm/gcc binutils specifically. Adding a sentence something like:

```
Easiest way to get the necessary tools is to install either `llvm` or `gcc`.
```

With regards to the `-buildmode=exe` requirement, I'm looking into implementing the necessary mapping and hopefully getting it into `go v1.17`.

wyk9787 (dismissed):

kalyanac:
@egonelbre 
I am not really familiar with the toolchain on Windows. I am not sure if Microsoft tools also include `addr2line` and `nm`. So I think it might be better for future users to know what toolchain is supported.  How about the following?

```
**Viewing disassembly on Windows:**
To view disassembly of profiles collected from Go programs compiled as Windows executables, the executable must be built with `go build -buildmode=exe`. LLVM or GCC must be installed so required tools like `addr2line` and `nm` are available to `pprof`.
```

kalyanac (approved):

internal/binutils/addr2liner.go
```diff
@@ -116,11 +120,7 @@ func newAddr2Liner(cmd, file string, base uint64) (*addr2Liner, error) {
 }
 
 func (d *addr2Liner) readString() (string, error) {
```
aalexand, line 122: Is this function needed now at all?

internal/binutils/addr2liner_llvm.go
```diff
@@ -98,11 +102,7 @@ func newLLVMSymbolizer(cmd, file string, base uint64) (*llvmSymbolizer, error) {
 }
 
 func (d *llvmSymbolizer) readString() (string, error) {
```
aalexand, line 104: Ditto.

internal/binutils/testdata/hello.c
```diff
@@ -0,0 +1,6 @@
+#include <stdio.h>
+
+int main() {
+  printf("Hello, world!\n");
+  return 0;
+}
```
aalexand, line 6: end of line is missing

internal/binutils/testdata/lib.c
```diff
@@ -0,0 +1,7 @@
+int foo() {
+  return 1;
+}
+
+int bar() {
+  return 2;
+}
```
aalexand, line 7: end of line is missing

internal/binutils/addr2liner.go
egonelbre: I think it could be removed, but the rest of the code would use `d.rw.readLine` in that case. I don't have enough context to know whether that's the desired code or not.

internal/binutils/testdata/hello.c
egonelbre: Is there a linter that should've caught this?

aalexand: I'm not well familiar with C/C++ OSS linters and it's probably not worth the effort integrating it into this repo.

internal/binutils/addr2liner.go
egonelbre: Created https://github.com/google/pprof/pull/596 to address the comments.

PR #597, opened 2021-01-13
Stop using math.Round to allow for pprof import

nolanmar511:
When @wyk9787 was importing pprof code from here into our internal repository, a test failed with the error `undefined: math.Round`. 

This change removes the usage of math.Round to allow for import of pprof (I manually applied this change while importing pprof and tests built and passed).

I'm open to other fixes; just want to be sure we unblock importing pprof.

### DIFF
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


aalexand:
It would be nice to understand why this happened - is this Go version incompatibility or what? I'd be surprised if math.Round is particularly new.

aalexand:
There is offline discussion on this, let's resolve that first. It's unclear that supporting pre-1.10 go versions is something we want to do.

PR #598, opened 2021-01-16
Enable symbolization of data objects with the llvm symbolizer

gmarin13:
This PR adds support for symbolizing virtual data addresses with the llvm-symbolizer. As a prerequisite, we have to compute the correct GetBase offset for data mappings, regardless of which local symbolizer is used. 

One of the commits, changes the current behavior from always computing a base offset relative to the executable program segment, to an approach that tries to identify the program segment associated with a given mapping and computing the offset relative to that segment. This change is enabled only for user space mappings. For kernel mappings, we continue to look for the kernel executable segment.

With this change, we drop support for fake mappings that span the entire address space. From an offline discussion with aalexand@:
- fake mappings were added by https://github.com/google/pprof/pull/89 to handle profiles with no mappings that would be generated by golang runtime sometimes;
- local experiments by aalexand@ show the fake mapping is present in Go 1.7 heap profiles, but not in Go 1.8 profiles. This may be related to the fact that the older Go versions didn't use profile.proto and generated legacy text format instead;
- Go 1.7 is not a supported version anymore, so we are OK with dropping support for the fake mapping.

### DIFF
--- a/internal/binutils/addr2liner_llvm.go
+++ b/internal/binutils/addr2liner_llvm.go
@@ -43,10 +43,12 @@ type llvmSymbolizerJob struct {
 	cmd *exec.Cmd
 	in  io.WriteCloser
 	out *bufio.Reader
+	// llvm-symbolizer requires the symbol type, CODE or DATA, for symbolization.
+	symType string
 }
 
 func (a *llvmSymbolizerJob) write(s string) error {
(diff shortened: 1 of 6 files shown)


kalyanac:
Tests failed. Please mark this as draft until it's ready for review.

aalexand (approved):

PR #601, opened 2021-01-20
Update the function comment for FindProgHeaderForMapping.

gmarin13:
Added text to specify that it can return a nil program header in some cases.

### DIFF
--- a/internal/elfexec/elfexec.go
+++ b/internal/elfexec/elfexec.go
@@ -286,7 +286,8 @@ func FindTextProgHeader(f *elf.File) *elf.ProgHeader {
 
 // FindProgHeaderForMapping returns the loadable program segment header that is
 // fully contained in the runtime mapping with file offset pgoff and memory size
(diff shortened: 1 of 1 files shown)


internal/elfexec/elfexec.go
```diff
@@ -286,7 +286,8 @@ func FindTextProgHeader(f *elf.File) *elf.ProgHeader {
 
 // FindProgHeaderForMapping returns the loadable program segment header that is
 // fully contained in the runtime mapping with file offset pgoff and memory size
-// memsz, or an error if the segment cannot be determined.
+// memsz, or an error if the segment cannot be determined. The function returns
+// a nil progran header and no error if the ELF binary has no loadable segments.
```
aalexand, line 290: typo: progran

gmarin13: Oops, updated.

aalexand (approved):

PR #604, opened 2021-02-02
Chrome/Chromium no longer allows access to local files

chewxy:
This leaves a very annoying process where one does something like `weblist funcname` and then get greeted with something that looks like this:

image

The only workaround is to launch Chrome/Chromium with the `-disable-web-security` flag. This is clearly not ideal for someone who wants to use the browser as you know, a web browser. 

So right now, when I do `web` or `weblist XXX`, pprof opens Chromium, which I then have to copy the URL and open it in Firefox. That's awfully annoying.

p/s: I have read the comments arguing for the rationale of putting `xdg-open` last, but anyone sane in 2021 would have their SVGs open in browsers by default anyway.


aalexand:
I'm somewhat hesitant to change the order of browsers in this list, this can have unexpected effect to some users. I think you should be able to use the `BROWSER` environment variable to set your browser of choice.

chewxy:
yeah I've been using that as a workaround. But it's still a pretty user hostile experience, as well as polluting of the environment variables, which does affect performance and performance optimization (it's inconsequential when you're running `go tool pprof` but on other programs, having an additional environmental variable DOES actually affect measurement of performance. So you'd have to run `BROWSER=firefox go tool pprof XXX` instead of `export BROWSER=firefox`)

PR #617, opened 2021-03-29
UI: add download link to web UI

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

This pr adds a download link in the UI to download the current profile.

image

### DIFF
--- a/internal/driver/webhtml.go
+++ b/internal/driver/webhtml.go
@@ -370,6 +370,12 @@ table tr td {
     </div>
   </div>
 
+  <div id="download" class="menu-item">
+    <div class="menu-name">
+      <a href="./download">Download</a>
+    </div>	
+  </div>
+
   <div>
     <input id="search" type="text" placeholder="Search regexp" autocomplete="off" autocapitalize="none" size=40>
   </div>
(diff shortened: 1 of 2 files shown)


sandangel:
hi, is there any update on this issue?

internal/driver/webhtml.go
```diff
@@ -370,6 +370,12 @@ table tr td {
     </div>
   </div>
 
+  <div id="download" class="menu-item">
+    <div class="menu-name">
+      <a href="./download">Download</a>
```
aalexand, line 375: Could we style this so that it doesn't look like a hyperlink in the UI? On the screenshot included in the PR description the new element stands out a lot, more than it should as it's not the most important action.

sandangel: I can help with styling this. Do you want the new element looks like other elements, just doesn't have the dropdown?

aalexand: Yes, that would work. Or make it an item in the "View" menu. It's not exactly a view, but I think it's close enough to be there.

sandangel: hi @aalexand I created the PR linked below. Could you please check?

aalexand: Merged #634. Thanks!

aalexand:
This was addressed in #634 - closing.

PR #619, opened 2021-04-06
Parse numeric labels with a value of zero if units are specified.

xyzsam:
By default, a numeric label with a value of zero is interpreted as an
empty label which will be ignored during parsing. However, a zero value
can be a valid label value. Instead of ignoring it altogether, parse it
as a zero-valued numeric label if num_unit is set.

### DIFF
--- a/profile/encode.go
+++ b/profile/encode.go
@@ -308,7 +308,7 @@ func (p *Profile) postDecode() error {
 			if l.strX != 0 {
 				value, err = getString(p.stringTable, &l.strX, err)
 				labels[key] = append(labels[key], value)
-			} else if l.numX != 0 {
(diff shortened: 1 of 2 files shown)


nolanmar511 (approved):

aalexand:
@xyzsam @nolanmar511 It looks like we didn't add a test here? Looking at the *_test.go file it looks like the test data was updated, but I don't see an actual check added. I could be missing something.

aalexand:
> @xyzsam @nolanmar511 It looks like we didn't add a test here? Looking at the *_test.go file it looks like the test data was updated, but I don't see an actual check added. I could be missing something.

Oh, I guess it checks the whole serialize <-> deserialize loop? I assume the test failed before the fix?

PR #622, opened 2021-04-12
Replace Base() with ObjAddr() in the plugin.ObjFile interface.

gmarin13:
Modify the plugin.ObjFile interface by replacing the "Base() uint64" method
with an "ObjAddr(addr uint64) (uint64, error)" method. The latter computes
the objdump address corresponding to the given runtime address.

The change removes direct access to the base value, but enables us to delay
computing the base until it is actually needed.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -159,8 +159,8 @@ type ObjFile interface {
 	// Name returns the underlying file name, if available.
 	Name() string
 
-	// Base returns the base address to use when looking up symbols in the file.
-	Base() uint64
+	// ObjAddr returns the objdump address corresponding to a runtime address.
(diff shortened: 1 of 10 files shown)


aalexand:
@gmarin13 Looks like some tests failed on Windows.

```
FAIL
FAIL	github.com/google/pprof/internal/binutils	3.685s
```

gmarin13:
> @gmarin13 Looks like some tests failed on Windows.
> 
> ```
> FAIL
> FAIL	github.com/google/pprof/internal/binutils	3.685s
> ```

I had one typo in the test file.

aalexand (approved):

PR #623, opened 2021-04-13
Delay base computation for ELF files until we have a sample address.

gmarin13:
We can use the sample's offset into the mapping to differentiate between small
segments that are in the same file page when associating a segment with the
mapping.


gmarin13:
Updated the code. PTAL.

gmarin13:
All the checks pass after the Golang runtime fix.

aalexand (approved):

PR #624, opened 2021-04-14
Source listing now handles locations that do not fall in a known mapping

### LINKED ISSUE #621: blank webui.source page  without any error
# blank webui.source page with self built pprof, but go tool pprof works.
 
* commit 94a9f03dee38882adc8bdfc42cdd6a04f8e7056e 

* ubuntu Linux  5.4.0-65-generic #73-Ubuntu SMP Mon Jan 18 17:25:17 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
* go1.15.5 linux/amd64

### What did you do?
* build
  * go version go1.15.5 linux/amd64
  * `go build`
* generate profile data file, with src file
  * [pprof-data-t3.zip](https://github.com/google/pprof/files/6276149/pprof-data-t3.zip) 
* run `./pprof -http :9999 t3.pf`

webui source page

### What did you see instead?
* top, graph, ...  other pages work fine
* blank source page, no error, no log

### `go tool pprof` works
`go tool pprof -http :9999 t3.pf`
* go version go1.15.5 linux/amd64
* go version go1.16.2 linux/amd64

### build for different OS
the same, blank source page
* linux
* windows 

different graphviz version, same.
* graphviz 2.26
* graphviz 2.40
* graphviz 2.47

### -source_path option
* `./pprof -http :9999  -source_path=.  t3.pf`

## Question
* is there any build flags should i put when `go build` ? (to build as go tool version)
* is there any secret pprof cmdline options to enable verbose output for debug?
  * ./pprof -v
* how to know binary pprof version?  (to build as go tool version)
  * `go tool pprof -version`

ghemawat:
Some profiles can contain locations that do not fall inside a known
mapping. Treat these better when producing source listings by using
any source information present directly in the profile.Location.

Details: any location that does not fall in a mapping is stored in
a synthetic code object. The synthetic code object is used (instead
of Disasm) to fetch source info for such locations.

Tweaked generated html to clicking for source lines that have no
inlining/assembly to display.

Fixes #621.

### DIFF
--- a/internal/report/source.go
+++ b/internal/report/source.go
@@ -139,6 +139,7 @@ type sourcePrinter struct {
 	sym        *regexp.Regexp             // May be nil
 	files      map[string]*sourceFile     // Set of files to print.
 	insts      map[uint64]instructionInfo // Instructions of interest (keyed by address).
+	addr2obj   map[uint64]plugin.ObjFile
 
 	// Set of function names that we are interested in (because they had
 	// a sample and match sym).
@@ -214,6 +215,7 @@ func newSourcePrinter(rpt *Report, obj plugin.ObjTool, sourcePath string) *sourc
 		synth:       synthCode{mappings: mappings},
 		objectTool:  obj,
 		objects:     map[string]plugin.ObjFile{},
+		addr2obj:    map[uint64]plugin.ObjFile{},
 		sym:         rpt.options.Symbol,
 		files:       map[string]*sourceFile{},
 		insts:       map[uint64]instructionInfo{},
@@ -235,14 +237,14 @@ func newSourcePrinter(rpt *Report, obj plugin.ObjTool, sourcePath string) *sourc
 	cum := map[uint64]int64{}
 
 	// Record an interest in the function corresponding to lines[index].
-	markInterest := func(addr uint64, lines []profile.Line, index int) {
-		fn := lines[index]
+	markInterest := func(addr uint64, loc *profile.Location, index int) {
+		fn := loc.Line[index]
 		if fn.Function == nil {
 			return
 		}
 		sp.interest[fn.Function.Name] = true
 		sp.interest[fn.Function.SystemName] = true
-		addrs[addr] = true
+		sp.registerAddress(addrs, addr, loc)
 	}
 
 	// See if sp.sym matches line.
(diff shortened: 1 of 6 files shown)


fanlix:
- [x] checked

This fix my problem, webui.source page works fine now.

zikaeroh:
This PR doesn't seem to fix things for Go profiles, unfortunately. The source panel is still empty.

ghemawat:
Zik, could you point at instructions on how to get a profile that breaks? Thanks.

zikaeroh:
Sure, here is an example that only needs Go.

```
$ go test -run=- -bench 'Write' -benchmem -cpuprofile cpu.out -memprofile mem.out encoding/csv
$ pprof -http=: mem.out
```

internal/report/source.go
```diff
@@ -1029,3 +1102,22 @@ func canonicalizeFileName(fname string) string {
...
+	mappings := append([]*profile.Mapping{}, src...)
+	sort.Slice(mappings, func(i, j int) bool { return mappings[i].Start < mappings[j].Start })
+	return mappingList(mappings)
+}
+
+// find returns the mapping in m that contains addr, or nil if no such mapping exists.
+func (m mappingList) find(addr uint64) *profile.Mapping {
+	// Find first mapping that ends after addr.
+	i := sort.Search(len(m), func(i int) bool { return m[i].Limit > addr })
+	if i < len(m) && m[i].Start <= addr {
```
aalexand, line 1119: I believe this makes an assumption that mappings are not overlapping in the virtual address space in the profile. This isn't necessarily true in the general case - e.g. the profile could be a merged profile of the data from multiple processes.

internal/report/source.go
```diff
@@ -132,11 +132,14 @@ func printWebSource(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
 // sourcePrinter holds state needed for generating source+asm HTML listing.
 type sourcePrinter struct {
 	reader     *sourceReader
+	mappings   mappingList
+	synth      synthCode
```
aalexand, line 136: I wonder if the code would be simpler if we restrict the case of absent mappings to zero address, rather than any address, for now. I believe I'm not aware of practical cases where the address is present and the mapping is missing.

aalexand (commented):
Should we add a test that fails before the fix and passes after?

ghemawat:
I redid the way things worked to respond to comments. Those changes will show up in a separate pull request and I will close this one.

PR #626, opened 2021-04-27
Add pseudo locations.

nilayvaish:
Pseudo locations can be added using tagroot and tagleaf commands.  For
example, the following flame graph was generated by adding pseudo
locations in the following fashion:
(pprof) tagroot=method
(pprof) tagleaf=duration

http://pprof.corp.google.com/user-profile?id=41836144a6bd473147ab7e1f248f55bb&tab=flame

### DIFF
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -199,6 +199,12 @@ var configHelp = map[string]string{
 		"Use name=value syntax to limit the matching to a specific tag.",
 		"Numeric tag filter examples: 1kb, 1kb:10kb, memory=32mb:",
 		"String tag filter examples: foo, foo.*bar, mytag=foo.*bar"),
+	"tagleaf": helpText(
+		"Adds a leaf location for samples with a matching tag.",
+		"Example: tagleaf=latency, where latency is numeric tag, adds leaf locations with name latency=<tagvalue>"),
+	"tagroot": helpText(
+		"Adds a root node for samples with a matching tag.",
+		"Example: tagroot=method, where method is string tag, adds root locations with name method=<tagvalue>"),
 	"tagshow": helpText(
 		"Only consider tags matching this regexp",
 		"Discard tags that do not match this regexp"),
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -99,6 +99,7 @@ func generateRawReport(p *profile.Profile, cmd []string, cfg config, o *plugin.O
 			return nil, nil, err
 		}
 	}
+	addRootLeafTagsAsLocations(p, cfg, o.UI)
 	if err := aggregate(p, cfg); err != nil {
 		return nil, nil, err
 	}
@@ -341,4 +342,10 @@ func valueExtractor(ix int) sampleValueFunc {
 	return func(v []int64) int64 {
 		return v[ix]
 	}
+}
+
+func addRootLeafTagsAsLocations(p *profile.Profile, cfg config, ui plugin.UI) {
+	tr, tl := p.AddRootLeafTagsAsLocations(cfg.TagRoot, cfg.TagLeaf)
+	warnNoMatches(len(cfg.TagRoot) == 0 || tr, "TagRoot", ui)
+	warnNoMatches(len(cfg.TagLeaf) == 0 || tl, "TagLeaf", ui)
 }
--- a/internal/driver/config.go
+++ b/internal/driver/config.go
@@ -29,6 +29,8 @@ type config struct {
 	DivideBy            float64 `json:"-"`
 	Normalize           bool    `json:"normalize,omitempty"`
 	Sort                string  `json:"sort,omitempty"`
+	TagLeaf             string  `json:"tagleaf,omitempty"`
+	TagRoot             string  `json:"tagroot,omitempty"`
 
 	// Filtering options
 	DropNegative bool    `json:"drop_negative,omitempty"`
@@ -145,6 +147,8 @@ func init() {
 		"show_from":            "sf",
 		"tagfocus":             "tf",
 		"tagignore":            "ti",
+		"tagleaf":              "tl",
+		"tagroot":              "tr",
 		"tagshow":              "ts",
 		"taghide":              "th",
 		"mean":                 "mean",
--- a/profile/root_leaf_tags.go
+++ b/profile/root_leaf_tags.go
@@ -1,0 +1,121 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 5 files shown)


internal/driver/driver.go
```diff
@@ -342,3 +343,9 @@ func valueExtractor(ix int) sampleValueFunc {
 		return v[ix]
 	}
 }
+
+func addPseudoLocs(p *profile.Profile, cfg config, ui plugin.UI) {
+	tr, tl := p.AddPseudoLocs(cfg.TagRoot, cfg.TagLeaf)
```
aalexand, line 348: Could we follow the "tag root / leaf" terminology, without introducing the "pseudo locations" speak? E.g. addRootLeafTagLocations.

internal/driver/driver.go
```diff
@@ -342,3 +343,9 @@ func valueExtractor(ix int) sampleValueFunc {
 		return v[ix]
 	}
 }
+
+func addPseudoLocs(p *profile.Profile, cfg config, ui plugin.UI) {
```
aalexand, line 347: nit: is abbreviating Locations to Locs common in the pprof code? Just something to check for consistency.

internal/driver/driver.go
aalexand, line 348: I'd prefer this method to be an internal method. We try to give higher guarantees about what's in the public profile package and here I am not sure we want to commit to maintain this API surface.

profile/pseudo_locs.go
```diff
@@ -0,0 +1,117 @@
...
+		return addStringLoc(p, tag, l1), true
+	}
+	if l2, ok2 := s.NumLabel[tag]; ok2 {
+		l3 := s.NumUnit[tag]
+		return addNumLoc(p, tag, l2, l3), true
+	}
+	return nil, false
+}
+
+// AddPseudoLocs adds root and leaf locations for tags in a profile as
```
aalexand, line 99: Document the optionality of the arguments.

Document the return values.

profile/pseudo_locs.go
```diff
@@ -0,0 +1,117 @@
...
+// requested in the provided options.
+func (p *Profile) AddPseudoLocs(rootTag, leafTag string) (bool, bool) {
```
aalexand, line 101: Please follow top-down order for functions in Go. E.g. public go first and so on in roughly BFS order.

profile/pseudo_locs_test.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, s := range p.Sample {
+		locs = append(locs, len(s.Location))
+	}
+	return locs
+}
+
+func TestAddPseudoLocs(t *testing.T) {
+	// Perform several forms of pseudo location addition on the test
+	// profile.
+	notfound := "notfound"
```
aalexand, line 33: const

profile/pseudo_locs_test.go
```diff
@@ -0,0 +1,64 @@
...
+		// One root tag.
+		{"AddRootTag", key1, "", true, false, []int{2, 3, 3, 3, 3}},
+		// One leaf tag.
+		{"AddLeafTag", "", key2, false, true, []int{2, 2, 3, 3, 3}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			prof := testProfile1.Copy()
+			gr, gl := prof.AddPseudoLocs(tc.rootTag, tc.leafTag)
+			if gr != tc.wr || gl != tc.wl {
+				t.Fatalf("Got AddPseudoLocs(%s, %s)=%v,%v want %v,%v", tc.rootTag, tc.leafTag, gr, gl, tc.wr, tc.wl)
```
aalexand, line 56: nit: please follow https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures recommendations for the error string. E.g. function name with inputs typically goes first, without the "Got " prefix.

internal/driver/commands.go
```diff
@@ -199,6 +199,14 @@ var configHelp = map[string]string{
...
 		"Numeric tag filter examples: 1kb, 1kb:10kb, memory=32mb:",
 		"String tag filter examples: foo, foo.*bar, mytag=foo.*bar"),
+	"tagleaf": helpText(
+		"Adds a pseudo leaf location for samples with a matching tag.",
+		"Use name=value syntax to specify the tag and the value for which nodes would be added.",
+		"Example: latency=1s."),
+	"tagroot": helpText(
+		"Adds a pseudo root node for samples with a matching tag.",
+		"Use name=value syntax to specify the tag and the value for which nodes would be added.",
+		"Example: latency=1s."),
```
aalexand, line 209: Is the dot part of the syntax? I assume not which makes this example non copy-pasteable.

internal/driver/commands.go
```diff
@@ -199,6 +199,14 @@ var configHelp = map[string]string{
 		"Use name=value syntax to limit the matching to a specific tag.",
 		"Numeric tag filter examples: 1kb, 1kb:10kb, memory=32mb:",
 		"String tag filter examples: foo, foo.*bar, mytag=foo.*bar"),
+	"tagleaf": helpText(
+		"Adds a pseudo leaf location for samples with a matching tag.",
+		"Use name=value syntax to specify the tag and the value for which nodes would be added.",
```
aalexand, line 204: Can we make sure and clarify that the same syntax as for tagfocus / tagignore can be used for values? E.g. numeric values with units, range and so on. And would be good to reflect this in the examples, like the documentation for other tag filters does.

profile/pseudo_locs_test.go
nilayvaish: Done

profile/pseudo_locs_test.go
nilayvaish: Done

profile/pseudo_locs.go
nilayvaish: Done

internal/driver/driver.go
nilayvaish: For function names, it seems abbreviation is not in use.  For variable names, the abbreviated usage exists in multiple files.  I have updated the function names to use the full form.

internal/driver/driver.go
nilayvaish: Removed the use of word "pseudo" in  functions and file names.

internal/driver/commands.go
nilayvaish: Dropped dot.

profile/pseudo_locs.go
nilayvaish: Done.

internal/driver/commands.go
nilayvaish: I have updated the description.  There is difference between tagfocus/tagignore syntax and the one used for tagroot/tagleaf (as I understand).  tagroot and tagleaf add locations when specific tags exist in a sample, there values are immaterial (though they would appear in the added location).

internal/driver/driver.go
nilayvaish: Can you clarify what you mean by internal?  Should I move this PR to a Google-internal CL?

mhansen:
This PR was towards #558 (I'm linking them up by referencing, because I think I just reimplemented this without knowing it was an open PR solving the same bug).

mhansen:
I think this PR is obsolete; implemented in https://github.com/google/pprof/pull/649

PR #628, opened 2021-05-06
Source listing handles missing binaries and addresses.

ghemawat:
Some profiles can contain locations without an address, or reference
binaries/libraries that are not available. Treat these better when
producing source listings by using any source information present
directly in the profile.Location.

Details:

If a location does not have an address, create a synthetic address
that does not fall in a known mapping.

If a location references an object file that is not found, use any
file/line information baked directly in the location.

Tweaked generated html to clicking for source lines that have no
inlining/assembly to display.

### DIFF
--- a/internal/report/source.go
+++ b/internal/report/source.go
@@ -132,6 +132,7 @@ func printWebSource(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
 // sourcePrinter holds state needed for generating source+asm HTML listing.
 type sourcePrinter struct {
 	reader     *sourceReader
+	synth      *synthCode
 	objectTool plugin.ObjTool
 	objects    map[string]plugin.ObjFile  // Opened object files
 	sym        *regexp.Regexp             // May be nil
@@ -144,6 +145,12 @@ type sourcePrinter struct {
 
 	// Mapping from system function names to printable names.
 	prettyNames map[string]string
+}
+
+// addrInfo holds information for an address we are interested in.
+type addrInfo struct {
+	loc *profile.Location // Always non-nil
+	obj plugin.ObjFile    // May be nil
 }
 
 // instructionInfo holds collected information for an instruction.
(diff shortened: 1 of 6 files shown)


internal/report/source.go
```diff
@@ -270,15 +280,21 @@ func newSourcePrinter(rpt *Report, obj plugin.ObjTool, sourcePath string) *sourc
...
 			}
 
-			cum[loc.Address] += value
+			addr := loc.Address
+			if addr == 0 {
+				// Some profiles are missing valid addresses.
+				addr = sp.synth.address(loc)
+			}
+
+			cum[addr] += value
```
aalexand, line 289: This is pre-existing and doesn't have to be changed (if it needs to be changed) in this CL, but: do we handle recursion correctly here? It feels that we'll double count the cumulative time if the same address is present multiple times in the stack.

internal/report/synth.go
```diff
@@ -0,0 +1,36 @@
...
+		if s.next < m.Limit {
+			s.next = m.Limit
+		}
+	}
+	return s
+}
+
+// address returns the synthetic address for loc, creating one if needed.
+func (s *synthCode) address(loc *profile.Location) uint64 {
+	if addr, ok := s.addr[loc]; ok {
```
aalexand, line 26: Maybe

```
if loc.Address != 0 {
  panic("can only synthesize zero-address locations")
}
```

?

internal/report/synth.go
```diff
@@ -0,0 +1,36 @@
...
+type synthCode struct {
+	next uint64
+	addr map[*profile.Location]uint64 // Synthesized address assigned to a location
+}
+
+func newSynthCode(mappings []*profile.Mapping) *synthCode {
+	// Find a larger address than any mapping.
+	s := &synthCode{next: 1}
+	for _, m := range mappings {
+		if s.next < m.Limit {
```
aalexand, line 17: I saw in the past profiles with mapping.Limit == ~0ULL set, but I think the combination of having both zero address locations AND such fake mapping in the same profile is very unlikely, so we are probably good.

internal/report/source.go
ghemawat: I think you are right. Let me look at that in a separate change.

internal/report/synth.go
ghemawat: Done.

internal/report/source.go
raidancampbell: Hate to butt in here, but is there a chance that this is related to #720?

I'm not familiar with this code, but a flavor of #720 manifested for me with the following conditions:
 - using `weblist` instead of `list`
 - providing the binary for disassembly

and had the following behavior:
 - in a sample function that had 4 caller locations, the count was 5x what the expected value (e.g. that returned by `list` or `weblist` without the binary) was.
 - the printed assembly was printed 5 times
image


As far as I can tell, that fits the bill for the code in question here: the call tree for this function is only invoked for web printing and not the CLI interactions

PR #631, opened 2021-05-11
internal/report: error if filter argument has no matches

### LINKED ISSUE #629: No warning if filter argument has no matches.

If you run pprof from GitHub, what's the Git revision?

923b5ab0fc1a67b7d1902a06c66233c47e099120 (latest HEAD)

Linux, amd64

### What did you do?

If possible, provide a recipe for reproducing the error.
Attaching a profile you are trying to analyze is good.

Attempt to `disasm`, `list`, `peek` a symbol that doesn't exist.

An explanation that there are no matching symbols. Something like:

```
$ ./pprof /tmp/cpu.pprof
File: runtime.test
Type: cpu
Time: May 11, 2021 at 12:26pm (EDT)
Duration: 12.01s, Total samples = 36.44s (303.40%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) disasm blargity
no matches found for regexp: blargity 
```

### What did you see instead?

```
$ ./pprof /tmp/cpu.pprof
File: runtime.test
Type: cpu
Time: May 11, 2021 at 12:26pm (EDT)
Duration: 12.01s, Total samples = 36.44s (303.40%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) disasm blargity
Total: 36.44s
```

To a user that doesn't fully understand this behavior, it is rather confusing to only get a "total" with no other output.

Note that commands with a focus argument do display a warning (in red):

```
(pprof) top blargity
Focus expression matched no samples
Active filters:
   focus=blargity
Showing nodes accounting for 0, 0% of 36.44s total
```

I can send a simple CL to add a warning to the few commands with an argument.

prattmic:
list, weblist, disasm, and peek all take a regexp argument that acts as
a filter.

Currently if this filter results in no matches we simply generate an
empty report, which can be confusing to users. Instead, explicitly treat
this as an error condition that is clearly reported to users.

Fixes #629

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -432,6 +432,10 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 		}
 	}
 
+	if len(syms) == 0 {
+		return fmt.Errorf("no matches found for regexp: %s", o.Symbol)
+	}
+
 	// Correlate the symbols from the binary with the profile samples.
 	for _, s := range syms {
 		sns := symNodes[s]
@@ -1054,6 +1058,7 @@ func printTree(w io.Writer, rpt *Report) error {
 	var flatSum int64
 
 	rx := rpt.options.Symbol
+	matched := 0
 	for _, n := range g.Nodes {
 		name, flat, cum := n.Info.PrintableName(), n.FlatValue(), n.CumValue()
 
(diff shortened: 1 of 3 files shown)


internal/report/report.go
```diff
@@ -432,6 +432,10 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 		}
 	}
 
+	if len(syms) == 0 {
+		return fmt.Errorf("no matches found for regexp: %v", o.Symbol)
```
aalexand, line 436: Here and in other error messages: %v -> %s? Using narrower format specifier might help catch mistakes (such as passing a wrong variable).

prattmic:
PTAL

aalexand (approved):

PR #638, opened 2021-06-21
add AddCommand to dirver

### LINKED ISSUE #637: output the report through UI

    GO111MODULE="auto"
    GOARCH="amd64"
    GOBIN=""
    GOCACHE="/Users/ianwoolf/Library/Caches/go-build"
    GOENV="/Users/ianwoolf/Library/Application Support/go/env"
    GOEXE=""
    GOFLAGS=""
    GOHOSTARCH="amd64"
    GOHOSTOS="darwin"
    GOINSECURE=""
    GOMODCACHE="/Users/ianwoolf/code/go/pkg/mod"
    GOOS="darwin"
    GOPATH="/Users/ianwoolf/code/go"
    GOPROXY="https://goproxy.io,direct"
    GOROOT="/usr/local/go"
    GOSUMDB="sum.golang.org"
    GOTMPDIR=""
    GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
    GOVCS=""
    GOVERSION="go1.16.3"
    GCCGO="gccgo"
    AR="ar"
    CC="clang"
    CXX="clang++"
    CGO_ENABLED="1"
    GOMOD="/Users/ianwoolf/code/go/src/github.com/google/pprof/go.mod"
    CGO_CFLAGS="-g -O2"
    CGO_CPPFLAGS=""
    CGO_CXXFLAGS="-g -O2"
    CGO_FFLAGS="-g -O2"
    CGO_LDFLAGS="-g -O2"
    PKG_CONFIG="pkg-config"
    GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/3f/x8jl0lgx629d_5bz0537gxnm0000gn/T/go-build3760221418=/tmp/go-build -gno-record-gcc-switches -fno-common"

Linux, but this issue is agnostic of operating system

### What did you do?

The system configuration causes some system calls to occupy too much cpu. So I want to expose the cpu analysis results of pprof (such as top) to facilitate monitoring, statistics and other purposes. But unfortunately, `pprof/driver` did not expose `internal/driver.AddCommand`, which made it impossible to output the result through `Options.UI`.

detail:

   https://github.com/google/pprof/blob/master/internal/driver/commands.go#L93
   https://github.com/google/pprof/blob/master/internal/driver/driver.go#L134

In our internal version of go, we resolved this problem. The monitoring system will periodically grab the cpu analysis results of the online service through http for monitoring and performance optimization.

I can call `driver.AddCommand` to set the `visualizer`  of a command, so i can output the result throught the interface `option.UI`.

or a func like `driver.UpdateCommand` is better?

### What did you see instead?

the `visualizer` of all command is nil，so the output is `os.Stdout`. i cannot output the profile reporter through `UI`.

ianwoolf:
fix #637


ianwoolf:
the `dirver.UI` is same as `internal/plugin.UI`， but `internal/driver.PostProcessor` require `internal/plugin.UI`. In my own go version, just set the type `dirver.UI` as `internal/plugin.UI`, and everything is ok. 

But I don’t know if there is a better way, such as put the `UI` interface into a new package like `Option`.

driver/driver.go
```diff
@@ -197,34 +197,7 @@ type Sym struct {
...
-	IsTerminal() bool
-
-	// WantBrowser indicates whether browser should be opened with the -http option.
-	WantBrowser() bool
-
-	// SetAutoComplete instructs the UI to call complete(cmd) to obtain
-	// the auto-completion of cmd, if the UI supports auto-completion at all.
-	SetAutoComplete(complete func(string) string)
-}
+type UI = plugin.UI
```
aalexand, line 200: This change seems unrelated so please remove it from this PR if it's not needed.

aalexand: Ah, looks like you say it's needed. OK, I'll look more.

ianwoolf:
@aalexand   thanks for your review.          is this pr can be accepted?

aalexand:
> @aalexand thanks for your review. is this pr can be accepted?

I haven't looked at it yet. Overall, we do not support any public API in pprof beyond the `profile` package and the command-line interface, so I'm a bit hesitant to make this change.

Nevertheless, some requests for now (as I'll look more at this change):
* There are typos in the PR title and description - could you please fix them?
* Please add documentation to the added public method.
* Do you have example code of how this new API is used?

ianwoolf:
First of all, `internal/driver.AddCommand` does not seem to be called. And because it is under the package `internal`, it cannot be called outside the package.

For example, in this scenario, the profile data of the online service needs to be read and sent to other services, such as monitoring system or database like elasticsearch. But because the `visualizer` of command `top` is written as nil, I cannot get profile data such as cpu through Option.UI interface. So I may need to execute the following code.

	type output struct {
		// r *bufio.Reader
		msg *bytes.Buffer
	}
	func (ui *output) Print(args ...interface{}) {
		...
	}
	...
	func (ui *output) SetAutoComplete(func(string) string) {
	}
	
	type fetcher struct {
		data []byte
	}
	func (f *fetcher) Fetch() {
		...
	}
	
	profData, _ := ioutil.ReadFile("/Users/ianwoolf/Downloads/profile")
	myfetch := fetcher{
		data: profData,
	}
	
	ui := output{
		msg: bytes.NewBufferString(""),
	}
	
	options := &driver.Options{
		UI:    &ui,
		Fetch: &myfetch,
	}
	
	
	f := func(input io.Reader, output io.Writer, ui driver.UI) error {
		msg, _ := ioutil.ReadAll(input)
		ui.Print(string(msg))
		return nil
	}
	
	driver.AddCommand("top", driver.Text, f, "Outputs top entries in text form", "top usage")
	
	if err := driver.PProf(options); err != nil {
		fmt.Fprintf(os.Stderr, "%v\n", err)
		os.Exit(2)
	}

aalexand:
This seems convoluted. Can you instead run `pprof` command as a child process redirecting its output to a pipe that your program reads from?

ianwoolf:
> This seems convoluted. Can you instead run `pprof` command as a child process redirecting its output to a pipe that your program reads from?

it is executalbe. but it is too complex to me, and require pprof and run some command.

PR #641, opened 2021-07-20
Add support for matching split mappings to segments.

gmarin13:
Update the heuristic that matches mappings to segments to support split mappings.

The existing logic assumes a mapping is at least as large as the program segment. However, the loader, or some managing service, may split a program segment into multiple mappings for different reasons:
- to map part of the segment to huge pages, while other parts use regular sized pages;
- the last page may be remapped with write protection to zero out the bytes after the segment end in the last partial mapping page;
- if the segment includes both initialized and uninitialized data.

The list is not exhaustive. While https://github.com/google/perf_data_converter tries to combine associated split mappings into a single mapping, that heuristic also has limitations. Being able to match split mappings to segments increases symbolization coverage.

### DIFF
--- a/internal/binutils/binutils_test.go
+++ b/internal/binutils/binutils_test.go
@@ -808,7 +808,7 @@ func TestELFObjAddr(t *testing.T) {
 		wantAddrError        bool
 	}{
 		{"exec mapping, good address", 0x5400000, 0x5401000, 0, false, 0x5400400, 0x400400, false},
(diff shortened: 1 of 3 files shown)


aalexand (commented):
For posterity, would it be possible to mention in the PR description the circumstances when this change is needed?

gmarin13:
> For posterity, would it be possible to mention in the PR description the circumstances when this change is needed?

Updated the PR description to better explain the problem we try to solve.

aalexand (approved):

PR #644, opened 2021-08-04
When matching mappings to segments, skip over segments with zero file size

gmarin13:
Segments with no file bits can have any file offset value, which can make them
appear to overlap other segments that start at those offsets.

This change prevents us from matching any mapping to such segments with zero
file content. But such segments would be loaded into anonymous mappings, which
we cannot symbolize anyway.


aalexand (approved):

gmarin13:
This will need a new approval I think.

PR #649, opened 2021-09-22
Add -tagroot and -tagleaf options

### LINKED ISSUE #558: Support visualizing tags as nodes

mhansen:
These add synthetic stack frames to samples.

Example usage:

```
$ pprof -tagroot thread pprof.proto
```

Will add synthetic stack frames at the root like `thread:UIThread` and
`thread:BackgroundPool-1`.

Closes #558

### DIFF
--- a/internal/driver/commands.go
+++ b/internal/driver/commands.go
@@ -189,6 +189,14 @@ var configHelp = map[string]string{
 		"Drops functions above the highest matched frame.",
 		"If set, all frames above the highest match are dropped from every sample.",
 		"Matching includes the function name, filename or object name."),
+	"tagroot": helpText(
+		"Adds pseudo stack frames for labels key/value pairs at the callstack root.",
+		"A comma-separated list of label keys.",
+		"The first key creates frames at the new root."),
+	"tagleaf": helpText(
+		"Adds pseudo stack frames for labels key/value pairs at the callstack leaf.",
+		"A comma-separated list of label keys.",
+		"The last key creates frames at the new leaf."),
 	"tagfocus": helpText(
 		"Restricts to samples with tags in range or matched by regexp",
 		"Use name=value syntax to limit the matching to a specific tag.",
--- a/internal/driver/config.go
+++ b/internal/driver/config.go
@@ -30,6 +30,10 @@ type config struct {
 	Normalize           bool    `json:"normalize,omitempty"`
 	Sort                string  `json:"sort,omitempty"`
 
+	// Label pseudo stack frame generation options
+	TagRoot string `json:"tagroot,omitempty"`
+	TagLeaf string `json:"tagleaf,omitempty"`
+
 	// Filtering options
 	DropNegative bool    `json:"drop_negative,omitempty"`
 	NodeCount    int     `json:"nodecount,omitempty"`
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -73,6 +73,10 @@ func generateRawReport(p *profile.Profile, cmd []string, cfg config, o *plugin.O
 
 	cfg = applyCommandOverrides(cmd[0], c.format, cfg)
 
+	// Create label pseudo nodes before filtering, in case the filters use
+	// the generated nodes.
+	generateTagRootsLeaves(p, cfg, o.UI)
+
 	// Delay focus after configuring report to get percentages on all samples.
 	relative := cfg.RelativePercentages
 	if relative {
@@ -206,6 +210,25 @@ func applyCommandOverrides(cmd string, outputFormat int, cfg config) config {
 		cfg.EdgeFraction = 0
 	}
 	return cfg
+}
+
+// generateTagRootsLeaves generates extra nodes from the tagroot and tagleaf options.
+func generateTagRootsLeaves(prof *profile.Profile, cfg config, ui plugin.UI) {
+	tagRootLabelKeys := dropEmptyStrings(strings.Split(cfg.TagRoot, ","))
+	tagLeafLabelKeys := dropEmptyStrings(strings.Split(cfg.TagLeaf, ","))
+	rootm, leafm := addLabelNodes(prof, tagRootLabelKeys, tagLeafLabelKeys, cfg.Unit)
+	warnNoMatches(cfg.TagRoot == "" || rootm, "TagRoot", ui)
+	warnNoMatches(cfg.TagLeaf == "" || leafm, "TagLeaf", ui)
+}
+
+// dropEmptyStrings filters a slice to only non-empty strings
+func dropEmptyStrings(in []string) (out []string) {
+	for _, s := range in {
+		if s != "" {
+			out = append(out, s)
+		}
+	}
+	return
 }
 
 func aggregate(prof *profile.Profile, cfg config) error {
--- a/internal/driver/tagroot.go
+++ b/internal/driver/tagroot.go
@@ -1,0 +1,124 @@
+package driver
+
+import (
+	"strings"
+
+	"github.com/google/pprof/internal/measurement"
+	"github.com/google/pprof/profile"
+)
+
+// addLabelNodes adds pseudo stack frames "label:value" to each Sample with
+// labels matching the supplied keys.
+//
+// rootKeys adds frames at the root of the callgraph (first key becomes new root).
+// leafKeys adds frames at the leaf of the callgraph (last key becomes new leaf).
+//
+// Returns whether there were matches found for the label keys.
+func addLabelNodes(p *profile.Profile, rootKeys, leafKeys []string, outputUnit string) (rootm, leafm bool) {
+	// Find where to insert the new locations and functions at the end of
+	// their ID spaces.
+	var maxLocID uint64
+	var maxFunctionID uint64
+	for _, loc := range p.Location {
+		if loc.ID > maxLocID {
+			maxLocID = loc.ID
+		}
+	}
+	for _, f := range p.Function {
+		if f.ID > maxFunctionID {
+			maxFunctionID = f.ID
+		}
+	}
+	nextLocID := maxLocID + 1
+	nextFuncID := maxFunctionID + 1
+
+	// Intern the new locations and functions we are generating.
+	type locKey struct {
+		functionName, fileName string
+	}
+	locs := map[locKey]*profile.Location{}
+
+	internLoc := func(locKey locKey) *profile.Location {
+		loc, found := locs[locKey]
+		if found {
+			return loc
+		}
+
+		function := &profile.Function{
+			ID:       nextFuncID,
+			Name:     locKey.functionName,
+			Filename: locKey.fileName,
+		}
+		nextFuncID++
+		p.Function = append(p.Function, function)
+
+		loc = &profile.Location{
+			ID: nextLocID,
+			Line: []profile.Line{
+				{
+					Function: function,
+				},
+			},
+		}
+		nextLocID++
+		p.Location = append(p.Location, loc)
+		locs[locKey] = loc
+		return loc
+	}
+
+	makeLabelLocs := func(s *profile.Sample, keys []string) ([]*profile.Location, bool) {
+		var locs []*profile.Location
+		var match bool
+		for i := range keys {
+			// Loop backwards, ensuring the first tag is closest to the root,
+			// and the last tag is closest to the leaves.
+			k := keys[len(keys)-1-i]
+			values := formatLabelValues(s, k, outputUnit)
+			if len(values) > 0 {
+				match = true
+			}
+			locKey := locKey{
+				functionName: strings.Join(values, ","),
+				fileName:     k,
+			}
+			loc := internLoc(locKey)
+			locs = append(locs, loc)
+		}
+		return locs, match
+	}
+
+	for _, s := range p.Sample {
+		rootsToAdd, sampleMatchedRoot := makeLabelLocs(s, rootKeys)
+		if sampleMatchedRoot {
+			rootm = true
+		}
+		leavesToAdd, sampleMatchedLeaf := makeLabelLocs(s, leafKeys)
+		if sampleMatchedLeaf {
+			leafm = true
+		}
+
+		var newLocs []*profile.Location
+		newLocs = append(newLocs, leavesToAdd...)
+		newLocs = append(newLocs, s.Location...)
+		newLocs = append(newLocs, rootsToAdd...)
+		s.Location = newLocs
+	}
+	return
+}
+
+// formatLabelValues returns all the string and numeric labels in Sample, with
+// the numeric labels formatted according to outputUnit.
+func formatLabelValues(s *profile.Sample, k string, outputUnit string) []string {
+	var values []string
+	values = append(values, s.Label[k]...)
+	numLabels := s.NumLabel[k]
+	numUnits := s.NumUnit[k]
+	if len(numLabels) != len(numUnits) {
+		return values
+	}
+	for i, numLabel := range numLabels {
+		unit := numUnits[i]
+		values = append(values, measurement.ScaledLabel(numLabel, unit, outputUnit))
+	}
+	return values
+}
--- a/internal/driver/tagroot_test.go
+++ b/internal/driver/tagroot_test.go
@@ -1,0 +1,368 @@
+package driver
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/google/pprof/internal/proftest"
+	"github.com/google/pprof/profile"
+)
+
+const mainBinary = "/bin/main"
+
+var cpuF = []*profile.Function{
+	{ID: 1, Name: "main", SystemName: "main", Filename: "main.c"},
+	{ID: 2, Name: "foo", SystemName: "foo", Filename: "foo.c"},
+	{ID: 3, Name: "foo_caller", SystemName: "foo_caller", Filename: "foo.c"},
+}
+
+var cpuM = []*profile.Mapping{
+	{
+		ID:              1,
+		Start:           0x10000,
+		Limit:           0x40000,
+		File:            mainBinary,
+		HasFunctions:    true,
+		HasFilenames:    true,
+		HasLineNumbers:  true,
+		HasInlineFrames: true,
+	},
+	{
+		ID:              2,
+		Start:           0x1000,
+		Limit:           0x4000,
+		File:            "/lib/lib.so",
+		HasFunctions:    true,
+		HasFilenames:    true,
+		HasLineNumbers:  true,
+		HasInlineFrames: true,
+	},
+	{
+		ID:              3,
+		Start:           0x4000,
+		Limit:           0x5000,
+		File:            "/lib/lib2_c.so.6",
+		HasFunctions:    true,
+		HasFilenames:    true,
+		HasLineNumbers:  true,
+		HasInlineFrames: true,
+	},
+	{
+		ID:              4,
+		Start:           0x5000,
+		Limit:           0x9000,
+		File:            "/lib/lib.so_6 (deleted)",
+		HasFunctions:    true,
+		HasFilenames:    true,
+		HasLineNumbers:  true,
+		HasInlineFrames: true,
+	},
+}
+
+var cpuL = []*profile.Location{
+	{
+		ID:      1000,
+		Mapping: cpuM[1],
+		Address: 0x1000,
+		Line: []profile.Line{
+			{Function: cpuF[0], Line: 1},
+		},
+	},
+	{
+		ID:      2000,
+		Mapping: cpuM[0],
+		Address: 0x2000,
+		Line: []profile.Line{
+			{Function: cpuF[1], Line: 2},
+			{Function: cpuF[2], Line: 1},
+		},
+	},
+	{
+		ID:      3000,
+		Mapping: cpuM[0],
+		Address: 0x3000,
+		Line: []profile.Line{
+			{Function: cpuF[1], Line: 2},
+			{Function: cpuF[2], Line: 1},
+		},
+	},
+	{
+		ID:      3001,
+		Mapping: cpuM[0],
+		Address: 0x3001,
+		Line: []profile.Line{
+			{Function: cpuF[2], Line: 2},
+		},
+	},
+	{
+		ID:      3002,
+		Mapping: cpuM[0],
+		Address: 0x3002,
+		Line: []profile.Line{
+			{Function: cpuF[2], Line: 3},
+		},
+	},
+}
+
+var testProfile1 = &profile.Profile{
+	TimeNanos:     10000,
+	PeriodType:    &profile.ValueType{Type: "cpu", Unit: "milliseconds"},
+	Period:        1,
+	DurationNanos: 10e9,
+	SampleType: []*profile.ValueType{
+		{Type: "samples", Unit: "count"},
+		{Type: "cpu", Unit: "milliseconds"},
+	},
+	Sample: []*profile.Sample{
+		{
+			Location: []*profile.Location{cpuL[0]},
+			Value:    []int64{1000, 1000},
+			Label: map[string][]string{
+				"key1": {"tag1"},
+				"key2": {"tag1"},
+			},
+		},
+		{
+			Location: []*profile.Location{cpuL[1], cpuL[0]},
+			Value:    []int64{100, 100},
+			Label: map[string][]string{
+				"key1": {"tag2"},
+				"key3": {"tag2"},
+			},
+		},
+		{
+			Location: []*profile.Location{cpuL[2], cpuL[0]},
+			Value:    []int64{10, 10},
+			Label: map[string][]string{
+				"key1": {"tag3"},
+				"key2": {"tag2"},
+			},
+			NumLabel: map[string][]int64{
+				"allocations": {1024},
+			},
+			NumUnit: map[string][]string{
+				"allocations": {""},
+			},
+		},
+		{
+			Location: []*profile.Location{cpuL[3], cpuL[0]},
+			Value:    []int64{10000, 10000},
+			Label: map[string][]string{
+				"key1": {"tag4"},
+				"key2": {"tag1"},
+			},
+			NumLabel: map[string][]int64{
+				"allocations": {1024, 2048},
+			},
+			NumUnit: map[string][]string{
+				"allocations": {"bytes", "b"},
+			},
+		},
+		{
+			Location: []*profile.Location{cpuL[4], cpuL[0]},
+			Value:    []int64{1, 1},
+			Label: map[string][]string{
+				"key1": {"tag4"},
+				"key2": {"tag1", "tag5"},
+			},
+			NumLabel: map[string][]int64{
+				"allocations": {1024, 1},
+			},
+			NumUnit: map[string][]string{
+				"allocations": {"byte", "kilobyte"},
+			},
+		},
+	},
+	Location: cpuL,
+	Function: cpuF,
+	Mapping:  cpuM,
+}
+
+func TestAddLabelNodesMatchBooleans(t *testing.T) {
+	type addLabelNodesTestcase struct {
+		name             string
+		tagroot, tagleaf []string
+		outputUnit       string
+		rootm, leafm     bool
+		// wantSampleFuncs contains expected stack functions and sample value after
+		// adding nodes, in the same order as in the profile. The format is as
+		// returned by stackCollapse function, which is "callee caller: <num>".
+		wantSampleFuncs []string
+	}
+	for _, tc := range []addLabelNodesTestcase{
+		{
+			name: "Without tagroot or tagleaf, add no extra nodes, and should not match",
+			wantSampleFuncs: []string{
+				"main(main.c) 1000",
+				"main(main.c);foo(foo.c);foo_caller(foo.c) 100",
+				"main(main.c);foo(foo.c);foo_caller(foo.c) 10",
+				"main(main.c);foo_caller(foo.c) 10000",
+				"main(main.c);foo_caller(foo.c) 1",
+			},
+		},
+		{
+			name:    "Keys that aren't found add empty nodes, and should not match",
+			tagroot: []string{"key404"},
+			tagleaf: []string{"key404"},
+			wantSampleFuncs: []string{
+				"(key404);main(main.c);(key404) 1000",
+				"(key404);main(main.c);foo(foo.c);foo_caller(foo.c);(key404) 100",
+				"(key404);main(main.c);foo(foo.c);foo_caller(foo.c);(key404) 10",
+				"(key404);main(main.c);foo_caller(foo.c);(key404) 10000",
+				"(key404);main(main.c);foo_caller(foo.c);(key404) 1",
+			},
+		},
+		{
+			name:    "tagroot adds nodes for key1 and reports a match",
+			tagroot: []string{"key1"},
+			rootm:   true,
+			wantSampleFuncs: []string{
+				"tag1(key1);main(main.c) 1000",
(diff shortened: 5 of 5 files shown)


internal/driver/commands.go
```diff
@@ -189,6 +189,14 @@ var configHelp = map[string]string{
 		"Drops functions above the highest matched frame.",
 		"If set, all frames above the highest match are dropped from every sample.",
 		"Matching includes the function name, filename or object name."),
+	"tagroot": helpText(
+		"Adds psuedo-stack-frames for labels key/value pairs at the callstack root.",
```
aalexand, line 193: s/psuedo/pseudo/

same below

internal/driver/commands.go
```diff
@@ -189,6 +189,14 @@ var configHelp = map[string]string{
...
+		"A comma-separated list of label keys.",
+		"The first key creates frames at the new root."),
```
aalexand, line 195: Could we clarify what happens with the samples that don't have the specified tag?

I think my expectation would be that:
* The levels corresponding to the specified tags will still be added
* No sample is dropped, instead they are attributed to a fake value like "foo:<unknown>" where "foo" is the tag name.

profile/label_nodes.go
```diff
@@ -0,0 +1,97 @@
...
+	nextLocID := maxLocID + 1
+	nextFuncID := maxFunctionID + 1
+
+	// Intern the new frames we are generating.
+	locsByName := map[string]*Location{}
+	functionsByName := map[string]*Function{}
+
+	internFunction := func(functionName string) *Function {
+		function, found := functionsByName[functionName]
+		if !found {
```
aalexand, line 34: I would

```
if found {
  return function
}
...
return function
```

to unindent

profile/label_nodes.go
```diff
@@ -0,0 +1,97 @@
+package profile
```
aalexand, line 1: RE label_nodes.go - it would be nice to avoid the underscore if possible, via a shorter name. Vast majority of other files in pprof codebase don't have an underscore (other than _test.go files).

profile/label_nodes.go
```diff
@@ -0,0 +1,97 @@
...
+
```
aalexand, line 2: Could we move this file under internal/ directory? We limit the API surface of the pprof package and mainly expose the tool itself to the users. So I'd prefer not to expose any new public API in this PR.

internal/driver/config.go
```diff
@@ -30,6 +30,10 @@ type config struct {
 	Normalize           bool    `json:"normalize,omitempty"`
 	Sort                string  `json:"sort,omitempty"`
 
+	// Label Psuedo-stack-frame generation options
```
aalexand, line 33: "pseudo" too

Also nit: lowercase (not a proper name)

profile/label_nodes.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	nextLocID := maxLocID + 1
+	nextFuncID := maxFunctionID + 1
+
+	// Intern the new frames we are generating.
+	locsByName := map[string]*Location{}
+	functionsByName := map[string]*Function{}
+
+	internFunction := func(functionName string) *Function {
```
aalexand, line 32: Maybe internFunction and internLocation could be a single intern() function? Both are interned at the same time and the function name is the key.

profile/label_nodes.go
```diff
@@ -0,0 +1,97 @@
...
+		return location
+	}
+
+	for _, s := range p.Sample {
+		rootsToAdd := []*Location{}
+		leafsToAdd := []*Location{}
+		for i := range rootKeys {
+			// Loop backwards, ensuring the first tag is the new root.
+			k := rootKeys[len(rootKeys)-1-i]
+			for _, v := range s.Label[k] {
```
aalexand, line 71: We seem to be added different number of frames depending on whether the tag is present and whether it has 1 or more values. Somehow, as a user, from reading the doc for the new flags I expect strictly static number of levels to be added.

aalexand: It looks like numeric labels are not supported. Note that numeric labels have units and we should support those.

internal/driver/commands.go
mhansen: done

profile/label_nodes.go
mhansen: I like that idea, thank you!

internal/driver/config.go
mhansen: thanks, appreciate you pointing it out, i've probably been spelling this wrong all my life :-)

profile/label_nodes.go
mhansen: Right. In an initial draft of this CL I didn't loop over duplicate keys to make separate stack frames; instead I outputted the `strings.Join` version of them: `label1:value1,value2`. Would that be preferable? I really don't know who would use multiple values for the same key, seems like trouble.

I wonder how we should support numeric labels, I don't use them too much. Are you thinking something like `label1:1023kB`? (I hope we don't have to bucket them, that'll be complicated)

profile/label_nodes.go
mhansen: Yeah good call about the public API surface. I'll have a go at this. I'll have to copy over a lot of the test profile code because it'll be in a different module now, but maybe a little copying is OK.

profile/label_nodes.go
mhansen: I'm struggling to think of a shorter name... is `labelnodes.go` OK? `tagnodes.go`? `tagroot.go`? I was going off examples like `driver_focus.go`.

mhansen: Gone with `labelnodes.go` for now, but happy to change to whatever.

aalexand: I'm not a fan of driver_focus.go either - I think I'd either move that code to driver.go or have it as focus.go. The package name is already "driver" there so "driver_" as a prefix stutters IMO. Anyway, not suggesting to change THAT :)

For this PR, wdyt about internal/driver/tagroot.go? It's close to the flag name so I think it makes it easier to relate the code and the functionality. It doesn't cover the leaf case of course but oh well.

internal/driver/commands.go
mhansen: As implemented now:
- a new frame corresponding to the specified tag is *not* added if that sample is missing a tag
- I'm not dropping any samples from the profile

But I think that adding the extra frame even if the tag is absent with something like `foo:` is reasonable. I'll code it up and see how it looks.

profile/label_nodes.go
aalexand: No, not suggesting bucketing, that's too complicated. I was only thinking of `label1:1023kB`, yes. In the general case numeric labels can have high cardinality but there are important use cases where this would be useful such as the alignment label in heapz profiles.

mhansen: I've updated the CL to `strings.Join` repeated key-values.

internal/driver/commands.go
mhansen: I've updated the code to add the extra frame even if the tag doesn't match.

profile/label_nodes.go
mhansen: I've added the numeric units. This was pretty tricky to zip them together, and I think I might have found a bug in `preEncode` too, which I'll file soon.

profile/label_nodes.go
mhansen: Yeah, I considered it, but I think there's an advantage to having them separate: you can early-return the found interned item. I personally think it's a little clearer with the early return, which wouldn't be possible if they were combined. Happy to merge them if you feel any stronger than that.

profile/label_nodes.go
mhansen: #651 is the bug in `preEncode`, I've added a test case that tickles the bug here.

internal/driver/commands.go
```diff
@@ -189,6 +189,14 @@ var configHelp = map[string]string{
 		"Drops functions above the highest matched frame.",
 		"If set, all frames above the highest match are dropped from every sample.",
 		"Matching includes the function name, filename or object name."),
+	"tagroot": helpText(
+		"Adds pseudo-stack-frames for labels key/value pairs at the callstack root.",
```
aalexand, line 193: nit nit: I'd drop dashes - "pseudo stack frames". These seem fine as separate words.

internal/driver/driver.go
```diff
@@ -73,6 +73,9 @@ func generateRawReport(p *profile.Profile, cmd []string, cfg config, o *plugin.O
 
 	cfg = applyCommandOverrides(cmd[0], c.format, cfg)
 
+	// Create label pseudo-nodes before filtering, in case the filters use the generated nodes.
```
aalexand, line 76: nit: please keep comments in 80 columns.

internal/driver/driver.go
```diff
@@ -208,6 +211,25 @@ func applyCommandOverrides(cmd string, outputFormat int, cfg config) config {
 	return cfg
 }
 
+// withoutEmptyStrings filters a slice to only non-empty strings
+func withoutEmptyStrings(in []string) (out []string) {
```
aalexand, line 215: Could you order functions top-down rather than bottom-up? This seems more common in Go as it allows to focus on the high level code logic first when reading the code top down.

internal/driver/driver.go
```diff
@@ -208,6 +211,25 @@ func applyCommandOverrides(cmd string, outputFormat int, cfg config) config {
...
+	for _, s := range in {
+		if s != "" {
+			out = append(out, s)
+		}
+	}
+	return
+}
+
+// generateLabelPseudoNodes generates extra nodes from the tagroot and tagleaf options.
+func generateLabelPseudoNodes(prof *profile.Profile, cfg config, ui plugin.UI) {
```
aalexand, line 225: Not a strong feeling but I'd name this generateTagRootsLeaves to bear resemblance to the flag names.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
+package driver
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/pprof/profile"
+)
+
+// addLabelNodes adds psuedo-stack-frames "label:value" to each Sample with
```
aalexand, line 10: s/psuedo/pseudo/g

:)

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
...
+			// Loop backwards, ensuring the last tag is the new leaf.
+			k := leafKeys[len(leafKeys)-1-i]
+			values := []string{}
+			values = append(values, s.Label[k]...)
+			numLabels := s.NumLabel[k]
+			numUnits := s.NumUnit[k]
+			if len(numLabels) == len(numUnits) {
+				for i, numLabel := range numLabels {
+					unit := numUnits[i]
+					values = append(values, fmt.Sprintf("%d%s", numLabel, unit))
```
aalexand, line 98: Unfortunately, there is one more piece of complexity you'll have to deal with: I think we should scale the numeric tag values in the same way we scale them when printing in other circumstances, such as `printTags` in report.go: see [code](https://github.com/google/pprof/blob/02619b876842e0d0afb5e5580d3a374dad740edb/internal/report/report.go#L688-L690).

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
...
+				}
+			}
+
+			functionName := k + ":" + strings.Join(values, ",")
+			function := internFunction(functionName)
+			location := internLocation(functionName, function)
+			leafsToAdd = append(leafsToAdd, location)
```
aalexand, line 105: nit: leaves?

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
...
+			if len(values) > 0 {
+				leafm = true
+			}
+		}
+		newLocations := []*profile.Location{}
```
aalexand, line 110: nit: var newLocations []*profile.Location

No functional difference, but nil slice is the same as empty slice in Go and it's more idiomatic to declare them as nil slices.

internal/driver/tagroot_test.go
```diff
@@ -0,0 +1,325 @@
...
+	Location: cpuL,
+	Function: cpuF,
+	Mapping:  cpuM,
+}
+
+// sampleFuncs returns a slice of strings where each string represents one
+// profile sample in the format "<fun1> <fun2> <fun3>: <value>". This allows
+// the expected values for test cases to be specified in human-readable
+// strings.
+func sampleFuncs(p *profile.Profile) []string {
```
aalexand, line 186: nit: move after the Test*() function (see the other note on top-down order for functions).

aalexand (commented):
This looks very nice, thanks for adding this! Except the comment on formatting / scaling the numeric tag values this review is a bunch of style comments.

internal/driver/tagroot_test.go
mhansen: Done, thanks

internal/driver/tagroot.go
mhansen: Done, thanks

internal/driver/tagroot.go
mhansen: haha, thanks, let me grep for any more occurrences too

internal/driver/driver.go
mhansen: I like your name better, changing.

internal/driver/driver.go
mhansen: Done

internal/driver/commands.go
mhansen: Done.

internal/driver/tagroot.go
mhansen: Done

internal/driver/driver.go
mhansen: Reordered it below

internal/driver/tagroot.go
mhansen: Thanks for the pointer, very helpful. Done

internal/driver/driver.go
```diff
@@ -208,6 +212,25 @@ func applyCommandOverrides(cmd string, outputFormat int, cfg config) config {
...
+// generateTagRootsLeaves generates extra nodes from the tagroot and tagleaf options.
+func generateTagRootsLeaves(prof *profile.Profile, cfg config, ui plugin.UI) {
+	tagRootLabelKeys := withoutEmptyStrings(strings.Split(cfg.TagRoot, ","))
+	tagLeafLabelKeys := withoutEmptyStrings(strings.Split(cfg.TagLeaf, ","))
+	rootm, leafm := addLabelNodes(prof, tagRootLabelKeys, tagLeafLabelKeys, cfg.Unit)
+	warnNoMatches(cfg.TagRoot == "" || rootm, "TagRoot", ui)
+	warnNoMatches(cfg.TagLeaf == "" || leafm, "TagLeaf", ui)
+}
+
+// withoutEmptyStrings filters a slice to only non-empty strings
```
aalexand, line 224: nit: maybe dropEmptyStrings to start the function name with a verb.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
+package driver
+
+import (
+	"strings"
+
+	"github.com/google/pprof/internal/measurement"
+	"github.com/google/pprof/profile"
+)
+
+// addLabelNodes adds pseudo-stack-frames "label:value" to each Sample with
```
aalexand, line 10: nit: drop dashes in pseudo-stack-frames like elsewhere.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	for _, f := range p.Function {
+		if f.ID > maxFunctionID {
+			maxFunctionID = f.ID
+		}
+	}
+	nextLocID := maxLocID + 1
+	nextFuncID := maxFunctionID + 1
+
+	// Intern the new frames we are generating.
```
aalexand, line 35: I would change "frames" to "locations and functions", I think it's more accurate here.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
...
+		for i := range rootKeys {
+			// Loop backwards, ensuring the first tag is the new root.
+			k := rootKeys[len(rootKeys)-1-i]
+			values := s.Label[k]
+			if len(values) > 0 {
+				rootm = true
+			}
+			functionName := k + ":" + strings.Join(values, ",")
+			function := internFunction(functionName)
+			location := internLocation(functionName, function)
```
aalexand, line 85: I would inline the call like

```
location := internLocation(functionName, internFunction(functionName))
```

to make it very clear the temporary is not used, but this is subjective.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,117 @@
...
+		return location
+	}
+
+	for _, s := range p.Sample {
+		rootsToAdd := []*profile.Location{}
+		leavesToAdd := []*profile.Location{}
+		for i := range rootKeys {
+			// Loop backwards, ensuring the first tag is the new root.
+			k := rootKeys[len(rootKeys)-1-i]
+			values := s.Label[k]
```
aalexand, line 79: Should we handle NumLabel in this loop too?

aalexand (commented):
All are nits except a question about handling NumLabel in the root tag loop.

internal/driver/tagroot.go
mhansen: Done

internal/driver/tagroot.go
mhansen: Done

internal/driver/driver.go
mhansen: Done. I had drop to start with and switched it to 'without' because it was copying not mutating. I was probably overthinking it.

internal/driver/tagroot.go
mhansen: Done

internal/driver/tagroot.go
mhansen: whoops! let me extract a common function

mhansen: and add a test

mhansen: Done!

aalexand (commented):
Apologies about more comments, I should have tried to identify all these things earlier. I think this is very close though.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,120 @@
...
+			},
+		}
+		nextLocID++
+		p.Location = append(p.Location, location)
+		locsByName[functionName] = location
+		return location
+	}
+
+	for _, s := range p.Sample {
+		rootsToAdd := []*profile.Location{}
```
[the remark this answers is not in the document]
mhansen: Done. No worries about more comments, I want this to be right and long-term maintainable.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,120 @@
...
+		newLocations = append(newLocations, s.Location...)
+		newLocations = append(newLocations, rootsToAdd...)
+		s.Location = newLocations
+	}
+	return
+}
+
+// formatLabelValues returns all the string and numeric labels in Sample, with
+// the numeric labels formatted according to outputUnit.
+func formatLabelValues(s *profile.Sample, k string, outputUnit string) (values []string) {
```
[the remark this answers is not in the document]
mhansen: Done.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,120 @@
...
+	values = append(values, s.Label[k]...)
+	numLabels := s.NumLabel[k]
+	numUnits := s.NumUnit[k]
+	if len(numLabels) == len(numUnits) {
```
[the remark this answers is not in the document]
mhansen: great idea!

internal/driver/tagroot.go
```diff
@@ -0,0 +1,120 @@
...
+		nextLocID++
+		p.Location = append(p.Location, location)
+		locsByName[functionName] = location
+		return location
+	}
+
+	for _, s := range p.Sample {
+		rootsToAdd := []*profile.Location{}
+		leavesToAdd := []*profile.Location{}
+		for i := range rootKeys {
```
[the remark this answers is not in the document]
mhansen: > the body of the rootKeys and leafKeys functions is the same

They're very similar... but they're "different enough" in which variables they mutate that I'm not sure it'd be clearer to pull out a function?

I'd have to pass a lot of pointers to variables like `leafm` vs `rootm`, `rootsToAdd` vs `leavesToAdd`, or have a lot of multiple return values (which I'm not sure would be clearer?) That said, I'm mostly a Java programmer day-to-day and maybe it's totally fine and normal to pass pointers in Go so that e.g. slices can be mutated. Is that normal, like to pass a `*bool`?

Maybe I should give it a go; after all, I did have some copy/paste errors here. I'll think a bit and try some more variations.

> sort the keys so that the order is deterministic

Yeah this is a great idea, I had forgotten I'm dealing with hashtables here.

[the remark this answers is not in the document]
mhansen: > sort the keys so that the order is deterministic

Hmm, I think we aren't iterating a hashtable here? (but I might be missing something):
- We loop over `rootKeys, leafKeys` which are `[]string` passed from command line flags
- We look up the map using the keys from `rootKeys, leafKeys` in `formatLabelValues`.

aalexand (dismissed):

aalexand:
@mhansen looks like there are test failures.

internal/driver/tagroot.go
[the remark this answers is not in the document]
mhansen: > the body of the rootKeys and leafKeys functions is the same

I extracted a function, PTAL. I wonder if I might have too many closures here, there's a few intern tables and outputUnit being passed through closure. If this is getting a bit many closures (I'm not sure of Go style) then maybe it's time to make a class.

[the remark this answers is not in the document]
aalexand: I don't seem to see the changes - not pushed yet?

aalexand (dismissed):

mhansen:
Sorry about all the re-pushing and reapprovals necessary. I think I needed to upgrade my go version to get gofmt to work the same as CI.

aalexand:
Looks like the CI tests fail on Go tip. We should probably fix this before we can submit this.

aalexand:
Filed #653.

mhansen:
I've been playing around with this a little, and I'm not so sure any more about prefixing with `labelname:`. On the flamegraph view, where horizontal space is limited, this usually pushes out the interesting bit (the label value) into the ellipsised section, meaning you have to hover to see the value.

image

This isn't a showstopper, I just think it might be suboptimal, there might be a better solution hiding, and maybe we should discuss options.

Let's generate some options:
1. Add two frames, one for the label key, one for the label value. Then we can see both. e.g. `threadpool` in one frame, then below that, `DefaultPool%d`. This makes the graph kinda self-explanatory in a nice way I think, and vertical space is cheap. But then again, the label is kinda meaningless to select, and it will visually associate with the frame above, rather than the frames below, as it's larger.
1. A variation on the above: show one frame `threadpool:DefaultPool%d` above, then `DefaultPool%d` below. This is just like the above, but ensures the label visually groups with the value by making them the same width. It feels a bit hacky to duplicate it though.
1. Only show the label value. Given we're adding frames unconditionally, even if the label value is empty, it's implicit that the first frames will be the label values. There's some prior art here: [Brendan Gregg's ACM Queue flamegraph article](https://queue.acm.org/detail.cfm?id=2927301) shows a flamegraph where [the root has frames like `javac` and `swapper`](https://queue.acm.org/downloads/2016/Gregg7.svg) -- thread comm names.
1. Instead of `labelKey:labelValue` invert the positioning somehow, like `labelValue (labelKey)` or `labelValue [labelKey]`? That might give trouble with regex search. Or we could get creative, `labelValue←labelKey`. Are there any programming languages with keys on the right hand side of hashtables? What's their syntax?
1. Keep `labelKey:labelValue`. This is similar to what felixge's pprofutils does: https://github.com/felixge/pprofutils#example-1-add-root-frames-for-pprof-label-values.

Thoughts? (there's probably other options!) I think I'm leaning towards "Only show the label value" and perhaps "Add two frames, one for key, one for value"

aalexand:
(why is there no way to reply to a comment in GitHub that is not a code review comment?..)

> Thoughts? (there's probably other options!) I think I'm leaning towards "Only show the label value" and perhaps "Add two frames, one for key, one for value"

I need to sit on this a bit longer but my initial reaction is that I don't like two frame approach for how weirdly it will interplay with the filtering - e.g. `-ignore=<tag name>` will non-atomically filter out the frames corresponding to the label name which seems weird to my engineer's mind...

nolanmar511:
> Thoughts? (there's probably other options!) I think I'm leaning towards "Only show the label value" and perhaps "Add two frames, one for key, one for value"

If it would work, I might suggest storing tag value as the ["function name"](https://github.com/google/pprof/blob/02619b876842e0d0afb5e5580d3a374dad740edb/proto/profile.proto#L204) and the tag key as the ["filename"](https://github.com/google/pprof/blob/02619b876842e0d0afb5e5580d3a374dad740edb/proto/profile.proto#L209).  For the flame graph, it should be be roughly equivalent visually "only show the label value", but the function name should still show up in the tooltip and be considered when filtering.

aalexand:
> If it would work, I might suggest storing tag value as the "function name" and the tag key as the "filename". For the flame graph, it should be be roughly equivalent visually "only show the label value", but the function name should still show up in the tooltip and be considered when filtering.

I think it's a great idea to explore!

mhansen:
> If it would work, I might suggest storing tag value as the "function name" and the tag key as the "filename". For the flame graph, it should be be roughly equivalent visually "only show the label value", but the function name should still show up in the tooltip and be considered when filtering.

I think this is a great idea! I've implemented it, and I like it. It looks great in our internal pprof flamegraph viewer (e.g. http://go/pprof-649) where the tooltip shows the filename.

FYI, in the pprof web flamegraph the tooltip doesn't show the filename (yet):

image

I think I still like this approach the best, using the filename, and perhaps we should file a separate issue to show the filename in the tooltip in the web UI's flamegraph.

mhansen:
I should add, thank you for your ideas! This is a fun collaboration for me :-), I'm enjoying the brainstorming

mhansen:
In my latest commit, I've changed the test strings to be https://github.com/brendangregg/FlameGraph "Folded Stack" format, which I'm fairly used to and I think demonstrates the empty stack frames (with semicolon-delimited frames) better than spaces.

And I've gotten rid of the two layers of interning. You were right Alexey, I should have done this a while ago when you suggested it.

internal/driver/tagroot.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, f := range p.Function {
+		if f.ID > maxFunctionID {
+			maxFunctionID = f.ID
+		}
+	}
+	nextLocID := maxLocID + 1
+	nextFuncID := maxFunctionID + 1
+
+	// Intern the new locations and functions we are generating.
+	type locationKey struct {
```
aalexand, line 36: nit: I would universally use "loc" instead of "location" / "locations" in this function, for brevity. `locKey`, `internLoc`, `locs`, `makeLabelLocs` and so on. The context here is fairly local so it's unambiguous enough. You also already have `nextLocID` above.

internal/driver/tagroot_test.go
```diff
@@ -0,0 +1,353 @@
...
+				"main(main.c);foo(foo.c);foo_caller(foo.c);(allocations) 10",
+				"main(main.c);foo_caller(foo.c);1kB,2kB(allocations) 10000",
+				"main(main.c);foo_caller(foo.c);1kB,1kB(allocations) 1",
+			},
+		},
+	} {
+		p := testProfile1.Copy()
+		rootm, leafm := addLabelNodes(p, tc.tagroot, tc.tagleaf, tc.outputUnit)
+		if rootm != tc.rootm {
+			t.Errorf("Test #%d, got rootm=%v, want=%v", tx, rootm, tc.rootm)
```
aalexand, line 320: RE "Test #%d" - maybe use subtests and have a `desc` / `name` field in each test? See https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks.

internal/driver/tagroot_test.go
```diff
@@ -0,0 +1,353 @@
...
+		}
+		if leafm != tc.leafm {
+			t.Errorf("Test #%d, got leafm=%v, want=%v", tx, leafm, tc.leafm)
+		}
+		if got, want := strings.Join(stackCollapse(p), "\n")+"\n", strings.Join(tc.wantSampleFuncs, "\n")+"\n"; got != want {
+			diff, err := proftest.Diff([]byte(want), []byte(got))
+			if err != nil {
+				t.Fatalf("failed to get diff: %v", err)
```
aalexand, line 328: nit: "failed" -> "Failed"

internal/driver/tagroot_test.go
```diff
@@ -0,0 +1,353 @@
...
+// allows the expected values for test cases to be specified in human-readable
+// strings.
+func stackCollapse(p *profile.Profile) []string {
+	var ret []string
+	for _, s := range p.Sample {
+		var funcs []string
+		for i := range s.Location {
+			loc := s.Location[len(s.Location)-1-i]
+			for _, line := range loc.Line {
+				funcs = append(funcs, line.Function.Name+"("+line.Function.Filename+")")
```
aalexand, line 347: Maybe `fmt.Sprintf("%s(%s)", line.Function.Name, line.Function.Filename)` would be a bit more readable but this may be subjective.

aalexand (dismissed):
I think this looks good. A few bikeshedding comments but LGTM otherwise - I think it's ready to be merged.

internal/driver/tagroot.go
mhansen: Done

internal/driver/tagroot_test.go
mhansen: Done

internal/driver/tagroot_test.go
mhansen: Done

internal/driver/tagroot_test.go
mhansen: Done

aalexand (approved):

PR #665, opened 2021-11-03
Avoid adding a redundant newline when printing weblistPageClosing

zpavlinovic:
weblistPageClosing string ends in a new line, so printing it via fmt.Println will add an extra new line.

### DIFF
--- a/internal/driver/testdata/pprof.cpu.flat.addresses.weblist
+++ b/internal/driver/testdata/pprof.cpu.flat.addresses.weblist
@@ -99,4 +99,3 @@ Duration: 10s, Total samples = 1.12s (11.20%)<br>Total: 1.12s</div><h2>line1000<
 
 </body>
 </html>
-
--- a/internal/report/source.go
+++ b/internal/report/source.go
@@ -874,7 +874,7 @@ func printFunctionClosing(w io.Writer) {
 
 // printPageClosing prints the end of the page in a weblist report.
 func printPageClosing(w io.Writer) {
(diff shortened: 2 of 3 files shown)


aalexand (approved):

aalexand (changes requested):
Could you elaborate why is this change needed? What's wrong with Println?

aalexand:
> Could you elaborate why is this change needed? What's wrong with Println?

If this is to avoid a redundant newline, then I think it's easier to update the value of the weblistPageClosing var to remove it there and leave the code as is.

And also the PR description needs to be updated I think. I don't think this change has anything to do with the linked 'go vet' changes. Those changes are about format strings for Printf.

zpavlinovic:
> > Could you elaborate why is this change needed? What's wrong with Println?
> 
> If this is to avoid a redundant newline, then I think it's easier to update the value of the weblistPageClosing var to remove it there and leave the code as is.

Sounds good.

> And also the PR description needs to be updated I think. I don't think this change has anything to do with the linked 'go vet' changes. Those changes are about format strings for Printf.

Yes, it is about avoiding a redundant new line. The reason I mentioned vet is because a previous [limitation](https://github.com/golang/go/issues/30436) has been recently addressed. I will change the description.

aalexand:
The description still mentions "go vet".

zpavlinovic:
> The description still mentions "go vet".

Hopefully, I have corrected it at all the relevant places now.

aalexand (approved):

PR #667, opened 2021-11-04
doc: clarify graph view docs to note negative values appear in profile comparison

### LINKED ISSUE #666: Why cum has negative values？
image

It saids negative cum values turns green.But i cannont understand,why cum value could be negative ?

nolanmar511:
Fixes #666

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -192,7 +192,7 @@ the graph, controlled by the *nodecount* option.
 
 * **Node Color**:
   * large positive cum values are red.
-  * large negative cum values are green.
+  * large negative cum values are green; negative values are most likely to appear during profile comparison, see [this section](https://github.com/google/pprof/tree/master/doc#comparing-profiles) for details.
   * cum values close to zero are grey.
 
 * **Node Font Size**:
--- a/doc/README.md
+++ b/doc/README.md
@@ -192,7 +192,9 @@ the graph, controlled by the *nodecount* option.
 
 * **Node Color**:
   * large positive cum values are red.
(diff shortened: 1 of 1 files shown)


amchiclet (dismissed):

doc/README.md
```diff
@@ -192,7 +192,7 @@ the graph, controlled by the *nodecount* option.
 
 * **Node Color**:
   * large positive cum values are red.
-  * large negative cum values are green.
+  * large negative cum values are green; negative values are most likely to appear during profile comparison, see [this section](https://github.com/google/pprof/tree/master/doc#comparing-profiles) for details.
```
aalexand, line 195: Keep in 80 columns?

Also, should we use a relative link so that it points to the right file when a part of a PR?

aalexand: @nolanmar511 PTAL

doc/README.md
```diff
@@ -192,7 +192,10 @@ the graph, controlled by the *nodecount* option.
 
 * **Node Color**:
   * large positive cum values are red.
-  * large negative cum values are green.
+  * large negative cum values are green; negative values are most likely to
+    appear during profile comparison, see [this section]
+    (https://github.com/google/pprof/tree/master/doc#comparing-profiles) for
```
amchiclet, line 197: I think the newline broke the link (see https://github.com/google/pprof/tree/nolanmar511-patch-1/doc).

And relative links like `[this section](#comparing-profiles)` would work in case this doc was branched.

amchiclet (approved):

PR #669, opened 2021-11-08
Add a folded stacks output format

### LINKED ISSUE #658: FR: Output Folded Stacks Format
I think if we built an exporter to [Brendan Gregg's folded stacks format](https://queue.acm.org/detail.cfm?id=2927301#:~:text=The%20folded%20stack%2Dtrace%20format,trace%2C%20followed%20by%20a%20semicolon.), we could:

1. Enable people to do differential flamegraphs using [difffolded.pl](https://github.com/brendangregg/FlameGraph/blob/master/difffolded.pl) and [flamegraph.pl](https://github.com/brendangregg/FlameGraph/blob/master/flamegraph.pl), perhaps addressing the user need in #483
2. Allow post-processing the profiles using text tools like grep & sed
3. Hook into some more profile analysis tools: see [profilerpedia](https://docs.google.com/spreadsheets/d/1cVcHofphkQqk1yGeuBPVTit8HQ0oa5SlRM6gkHIagtw/edit#gid=0&range=K25), folded stacks is a very well-connected format.

I'm picturing an option like `-callgrind` and `-dot`, perhaps `-folded`.

The trouble might be that folded stacks format only accepts one output number per stack, but pprof can have multiple numbers per stack. So we'd probably have to choose the number to output with a flag, if it was ambiguous. But perhaps parts of pprof already handle this? e.g. the `-dot` output probably chooses an event to use as the weight of the graph size?

Prior art in this space: @felixge's `pprofutils folded`: https://github.com/felixge/pprofutils#folded. Felix has an extension to the folded stacks format to have numbers with multiple columns. I think that's a neat idea for representing pprof and allows roundtripping back to pprof format. It would be nice if this extension caught on, but I wonder if it might limit the interoperability with other folded stacks tools.

Another idea might be to output only one number column per-file (e.g. `pprof -folded=cpu-cycles` and `pprof -folded=instructions`), and then to roundtrip, we could output multiple files, convert both of them back to pprof, then concatenate the files with `pprof -proto <profile1> <profile2>`.

There are plenty of other challenges with roundtripping too (labels, filenames, mappings...) but perhaps just solving part of the problem would still be useful and enable more analyses. Anyway, I mostly wanted to open a discussion about this. Thoughts?

mhansen:
Fixes #658

I'm not sure I've covered everything here; mailing it out early for feedback.

I based this exporter off `printTraces`, removing the extraneous bits like labels and unit formatting.

I copied the sample mean divisor logic from `printTraces`.

This uses the `sample_index` to choose which number to output.

I'm not sure the best way to deal with semicolons or newlines in names. I just removed them for now, although this might lead to colissions. I don't think there's a way to escape them in folded stacks format. Maybe I could just ignore them, and output a corrupted output. That's what a lot of other exporters do. Or say that any profile that contains them is invalid, and return an error. For now, just remove them.

Prior art:
* As usual, credit to @felixge for doing this earlier in https://github.com/felixge/pprofutils.
* https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-go.pl also converts pprof to folded stacks

This connects pprof with visualisation tools that accept the folded stack format: https://profilerpedia.markhansen.co.nz/formats/folded-stacks/


mhansen:
Todo: I should add a test case for same stacks different labels and make sure they're aggregated together. I suspect that's not working right now, that instead it'll output duplicate stacks.

mhansen:
Particularly I'd love some feedback about if there's a different function or something I should use to print the name of the stack. For example, I'd expect `-filefunctions` to show the stack frame name and the file name, but it doesn't seem to with `-format` as coded up today.

internal/report/report.go
```diff
@@ -828,6 +831,55 @@ func printText(w io.Writer, rpt *Report) error {
...
+			// Folded stack convention: start with root frame, end
+			// with leaves.
+			s := stack[len(stack)-i-1]
+			if i > 0 {
+				fmt.Fprint(w, ";")
+			}
+			// TODO: should we print more than just s.Name?
+			// NodeInfo.PrintableName() has a lot more.
+
+			// Remove semicolons and newlines.
```
aalexand, line 872: Add a note on why?

aalexand: Err, didn't mean to click "Add single comment", I'll have more comments!

internal/report/report.go
```diff
@@ -828,6 +831,55 @@ func printText(w io.Writer, rpt *Report) error {
...
+		}
+		// Print call stack.
+		for i := range stack {
+			// Folded stack convention: start with root frame, end
+			// with leaves.
+			s := stack[len(stack)-i-1]
+			if i > 0 {
+				fmt.Fprint(w, ";")
+			}
+			// TODO: should we print more than just s.Name?
```
aalexand, line 869: Yes, I think this needs more discussion. There is an implicit interplay with the granularity. See below. I would expect that we utilize that - i.e. use PrintableName() or NameComponents() directly defining how they should join. This is in particular important because graph.CreateNodes() respects the granularity and so if we don't encode all components in the string, then something like "-folded -filefunctions" may produce duplicate elements.

```
$ pprof -top -nodecount=5 -symbolize=none internal/report/testdata/sample.cpu
File: sample.bin
Type: cpu
Time: Sep 29, 2017 at 1:57pm (PDT)
Duration: 2s, Total samples = 1760ms (87.91%)
Showing nodes accounting for 1640ms, 93.18% of 1760ms total
Showing top 5 nodes out of 14
    1280ms 72.73% 72.73%     1370ms 77.84%  runtime.mapiternext
     150ms  8.52% 81.25%     1760ms   100%  main.busyLoop
     110ms  6.25% 87.50%      110ms  6.25%  math.Abs (inline)
      60ms  3.41% 90.91%       70ms  3.98%  runtime.evacuate
      40ms  2.27% 93.18%       40ms  2.27%  runtime.add (inline)
$ pprof -top -nodecount=5 -symbolize=none -filefunctions internal/report/testdata/sample.cpu
File: sample.bin
Type: cpu
Time: Sep 29, 2017 at 1:57pm (PDT)
Duration: 2s, Total samples = 1760ms (87.91%)
Showing nodes accounting for 1640ms, 93.18% of 1760ms total
Showing top 5 nodes out of 14
    1280ms 72.73% 72.73%     1370ms 77.84%  runtime.mapiternext /usr/lib/google-golang/src/runtime/hashmap.go
     150ms  8.52% 81.25%     1760ms   100%  main.busyLoop internal/report/testdata/sample/sample.go
     110ms  6.25% 87.50%      110ms  6.25%  math.Abs /usr/lib/google-golang/src/math/abs.go (inline)
      60ms  3.41% 90.91%       70ms  3.98%  runtime.evacuate /usr/lib/google-golang/src/runtime/hashmap.go
      40ms  2.27% 93.18%       40ms  2.27%  runtime.add /usr/lib/google-golang/src/runtime/stubs.go (inline)
$ pprof -top -nodecount=5 -symbolize=none -files internal/report/testdata/sample.cpu
File: sample.bin
Type: cpu
Time: Sep 29, 2017 at 1:57pm (PDT)
Duration: 2s, Total samples = 1760ms (87.91%)
Showing nodes accounting for 1740ms, 98.86% of 1760ms total
Showing top 5 nodes out of 8
    1400ms 79.55% 79.55%     1450ms 82.39%  /usr/lib/google-golang/src/runtime/hashmap.go
     150ms  8.52% 88.07%     1760ms   100%  internal/report/testdata/sample/sample.go
     110ms  6.25% 94.32%      110ms  6.25%  /usr/lib/google-golang/src/math/abs.go (inline)
      40ms  2.27% 96.59%      130ms  7.39%  /usr/lib/google-golang/src/runtime/hashmap_fast.go
      40ms  2.27% 98.86%       40ms  2.27%  /usr/lib/google-golang/src/runtime/stubs.go (inline)
$ pprof -top -nodecount=5 -symbolize=none -addresses internal/report/testdata/sample.cpu
File: sample.bin
Type: cpu
Time: Sep 29, 2017 at 1:57pm (PDT)
Duration: 2s, Total samples = 1760ms (87.91%)
Showing nodes accounting for 350ms, 19.89% of 1760ms total
Showing top 5 nodes out of 96
     100ms  5.68%  5.68%      100ms  5.68%  00000000004b4463 math.Abs /usr/lib/google-golang/src/math/abs.go:16
      70ms  3.98%  9.66%       70ms  3.98%  00000000004095d0 runtime.mapiternext /usr/lib/google-golang/src/runtime/hashmap.go:823
      70ms  3.98% 13.64%       70ms  3.98%  00000000004096a2 runtime.mapiternext /usr/lib/google-golang/src/runtime/hashmap.go:856
      60ms  3.41% 17.05%       60ms  3.41%  000000000040957a runtime.mapiternext /usr/lib/google-golang/src/runtime/hashmap.go:819
      50ms  2.84% 19.89%       50ms  2.84%  00000000004095b6 runtime.mapiternext /usr/lib/google-golang/src/runtime/hashmap.go:822
```

internal/report/report_test.go
```diff
@@ -33,9 +33,12 @@ type testcase struct {
 	want string
 }
 
-func TestSource(t *testing.T) {
+func TestTextReports(t *testing.T) {
```
aalexand, line 36: Could we have a separate test for the folded report? I find such combined meta-tests somewhat complicated... If this is to share setup / teardown code then maybe we can share it by other means?

aalexand:
Looks like this PR stalled, closing it. Feel free to re-open if needed.

PR #675, opened 2021-12-01
Parse and propagate the name of the kernel relocation symbol

s-kanev:
For linux kernel mappings generated by recent version of the perf tool, the suffix of the mapping file name (not the name under which we found the actual symbols file) contains the kernel relocation symbol ("_text" or "_stext" in, for example "[kernel.kallsyms]_stext"). The symbol name is required to correctly compute the relocation offset of the kernel mapping.

We add it to Mapping as a public, but non-persisted field, so it can be preserved between profile fetching and symbolization, even if we found the kernel image locally under a different filename.
We also propagate it through the ObjTool interface.

Tested: unit.


s-kanev:
Split the refactoring part from #674. @nolanmar511, could you take a look?

aalexand (commented):
When the binary is located by the build ID via $buildid/$name pattern such as in example in b/168662041, the binary name will be overridden - see [here](https://github.com/google/pprof/blob/2007db6d4f53c44a417ddae675d50f56b8e8c2fd/internal/driver/fetch.go#L429). I believe that happens before the symbolization is done. And I am not sure that this PR ensures that the original name will be stored anywhere, so unless I missed it seems that the proposed solution is somewhat fragile...

s-kanev:
> When the binary is located by the build ID via $buildid/$name pattern such as in example in b/168662041, the binary name will be overridden - see [here](https://github.com/google/pprof/blob/2007db6d4f53c44a417ddae675d50f56b8e8c2fd/internal/driver/fetch.go#L429). I believe that happens before the symbolization is done. And I am not sure that this PR ensures that the original name will be stored anywhere, so unless I missed it seems that the proposed solution is somewhat fragile...

Oh, good catch. The overwriting happens after the successful call to `binutils.openELF()` (the only part that cares about the correct kernel relocation offset) -- which is why this approach works at all in the first place. Agree it's unnecessarily fragile.

aalexand:
> > When the binary is located by the build ID via $buildid/$name pattern such as in example in b/168662041, the binary name will be overridden - see [here](https://github.com/google/pprof/blob/2007db6d4f53c44a417ddae675d50f56b8e8c2fd/internal/driver/fetch.go#L429). I believe that happens before the symbolization is done. And I am not sure that this PR ensures that the original name will be stored anywhere, so unless I missed it seems that the proposed solution is somewhat fragile...
> 
> Oh, good catch. The overwriting happens after the successful call to `binutils.openELF()` (the only part that cares about the correct kernel relocation offset) -- which is why this approach works at all in the first place. Agree it's unnecessarily fragile.

Two solutions come to mind, unfortunately both are somewhat intrusive:

1) Disallow the file search mode where the basename changes. This would affect scenario like b/168662041 - the user would need to ensure the filename matches what's in profile.
2) Or modify the Mapping struct to carry the "located filename" separately so that the original filename value is maintained as is and both first mapping override (`pprof /path/binary profile.pb.gz` mode) and the `$buildid/$name` lookup mode (and other modes actually) track the located path separately from the path stored in the profile.

I'll keep thinking of a better alternative.

s-kanev:
> I'll keep thinking of a better alternative.

How about a simpler twist on 2.? Instead of preserving the "located filename" in Mapping, we preserve the original one. We can even just store its baseName, which has semantic meaning -- it's the DSO. This way, for every usage of Mapping.File (there's quite a few, including internally), we don't have to decide if it needs "original" or "located" -- we keep the current behavior.

I sketched out how that might look like. PTAL.

aalexand:
> > I'll keep thinking of a better alternative.
> 
> How about a simpler twist on 2.? Instead of preserving the "located filename" in Mapping, we preserve the original one. We can even just store its baseName, which has semantic meaning -- it's the DSO. This way, for every usage of Mapping.File (there's quite a few, including internally), we don't have to decide if it needs "original" or "located" -- we keep the current behavior.
> 
> I sketched out how that might look like. PTAL.

What worries me a bit is that this appears to be the first time where we have a public field in profile.proto that is not a part of the persisted state. We would need to mention this in the comment to the field, as well as document semantics for operations like profile merge. Including handling outside pprof - e.g. see hasher.go in the internal profiler source code.

If we are going this route, I wonder if we should look more seriously into having something like "string type" field in the mapping proto - basically a way to disambiguate different mappings within the same file. We could define the semantics for that reasonably well (such as make sure the type is the part of the merge key) and that seems like a universally useful thing, even though there are upfront costs to pay like with any new field in profile.proto.

aalexand:
I think one scenario that is problematic with the current approach is that in some cases the original name (that guides the correctness of the kernel symbolization) can still be irreversibly lost. For example, I think running `pprof -output profile.pb.gz -symbolize=none input.pb.gz` where pprof is capable to find the kernel image and rename it to, say, `vmlinux` will lose the "text" vs "stext" distinction but won't symbolize the profile and the subsequent opening that will attempt to symbolize the profile won't be able to symbolize the kernel properly.

aalexand:
One possible idea (which is a hack but less invasive):

Earlier investigation says "When the perf tool generates an PERF_RECORD_MMAP, it can use either _text or _stext for the relocation symbol offset. Which one seems to just depend on the order of parsing the symbol table."

Could we apply similar logic in the local kernel symbolization and use _text or _stext depending on which one of the two is first in the symbol table?

I understand this is fragile but this is one of those hacks that if it works it can happily stay functional forever...

s-kanev:
> Could we apply similar logic in the local kernel symbolization and use _text or _stext depending on which one of the two is first in the symbol table?
> 
> I understand this is fragile but this is one of those hacks that if it works it can happily stay functional forever...

I'm a little hesitant, for obvious hackery reasons. It could be a thing that stays functional forever, but we'd need to replicate it in at least two places where we parse the symbol tables -- for local and remote symbolization -- here in pprof, and the internal remote symbol injestion pipeline. So the main benefit of being an isolated hack is somewhat decreased.

s-kanev:
Any way to make progress here?

I can take a stab at implementing the original "located filename" idea from [here](https://github.com/google/pprof/pull/675#issuecomment-991448362)?

That still has the downside of being a public non-persisted field in Mapping, and will be relatively invasive wrt to users of Mapping. But, at least, compared to the current proposal, it is idempotent, so something like `pprof -output profile.pb.gz -symbolize=none input.pb.gz` should continue to work.

s-kanev:
@aalexand PTAL.

For the record, we discussed offline in a lot more detail and extending this many public interfaces is no-one's favorite solution. But we ran out of other options to correctly symbolize kernels that can have either "_text" or "_stext" as the relocation symbol.

We decided to name the extra Mapping field narrowly (KernelRelocationOffset) to discourage more general use of this kludge.

internal/driver/fetch.go
```diff
@@ -420,12 +420,14 @@ mapping:
...
 			for _, name := range fileNames {
-				if f, err := obj.Open(name, m.Start, m.Limit, m.Offset); err == nil {
+				if f, err := obj.Open(name, m.Start, m.Limit, m.Offset, m.KernelRelocationSymbol); err == nil {
 					defer f.Close()
 					fileBuildID := f.BuildID()
 					if m.BuildID != "" && m.BuildID != fileBuildID {
 						ui.PrintErr("Ignoring local file " + name + ": build-id mismatch (" + m.BuildID + " != " + fileBuildID + ")")
 					} else {
+						// Explicitly do not update KernelRelocationSymbol --
+						// the relocated file name is most likely missing it.
```
aalexand, line 430: "relocation" vs. "relocated file name" is ambiguous here.

driver/driver.go
```diff
@@ -137,8 +137,10 @@ type MappingSources map[string][]struct {
 type ObjTool interface {
 	// Open opens the named object file. If the object is a shared
 	// library, start/limit/offset are the addresses where it is mapped
-	// into memory in the address space being inspected.
-	Open(file string, start, limit, offset uint64) (ObjFile, error)
+	// into memory in the address space being inspected. If the object
+	// is a linux kernel, relocationSymbol is the name of the symbol
+	// corresponding to the start address.
+	Open(file string, start, limit, offset uint64, relocationSymbol string) (ObjFile, error)
```
aalexand, line 143: Maybe `kernelRelocationSymbol` or (to shorten) `kernelRelocSym` or `kernelRelocSymbol`?

Trying to squeeze `kernel` in.

profile/encode.go
```diff
@@ -252,6 +253,13 @@ func (p *Profile) postDecode() error {
 		} else {
 			mappings[m.ID] = m
 		}
+
+		// If this a main linux kernel mapping with a relocation symbol suffix
```
aalexand, line 257: Maybe state explicitly that we understand it's a hack.

aalexand (dismissed):

internal/driver/fetch.go
s-kanev: Done, good catch.

profile/encode.go
s-kanev: Done.

driver/driver.go
s-kanev: I thought about that too, but the other parameters also only make sense for certain image types -- see "If the object is a shared library" comment above, so I'm inclined to leave as is.

Lmk if you feel strongly.

s-kanev:
FWIW, github requires re-approval once I pushed more commits (Is that a new feature? Seems counter-intuitive...)

driver/driver.go
aalexand: Sgtm.

aalexand (approved):

PR #676, opened 2021-12-04
Add badge link to Go API docs in pkg.go.dev

mhansen:
Thought this might be useful to point people at the API docs (after I tried searching for pprof API docs and couldn't find them easily on Google).

Points at https://pkg.go.dev/github.com/google/pprof/profile, as that's the only package we try to declare as an API.

Badge generated at https://pkg.go.dev/badge/?path=https%3A%2F%2Fpkg.go.dev%2Fgithub.com%2Fgoogle%2Fpprof%2Fprofile


aalexand:
> Perhaps this should point at https://pkg.go.dev/github.com/google/pprof/profile instead, WDYT?

Yes, I think so - the profile package is the only one we try to declare an "API".

mhansen:
Sounds good. Done. Funny story, I once tried writing my own go library to parse profiles because I didn't know this one existed! Hopefully this will help avoid that :-)

aalexand (approved):

PR #677, opened 2021-12-04
Update d3-flame-graph from 2.0.0-alpha to 4.1.3

mhansen:
This pulls in a fix for an XSS vulnerability, where profiles with HTML
tag frame labels would get executed by the browser:
https://github.com/spiermar/d3-flame-graph/pull/200

The flamegraph still looks the same, with the same theme.

d3-flame-graph's packaging method has changed between 2.0 and 4.1 (from
Rollup to Webpack). Rollup no longer works to build the bundle, so I've
migrated the bundling code to Webpack.

I've ~halved the file size, by:
- reducing the API surface of exports to just d3-flame-graph and
  d3-selection; these are the only functions used by pprof.
- combining d3 and d3-flame-graph into one bundle, to avoid duplicating
  common d3 code in both bundles.

Filesize before (separate):

```
$ du -sh d3/d3.go d3flamegraph/d3_flame_graph.go
136K    d3/d3.go
 24K    d3flamegraph/d3_flame_graph.go
```

After (combined):

```
$ du -h d3_flame_graph.go
72K    d3_flame_graph.go
```

I've tried to make the build more reproducible too, by checking in the
`package.json` and `package-lock.json` files that npm uses to freeze
dependency versions. Previously these files lived over at
https://github.com/spiermar/d3-pprof. I've bumped the version in
`package.json` to 2.0 as this seems like a large change to that package.

A few no-op changes:
- I removed an explicit setting of a transition animation in pprof's
  code (it's the default in d3-flame-graph, so it's redundant to set it,
  and increases the bundle complexity).
- I replaced a commented hack to set the `warm` colourscheme with the
  new API method that sets the colourscheme.

FYI @spiermar (d3-flame-graph maintainer).


aalexand:
@mhansen looks like some CI tests / checks failed.

mhansen:
Got tests passing locally. Rerunning tests on CI now.

aalexand (approved):

mhansen:
Thanks Alexey. FYI https://github.com/golang/go/pull/50156 will track
updating the pprof version in go downstream.

PR #683, opened 2022-02-17
internal/graph: Escape tag labels correctly for dot

evanj:
Fixes syntax errors when trying to render pprof profiles that have
double quotes in tags. These can be created with Go's pprof labels
feature, for example with:

    pprof.Labels("key", "label \"double quote\"\nline two")

Without this change, trying to render a diagram will fail with:

    Error: <stdin>: syntax error in line 5 near 'quote'

The problem is that the double quote ('"') was never escaped in the
label strings. This required me changing how multiple tags are
joined. Previously they were joined with the two character `\n`
sequence, which is a centered newline in dot. This changes it to
a single character newline "\n", which is then escaped as
left-justified strings "\l" in dot.

* Add a new graph test for labels.
* Add a test for joinLabels.
* Update the tests in driver which converts \n to \l.


internal/graph/graph.go
```diff
@@ -531,7 +532,8 @@ func joinLabels(s *profile.Sample) string {
 		}
 	}
 	sort.Strings(labels)
-	return strings.Join(labels, `\n`)
+	// join labels with a newline: this can be a bit confusing for labels with newlines
```
aalexand, line 535: nit: please make comments full sentences (titlecase, trailing dot).

Also here and in other places please make comments fit 80 columns, like the rest of the pprof codebase.

internal/driver/testdata/pprof.cpu.flat.functions.call_tree.dot
```diff
@@ -2,12 +2,12 @@ digraph "testbinary" {
 node [style=filled fillcolor="#f8f8f8"]
 subgraph cluster_L { "File: testbinary" [shape=box fontsize=16 label="File: testbinary\lType: cpu\lDuration: 10s, Total samples = 1.12s (11.20%)\lShowing nodes accounting for 1.11s, 99.11% of 1.12s total\lDropped 3 nodes (cum <= 0.06s)\l\lSee https://git.io/JfYMW for how to read the graph\l" tooltip="testbinary"] }
 N1 [label="line1000\n1s (89.29%)" id="node1" fontsize=24 shape=box tooltip="line1000 (1s)" color="#b20500" fillcolor="#edd6d5"]
-N1_0 [label = "key1:tag1\nkey2:tag1" id="N1_0" fontsize=8 shape=box3d tooltip="1s"]
+N1_0 [label = "key1:tag1\lkey2:tag1" id="N1_0" fontsize=8 shape=box3d tooltip="1s"]
```
aalexand, line 5: I think I would expect the current behavior to be preserved - centered tags.

I'd also expect any non-printable characters like newlines to be rendered as their C escaped version, so that the tag value is always one line when rendered. Same for function names. The escaping was initially added in https://github.com/google/pprof/pull/557 for graph legend labels that come from the profile comments field and for that we did support newlines to render newlines from comments as newlines since some comments intentionally use that kind of formatting. I think for function names and tag values it should be different and we shouldn't allow any formatting sequences in those strings.

internal/graph/graph.go
evanj: Okay! I've edited this to describe the escaping change you mentioned below.

evanj:
I do not like that `joinLabels` now escapes the string, but the output on my Go test program looks good to me now:
Screen Shot 2022-02-18 at 9 23 27 AM

evanj:
Hello! Is there anything else I should do to get this merged? I ran into another profile today that failed to render because of this bug. Thanks!

aalexand:
I tested the change a bit and I think it breaks how profile comments with newline characters in them are rendered - they are not displayed as newline characters anymore.

evanj:
Closing in favour of #689

PR #684, opened 2022-02-18
allow rendering big flame graphs by avoiding stack overflow in JS parser

### LINKED ISSUE #663: Sometimes very large flamegraphs don't render

Git Commit: `20978b51388db0648809a2c5cc88b494c7945ec1` (https://github.com/google/pprof/commit/20978b51388db0648809a2c5cc88b494c7945ec1)

We are generating data for PProf through the julia package https://github.com/JuliaPerf/PProf.jl

MacOS Intel x86

### What did you do?

Very large profile data sometimes does not render in the flamegraphs view. It will render okay in the Graph view, but not in the Flamegraph view.

Screen Shot 2021-11-02 at 2 45 53 PM

Screen Shot 2021-11-02 at 2 45 33 PM

I expected to see the standard flamegraph view.

### What did you see instead?

Nothing 😢 

-----

I'm sorry that I don't have a .pb.gz to upload, because the profile i currently have is from our internal, closed source codebase.

I'll try to find a reproducer profile that I can upload here with the issue. In the meantime, maybe this report is already useful.

Thanks! :)

vilterp:
Fixes #663 

## Before

Very large flame graphs fail to render, with a mysterious error in the console: `Maximum call stack size exceeded`, without a stack trace.

e.g. [big-profile.pb.gz](https://github.com/google/pprof/files/8094230/big-profile.pb.gz):

```
pprof -http localhost:9876 ~/Downloads/big-profile.pb.gz
```

image

I suspected that the very large array in `var data = …` might be a problem, so I inserted `console.log`s before and after it. To my surprise, neither log showed up. This, combined with the missing stack trace, indicated to me that the stack overflow was not in javascript execution, but in parsing the javascript itself.

## After

The same flame graph now renders.

image

The fix is to use the browser's JSON parser (via `JSON.parse`) instead of its JS parser. Due to the JSON grammar being simpler, the browser can handle larger strings.

Presumably JSON parsing has a limit too, past which you'd have to chunk the string or something; but this should be good enough for profiles that are quite big.

## Drawbacks

Go's `template.JS` seems to escape the string in a very verbose way. Before, it came out as

```js
var data = {"n":"root", ...}
```

Now it's:

```js
var data = JSON.parse("{\u0022n\u0022:\u0022root\u0022…")
```

It works, but it kind of blows up the size. Not sure what an idiomatic Go way of fixing this is.

cc @NHDaly


NHDaly (approved):
Amazing! :) So simple 🤯 

Thanks Pete, this is awesome. The explanation makes sense to me. Thanks for the fix. Yeah, i'm not fluent in go myself, so i can't say what's idiomatic. But this seems like a nice change 👍

NHDaly:
Hi again, @aalexand. If you're still managing this repository, could you please have a look at this PR when you get a moment? We are currently unable to open many of the profiles we're recording, such as the one uploaded in the PR description, and it would be great to have this fixed!!

Please let us know if there's anything we can do to help with this review!
Best, -Nathan

internal/driver/webhtml.go
```diff
@@ -1329,7 +1329,7 @@ function viewer(baseUrl, nodes) {
   <script>viewer(new URL(window.location.href), {{.Nodes}});</script>
   <script>{{template "d3flamegraphscript" .}}</script>
   <script>
-    var data = {{.FlameGraph}};
+    var data = JSON.parse("{{.FlameGraph}}");
```
NHDaly, line 1332: Maybe you could add a comment indicating why we're calling JSON.parse here? Basically a short summary of what you have in the PR describe?

aalexand: +1

aalexand:
I played a bit with the escaping here - https://go.dev/play/p/X0_CvkZedlo - but couldn't find an easy way to avoid this \u sequences.

internal/driver/webhtml.go
NHDaly: Thanks for adding the comment, @vilterp 👍

NHDaly:
Thanks for looking into it, @aalexand!! From what I read here, it seems like the escaping is maybe the expected behavior?: https://github.com/golang/go/issues/42506

It's a bit too bad, but probably fine to leave it. Like @vilterp said, hopefully we should have a lot more room before hitting a stack overflow with the JSON parser, so the extra size should be okay here.. 🤞

NHDaly (approved):
LGTM after adding the comment. thanks

NHDaly:
Thanks @aalexand. Is this waiting on anything else? it looks like maybe it's blocked on CI failures?

aalexand:
> Thanks @aalexand. Is this waiting on anything else? it looks like maybe it's blocked on CI failures?

Updated the branch. #692 should fix the CI failures.

NHDaly:
amazing, thank you! Indeed that did it. :)

aalexand (approved):

NHDaly:
🎉 thanks @aalexand! :)

vilterp:
Seconded; thank you!

PR #690, opened 2022-03-24
Support parsing jemalloc profile

dqminh:
This is related to https://github.com/jemalloc/jemalloc/issues/2224

We have several applications that use jemalloc, which exposes slightly different text-based heap profile compared to legacy tcmalloc format. As pprof has a lot more features than jeprof which is shipped with with jemalloc, i wonder if pprof would like to take a patch that includes support for parsing jemalloc heap profile.

This adds priliminary support for parsing jemalloc heap profile. The heap profile is in fomrmat:

  heap_v2/SAMPLE
    t*: <count1>: <bytes1> [<count2>: <bytes2>]
    t1: <count1>: <bytes1> [<count2>: <bytes2>]
      ...
    tn: <count1>: <bytes1> [<count2>: <bytes2>]
  @ a1 a2 ... an
    t*: <count1>: <bytes1> [<count2>: <bytes2>]
    t1: <count1>: <bytes1> [<count2>: <bytes2>]
      ...
    tn: <count1>: <bytes1> [<count2>: <bytes2>]

  MAPPED_LIBRARIES:
    ...

As jemaloc heap profile is thread-based, we record each thread sample individually, with label thread=TID


aalexand:
@dqminh Is there a way to make jemalloc produce profiles in the profile.proto format instead, like tcmalloc does? See [this commit](https://github.com/google/tcmalloc/commit/2bed1bd64c83595548dc9bb1b4bdd7e0ed116540). The legacy formats are, well, legacy, and I don't feel enthusiastic about adding more parsing code to maintain.

dqminh:
maybe @lapenko can discuss probability of adding profile.proto format for jemalloc.

One alternative that i can think of in addition to current approach is adding this PR to https://github.com/felixge/pprofutils instead. It should work but require one more step to convert jemalloc profile to profile.proto, which is not ideal but does not require changes in both jemalloc and pprof.

PR #691, opened 2022-03-31
Fix tagroot to properly format unitless numeric tags.

### LINKED ISSUE #651: Profile.Copy throws away all NumLabels after it hits NumUnit==""

`02619b876842e0d0afb5e5580d3a374dad740edb`

```
$ uname -a
Linux markhansen.syd.corp.google.com 5.10.40-1rodete2-amd64 #1 SMP Debian 5.10.40-1rodete2 (2021-06-22) x86_64 GNU/Linux
```

### What did you do?
See https://github.com/google/pprof/pull/649/commits/4c12914e03def8579623408e741bb6f5879e0f92#diff-bbbb7a39a412ce191a6c45df8a9ccaba7ac1ca16b71b80c4d9ae4a14fa075a15, in particular this sample, with a numeric label `allocations` with a number value `1024` and no units (`""`):

```
		{
			Location: []*profile.Location{cpuL[2], cpuL[0]},
			Value:    []int64{10, 10},
			Label: map[string][]string{
				"key1": {"tag3"},
				"key2": {"tag2"},
			},
			NumLabel: map[string][]int64{
				"allocations": []int64{1024},
			},
			NumUnit: map[string][]string{
				"allocations": []string{""},
			},
		},
```

This should be printed as `allocations:1024 foo foo_caller main: 10`, but instead, (see the passing test), is printed without the numbers:

```
				// TODO: bug, this line should say allocations:1024
				"allocations: foo foo_caller main: 10",
```

The blame seems to be somewhere in `Profile.preEncode` or `Profile.postDecode` (called when I call `Profile.Copy`), which seems to skip any numeric values after it hits an empty string unit.

aalexand:
It's possible for the unit array to be empty for a unitless numeric tag.
In this case the tag value should be formatted properly but before this
change it was output empty.

Fixes #651. From the bug description it doesn't look like exactly this
issue but it does get fixed.

The added tests fail before the change and pass after.

### DIFF
--- a/internal/driver/tagroot.go
+++ b/internal/driver/tagroot.go
@@ -113,12 +113,17 @@ func formatLabelValues(s *profile.Sample, k string, outputUnit string) []string 
 	values = append(values, s.Label[k]...)
 	numLabels := s.NumLabel[k]
 	numUnits := s.NumUnit[k]
-	if len(numLabels) != len(numUnits) {
+	if len(numLabels) != len(numUnits) && len(numUnits) != 0 {
 		return values
 	}
 	for i, numLabel := range numLabels {
-		unit := numUnits[i]
-		values = append(values, measurement.ScaledLabel(numLabel, unit, outputUnit))
+		var value string
+		if len(numUnits) != 0 {
+			value = measurement.ScaledLabel(numLabel, numUnits[i], outputUnit)
+		} else {
+			value = measurement.ScaledLabel(numLabel, "", "")
+		}
+		values = append(values, value)
 	}
 	return values
 }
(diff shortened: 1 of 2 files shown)


internal/driver/tagroot.go
```diff
@@ -113,12 +113,17 @@ func formatLabelValues(s *profile.Sample, k string, outputUnit string) []string
...
 	numUnits := s.NumUnit[k]
-	if len(numLabels) != len(numUnits) {
+	if len(numLabels) != len(numUnits) && len(numUnits) != 0 {
 		return values
 	}
 	for i, numLabel := range numLabels {
-		unit := numUnits[i]
-		values = append(values, measurement.ScaledLabel(numLabel, unit, outputUnit))
+		var value string
+		if len(numUnits) != 0 {
```
gmarin13, line 121: Should this be:
if len(numUnits) > i ?

aalexand: I think this is not necessary because there is `len(numLabels) != len(numUnits)` check above.

gmarin13: I see. I missed that part.

gmarin13 (approved):

PR #703, opened 2022-05-17
Made flamegraph test less brittle.

ghemawat:
Different environments may have compiled flamegraph JS code
differently. Generalize regular expression that checks
generated code to allow some variation.


nolanmar511 (approved):

nolanmar511:
@ghemawat - Sorry for comment after reviewing and merging. 

I'm realizing this change allows the test to pass in cases in which the version of d3-flame-graph used by the CLI does not expose the correct methods to allow the flame graph to render. Is this the desired behavior for this test?

Issue #710, opened 2022-07-11
internal/driver/fetch: searched files list looks weird


egonelbre:
Based on c488b8fa1db3fa467bf30beb5a1d6f4f10bb1b87.

I'm currently working through and trying to get ASLR disassembly support working in Go + Windows. One of the weird things I noticed is the line https://github.com/google/pprof/blob/c488b8fa1db3fa467bf30beb5a1d6f4f10bb1b87/internal/driver/fetch.go#L420.

The `fileNames = append(fileNames, filepath.Join(path, m.File))` doesn't make sense when `m.File` is assumed to be absolute. Is that a typo and should it be `fileNames = append(fileNames, m.File)`? The comment does seems to indicate the latter. Or does it have some other purpose (e.g. compatibility in case the `m.File` is not absolute).

aalexand:
There is a comment there:

```
// Try both the basename and the full path, to support the same directory
// structure as the perf symfs option.
```

egonelbre:
@aalexand I'm not quite following. If `m.File = "C:\example\test.exe"` then the `filepath.Join(path, m.File) == "C:\UserData\egon\pprof\binaries\C:\example\test.exe"`, which is not a valid filepath.

aalexand:
@egonelbre Ah, right - the volume name needs to be trimmed from m.File before appending it.

PR #711, opened 2022-07-11
internal/driver: fixes for fetch.go

egonelbre:
url.Parse can parse a Windows absolute path without problems.
This can cause mapping file to be cleared, even though it's not a
remote source.

For #710

### DIFF
--- a/internal/driver/fetch.go
+++ b/internal/driver/fetch.go
@@ -408,13 +408,13 @@ func locateBinaries(p *profile.Profile, s *source, obj plugin.ObjTool, ui plugin
 	}
 mapping:
 	for _, m := range p.Mapping {
-		cleanFile := strings.TrimPrefix(m.File, filepath.VolumeName(m.File))
-
+		var noVolumeFile string
 		var baseName string
 		var dirName string
 		if m.File != "" {
+			noVolumeFile = strings.TrimPrefix(m.File, filepath.VolumeName(m.File))
 			baseName = filepath.Base(m.File)
-			dirName = filepath.Dir(cleanFile)
+			dirName = filepath.Dir(noVolumeFile)
 		}
 
 		for _, path := range filepath.SplitList(searchPath) {
@@ -424,7 +424,7 @@ mapping:
 				if matches, err := filepath.Glob(filepath.Join(path, m.BuildID, "*")); err == nil {
 					fileNames = append(fileNames, matches...)
 				}
-				fileNames = append(fileNames, filepath.Join(path, cleanFile, m.BuildID)) // perf path format
+				fileNames = append(fileNames, filepath.Join(path, noVolumeFile, m.BuildID)) // perf path format
 				// Llvm buildid protocol: the first two characters of the build id
 				// are used as directory, and the remaining part is in the filename.
 				// e.g. `/ab/cdef0123456.debug`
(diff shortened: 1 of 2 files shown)


internal/driver/fetch.go
```diff
@@ -378,14 +379,31 @@ func collectMappingSources(p *profile.Profile, source string) plugin.MappingSour
...
+// isWindowsAbsPath detects whether path starts with drive letter such as "C:\".
+func isWindowsAbsPath(path string) bool {
+	if len(path) < 3 {
+		return false
+	}
+	// This catches most Windows absolute paths.
+	return unicode.IsLetter(rune(path[0])) && path[1] == ':' && (path[2] == '/' || path[2] == '\\')
+}
+
+// trimWindowsVolume removes windows volume name from path, if it exists.
```
aalexand, line 399: I am not sure the function is needed, you can just do `strings.TrimPrefix(filepath.VolumeName(path))`.

internal/driver/fetch.go
```diff
@@ -378,14 +379,31 @@ func collectMappingSources(p *profile.Profile, source string) plugin.MappingSour
...
-		if m.BuildID == "" {
+		if m.BuildID == "" && !isWindowsAbsPath(m.File) {
 			if u, err := url.Parse(m.File); err == nil && u.IsAbs() {
 				m.File = ""
 			}
 		}
 	}
 }
 
+// isWindowsAbsPath detects whether path starts with drive letter such as "C:\".
```
aalexand, line 390: Can we just do `if m.BuildID && filepath.VolumeName(m.File) == ""`?

egonelbre: Gentle ping.

aalexand: Following up on this, apologies for the delay. I am hesitant to claim pprof support of opening Windows-collected profiles on Linux. This seems exotic. Hence, I would suggest we use `filepath.VolumeName` here. Similar below in the trimming code.

egonelbre: It does happen quite often for me, but very well, I'll use filepath.VolumeName.

internal/driver/fetch.go
```diff
@@ -408,11 +408,13 @@ func locateBinaries(p *profile.Profile, s *source, obj plugin.ObjTool, ui plugin
 	}
 mapping:
 	for _, m := range p.Mapping {
+		cleanFile := strings.TrimPrefix(m.File, filepath.VolumeName(m.File))
```
aalexand, line 411: nit: I'd rename to `noVolumeFile` for clarity. "Clean" can mean many things.

Another nit: perhaps declare the var and handle its assignment in the same way as baseName and dirName which is under the condition? No need to handle this one differently.

internal/driver/fetch_test.go
```diff
@@ -126,23 +126,31 @@ func TestCollectMappingSources(t *testing.T) {
...
+		{"", "http://example.com", "", ""},
+		{"windows", `C:\example.exe`, "", `C:\example.exe`},
+		{"windows", `c:/example.exe`, "", `c:/example.exe`},
 	} {
-		p := &profile.Profile{
-			Mapping: []*profile.Mapping{
-				{
-					File:    tc.file,
-					BuildID: tc.buildID,
+		t.Run("", func(t *testing.T) {
```
aalexand, line 136: Empty name for t.Run() is unusual. I'd pass tc.file or tc.file+"-"+tc.os (and use "any" instead of empty string to mean any OS).

egonelbre: Ah, sorry, accidentally forgot to fill the name

PR #716, opened 2022-07-27
Added alternative flamegraph implementation that can show callers.

ghemawat:
Add an experimental flame-graph implementation. It can be selected in
pprof's web interface using the new "Flame (experimental)" menu entry.
At some point this new implementation may become the default.

The new view is similar to flame-graph view. But it can show caller
information as well. This should allow it to satisfy many users of
Graph and Peek views as well.

Let's illustrate with an example. Suppose we have profile data that
consists of the following stacks:

```
1000	main -> foo -> malloc
2000	main -> bar -> malloc
```

When main is selected, both the old and new views show:

```
[-------------------3000 main---------------------]
[---1000 foo-----] [----------2000 bar------------]
[---1000 malloc--] [----------2000 malloc---------]
```

But suppose now the user selects the right-most malloc slot.
The old view will show just the path leading to that malloc:

```
[----------2000 main-----------]
[----------2000 bar------------]
[----------2000 malloc---------]
```

The new view will however show a flame-graph view that grows
upwards that displays the call stacks leading to malloc:

```
[---1000 main----] [----------2000 main-----------]
[---1000 foo-----] [----------2000 bar------------]
[-------------------3000 malloc-------------------]
```

This caller display is useful when trying to determine expensive
callers of function.

A list of important differences between the new view and flame graphs:

New view pros:

1.  Callers are shown, e.g., all paths leading to malloc.
2.  Shows self-cost clearly with a different saturation.
3.  Font size is adjusted to fit more text into boxes.
4.  Highlighting on hover shows other occurrences of a function.
5.  Search works more like other views.
6.  Pivot changes are reflected in browser history (so back and forward
    buttons can be used to navigate between different selections).
7.  Allows eventual removal of the D3 dependency, which may make
    integrations into various environments easier.
8.  Colors provide higher contrast between foreground and background.

New view cons:

1.  There are small differences in how things look and feel.
2.  Color-scheme is very different.
3.  Change triggered by selecting a new entry is not animated.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -399,6 +399,40 @@ the user to interactively view profile data in multiple formats.
 
 The top of the display is a header that contains some buttons and menus.
 
+## View
+
+The `View` menu allows the user to switch between different visualizations of
+the profile.
+
+Top
+:   Displays a tabular view of profile entries. The table can be sorted
+    interactively.
+
+Graph
+:   Displays a scrollable/zoomable graph view; each function (or profile entry)
+    is represented by a node and edges connect callers to callees.
+
+Flame Graph
+:   Displays a [flame graph](https://www.brendangregg.com/flamegraphs.html).
+
+Flame (experimental)
+:   Displays a view similar to a flame graph that can show the selected node's
+    callers and callees simultaneously.
+
+NOTE: This view is currently experimental and may eventually replace the normal
+Flame Graph view.
+
+Peek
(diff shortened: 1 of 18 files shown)


amchiclet (approved):

asad-awadia:
Can't even fathom how it feels to review a PR from `ghemawat` or jdean`

PR #725, opened 2022-09-23
proposal: add timestamp breakdown field for samples

felixge:
TODO: Update pprof encoding/decoding code, se https://github.com/felixge/pprof/pull/1/files.

### DIFF
--- a/proto/profile.proto
+++ b/proto/profile.proto
@@ -129,7 +129,7 @@ message Breakdown {
   // profile start, this list should hold the values [3, 1, 2].
   repeated int64 time_offset_nanos = 1;
   // A list that is either empty or holding exactly the same number of elements
-  // as Sample.time_offset_nanos, associating each point in time with a
+  // as Breakdown.time_offset_nanos, associating each point in time with a
   // recorded value. For cumulative sample types the sum of these values should
   // add up to the corresponding Sample.value.
   repeated int64 value = 2;
--- a/proto/profile.proto
+++ b/proto/profile.proto
@@ -125,7 +125,7 @@ message Sample {
 message Breakdown {
   // A list of timestamps at which the values of the associated sample were
   // collected. The time is given relative to Profile.time_nanos and values are
-  // delta encoded. So for two events captured at 3, 4 and 6 nsec after the
+  // delta encoded. So for three events captured at 3, 4 and 6 nsec after the
   // profile start, this list should hold the values [3, 1, 2].
   repeated int64 time_offset_nanos = 1;
   // A list that is either empty or holding exactly the same number of elements
--- a/proto/profile.proto
+++ b/proto/profile.proto
@@ -75,7 +75,7 @@ message Profile {
   // The following fields are informational, do not affect
   // interpretation of results.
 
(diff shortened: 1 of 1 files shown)


proto/profile.proto
```diff
@@ -115,6 +115,24 @@ message Sample {
   // label includes additional context for this sample. It can include
   // things like a thread id, allocation size, etc
   repeated Label label = 3;
+  // Breakdown can hold the individual timestamps and values for a sample that
+  // has aggregated values. The type and unit of each breakdown is defined by
+  // the corresponding entry in Profile.sample_type. It's acceptable to have
+  // less, but not more, breakdown entries than Profile.sample_type entries.
+  repeated Breakdown breakdown = 4;
```
felixge, line 122: Christophe had concerns about the size implication of having one `Breakdown` message per sample type for the case where the data is sparse, i.e. there is no breakdown information available for some of the sample types. My thinking is this:

In the worst case each empty breakdown message will take up 2 bytes in size (1 byte for the field number and type info, and one byte for the length of the message which is `0`).

This might not seem ideal, but I expect that a typical Sample will be at least ~50 bytes (we could analyze some data to confirm), so the having an empty breakdown is only a 4% (2/50) increase in size.

Typically I'd expect that it's always the same sample types that have empty breakdowns (because they don't support the feature), so one could eliminate that cost completely by ordering the sample types without breakdown last and only having breakdowns for the first few sample types that support it.

felixge:
Ignore, not ready yet.

aalexand:
FYI - for a different reason I was discussing with a colleague adding the timestamp unit support for pprof labels in some way. We have the unit support in general, I was thinking of standardizing on something like "epoch-microseconds" or similar and adding support for rendering this unit in the code. The motivation there is related to the contention profiles we gather internally where we we want to store an exemplar timestamp for the peak / longest contention event observed for a given stack. So it's not exactly what people have in mind when they ask for timestamp support in profile.proto but I thought I'd mention this.

aalexand:
> The main advantage for this is that ticks_per_second allows the profiler to chose the resolution of the timestamps (e.g. 1e9 = nsec, 1e6 = usec, 1e3 = msec, or something like 2.6e9 for TSC) which has a significant impact on file size. This is also what JFR does.

The current profile.proto approach to this is to allow expressing the magnitude in the unit. See [here](https://github.com/google/pprof/blob/70bd9ae97f40a5e0ef03e61bec8e90f3a5732191/internal/measurement/measurement.go#L272). I'd prefer to reuse this approach.

felixge:
> The current profile.proto approach to this is to allow expressing the magnitude in the unit. See [here](https://github.com/google/pprof/blob/70bd9ae97f40a5e0ef03e61bec8e90f3a5732191/internal/measurement/measurement.go#L272). I'd prefer to reuse this approach.

Thanks for pointing this out. I wasn't aware that the units had been formalized in `measurement.go`. I'm okay with that approach. The only downside I see is that it would make the TSC use case more difficult to support. But I personally don't care for it too much, the only thing I feel strongly about is letting the profilers chose the resolution.

aalexand:
True, but I don't want to think about TSC support yet :)

PR #727, opened 2022-10-10
Increase chunk size for concurrent fetch to 128

craig-chasseur:
This makes it more convenient to profile jobs with many replicas. Per offline discussion with aalexand, we prefer to just raise the limit rather than adding another config knob.


aalexand (approved):

craig-chasseur:
Single failed check on mac appears totally unrelated to PR:

```
Run go get -d .
# cd /Users/runner/work/pprof/pprof/src/github.com/ianlancetaylor; git clone -- https://github.com/ianlancetaylor/demangle /Users/runner/work/pprof/pprof/src/github.com/ianlancetaylor/demangle
Cloning into '/Users/runner/work/pprof/pprof/src/github.com/ianlancetaylor/demangle'...
error: inflate: data stream error (incorrect data check)
fatal: pack has bad object at offset 2066: inflate returned -3
fatal: fetch-pack: invalid index-pack output
package github.com/ianlancetaylor/demangle: exit status [12](https://github.com/google/pprof/actions/runs/3221673892/jobs/5269932252#step:8:13)8
Error: Process completed with exit code 1.
```

Maybe a flake. Can we re-run?

PR #731, opened 2022-11-02
Start using golangci-lint instead of the deprecated x/lint.

aalexand:
See https://github.com/golang/go/issues/38968 for the x/lint deprecation message.

I'm also putting all tool invocations into test.sh so that it's easier to run all checks locally during the development time. GitHub Actions are nice but they make running the checks locally more difficult.


profile/legacy_profile.go
```diff
@@ -865,7 +865,6 @@ func parseThread(b []byte) (*Profile, error) {
 	// Recognize each thread and populate profile samples.
 	for !isMemoryMapSentinel(line) {
 		if strings.HasPrefix(line, "---- no stack trace for") {
-			line = ""
```
wyk9787, line 868: Is this intentional?

test.sh
```diff
@@ -40,3 +40,7 @@ for d in $PKG; do
   fi
 done
 
+go vet -all ./...
+golangci-lint run -D errcheck ./...  # TODO: Enable errcheck back.
```
wyk9787, line 44: I assume we have places in the pprof CLI that we don't check errors so we need to disable this in the linter? Are you  going to fix it separately?

wyk9787 (changes requested):

test.sh
aalexand: Yes, the linter complained about a few places there. I don't plan to work on this soon as it's not a regression.

profile/legacy_profile.go
aalexand: Yes, it fixed a lint warning that this assignment is redundant - we break out of the loop and this var is not used after the loop.

wyk9787: Ah, okay. Thanks.

wyk9787 (approved):

ghemawat (commented):
LGTM

Labutin:
golangci-lint run -D errcheck ./... fails for my pull request https://github.com/google/pprof/pull/734
See an example: https://github.com/google/pprof/actions/runs/3422932933/jobs/5700946312
Is it expected?

aalexand:
Not expected. Looks like it fails at Go tip - perhaps Go mainline added something that the linter doesn't like.

PR #734, opened 2022-11-08
Compatibilize profiles before merging

Labutin:
pprof merging depends on the number of sample type and their order. That requires additional preparation steps before merging profiles that are not entirely identical in terms of sample type sets. The PR makes that process easier.

CompatibilizeSampleTypes is called to allow merging profiles with different sets of sample types. The resulting profile has sample types that appear in all profiles only.


aalexand (commented):
It would be good to explain a bit more in the PR description why the change is made.

Labutin:
The description is updated.

aalexand (approved):

PR #735, opened 2022-11-09
Disable golangci-lint on tip

Labutin:
golangci-lint fails on tip.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -41,6 +41,8 @@ for d in $PKG; do
 done
 
 go vet -all ./...
-golangci-lint run -D errcheck ./...  # TODO: Enable errcheck back.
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -41,6 +41,8 @@ for d in $PKG; do
 done
 
 go vet -all ./...
-golangci-lint run -D errcheck ./...  # TODO: Enable errcheck back.
+if [ `go version | { read _ _ v _; echo ${v#go}; }` != "devel" ];  then
```
aalexand, line 44: Could we rather do this similar to how we handle the staticcheck? We also skip it on tip, via RUN_STATICCHECK env var. See above and ci.yaml (note that it's set in two places there).

Labutin: Done.

aalexand (approved):

PR #741, opened 2022-12-09
Extend search for debug symbols

adriaan42:
Look for debug symbols in all locations GDB does (https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html)
, e.g.:

- /usr/lib/debug/.build-id/ab/cdef1234.debug
- /usr/bin/binary.debug
- /usr/bin/.debug/binary.debug
- /usr/lib/debug/usr/bin/binary.debug.

My use case is development for embedded devices using Yocto. There I get a rootfs which I boot on target, and a rootfs-dbg with all debug symbols in `/usr/lib/debug/path/to/file.debug`. I record data with tcmalloc on target, and run pprof on that data after copying it to my development host. This change (specifically option 4 from the list below) lets me point `PPROF_BINARY_PATH=/path/to/rootfs-dbg/`, and find debug symbols for my application and all shared libraries it uses.

Option 1 from the list above is already implemented, option 4 is the one I would like to support, and I added 2 and 3 to the patch for completeness.

Signed-off-by: Adriaan Schmidt <adriaan.schmidt@siemens.com>

### DIFF
--- a/internal/driver/cli.go
+++ b/internal/driver/cli.go
@@ -364,5 +364,7 @@ var usageMsgVars = "\n\n" +
 	"   PPROF_BINARY_PATH  Search path for local binary files\n" +
 	"                      default: $HOME/pprof/binaries\n" +
 	"                      searches $buildid/$name, $buildid/*, $path/$buildid,\n" +
-	"                      ${buildid:0:2}/${buildid:2}.debug, $name, $path\n" +
+	"                      ${buildid:0:2}/${buildid:2}.debug, $name, $path,\n" +
+	"                      ${name}.debug, $dir/.debug/${name}.debug,\n" +
+	"                      usr/lib/debug/$dir/${name}.debug\n" +
 	"   * On Windows, %USERPROFILE% is used instead of $HOME"
--- a/internal/driver/fetch.go
+++ b/internal/driver/fetch.go
@@ -409,8 +409,10 @@ func locateBinaries(p *profile.Profile, s *source, obj plugin.ObjTool, ui plugin
 mapping:
 	for _, m := range p.Mapping {
 		var baseName string
+		var dirName string
 		if m.File != "" {
 			baseName = filepath.Base(m.File)
+			dirName = filepath.Dir(m.File)
 		}
 
 		for _, path := range filepath.SplitList(searchPath) {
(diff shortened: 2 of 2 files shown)


adriaan42:
ping...
@aalexand do you need any more information/input from me?

aalexand:
@adriaan42 Could you add in the description a bit of details on what use case / scenario puts debug information files into each of these locations?

A note on how the change was tested end-to-end would also be good.

I'm trying to make sure we are adding a fix for a real problem, not simply align the search path with GDB in abstract.

adriaan42:
> Could you add in the description a bit of details on what use case / scenario puts debug information files into each of these locations?

@aalexand Sure! I've updated the description with details on my use case.

> A note on how the change was tested end-to-end would also be good.

I've tested manually to confirm that indeed all debug symbols for the application as run on my embedded target are found during analysis on my development machine.

internal/driver/fetch.go
```diff
@@ -429,10 +431,12 @@ mapping:
 			if m.File != "" {
 				// Try both the basename and the full path, to support the same directory
 				// structure as the perf symfs option.
-				if baseName != "" {
-					fileNames = append(fileNames, filepath.Join(path, baseName))
-				}
+				fileNames = append(fileNames, filepath.Join(path, baseName))
 				fileNames = append(fileNames, filepath.Join(path, m.File))
+				// Try other common locations
```
aalexand, line 436: "other common locations" is not very specific. Could you copy some words from the PR description to here to make this a bit cleerer? https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html is the most important bit.

aalexand:
[Docs in cli.go](https://github.com/google/pprof/blob/76d1ae5aea2b3f738f2058d17533b747a1a5cd01/internal/driver/cli.go#L366) need to be updated.

internal/driver/fetch.go
adriaan42: I have extended the comment (included the link), and added the Yocto example to the commit message.

internal/driver/cli.go
```diff
@@ -364,5 +364,7 @@ var usageMsgVars = "\n\n" +
 	"   PPROF_BINARY_PATH  Search path for local binary files\n" +
 	"                      default: $HOME/pprof/binaries\n" +
 	"                      searches $buildid/$name, $buildid/*, $path/$buildid,\n" +
-	"                      ${buildid:0:2}/${buildid:2}.debug, $name, $path\n" +
+	"                      ${buildid:0:2}/${buildid:2}.debug, $name, $path,\n" +
+	"                      ${name}.debug, $path/.debug/${name}.debug,\n" +
```
aalexand, line 368: $path should be $dir here I think. Because two lines above $path is used for full path.

Same below.

aalexand (approved):

PR #743, opened 2022-12-16
Modify the name shortening heuristic to handle non-Go names

aalexand:
Names like foo/xyz are not Go names but match the current regular expression. Modify it to require a dot to be present for the Go shortening to trigger.

This should fix b/262922429.

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -33,7 +33,7 @@ var (
 	javaRegExp = regexp.MustCompile(`^(?:[a-z]\w*\.)*([A-Z][\w\$]*\.(?:<init>|[a-z][\w\$]*(?:\$\d+)?))(?:(?:\()|$)`)
 	// Removes package name and method arguments for Go function names.
 	// See tests for examples.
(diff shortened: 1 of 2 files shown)


internal/graph/graph.go
```diff
@@ -33,7 +33,7 @@ var (
 	javaRegExp = regexp.MustCompile(`^(?:[a-z]\w*\.)*([A-Z][\w\$]*\.(?:<init>|[a-z][\w\$]*(?:\$\d+)?))(?:(?:\()|$)`)
 	// Removes package name and method arguments for Go function names.
 	// See tests for examples.
-	goRegExp = regexp.MustCompile(`^(?:[\w\-\.]+\/)+(.+)`)
+	goRegExp = regexp.MustCompile(`^(?:[\w\-\.]+\/)+([^.]+\..+)`)
```
gmarin13, line 36: The only concern is if these non-Go names can also have periods in them.
But it is an heuristic, and it seems strictly better than the old one.

PR #747, opened 2023-01-14
Update pprof's legacy profile code to handle tagged anon VMAs.

blakej11:
This is an internal change needed to support legacy profiles.


s-kanev (approved):

aalexand:
We decided not to make this change.

PR #749, opened 2023-01-31
NumLabel helper functions for Profile.

blakej11:
Profile has some helper functions for manipulating the string-valued Labels associated with a Sample. These are some corresponding routines for manipulating the corresponding integer-valued Labels.

### DIFF
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -715,6 +715,36 @@ func (s *Sample) HasLabel(key, value string) bool {
 	return false
 }
 
+// SetNumLabel sets the specified key to the specified value for all samples in the
+// profile.
+func (p *Profile) SetNumLabel(key string, value []int64) {
+	for _, sample := range p.Sample {
+		if sample.NumLabel == nil {
+			sample.NumLabel = map[string][]int64{key: value}
+		} else {
+			sample.NumLabel[key] = value
+		}
+	}
+}
+
+// RemoveNumLabel removes all labels associated with the specified key for all
+// samples in the profile.
+func (p *Profile) RemoveNumLabel(key string) {
+	for _, sample := range p.Sample {
+		delete(sample.NumLabel, key)
+	}
+}
+
+// HasNumLabel returns true if a sample has a label with indicated key and value.
+func (s *Sample) HasNumLabel(key string, value int64) bool {
+	for _, v := range s.NumLabel[key] {
+		if v == value {
+			return true
+		}
+	}
+	return false
+}
+
 // DiffBaseSample returns true if a sample belongs to the diff base and false
 // otherwise.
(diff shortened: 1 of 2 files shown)


profile/profile.go
```diff
@@ -715,6 +715,36 @@ func (s *Sample) HasLabel(key, value string) bool {
 	return false
 }
 
+// SetNumLabel sets the specified key to the specified value for all samples in the
+// profile.
+func (p *Profile) SetNumLabel(key string, value []int64) {
```
aalexand, line 720: I think this helper should take the unit parameter as in vast majority of cases numeric labels need to specify the unit. Besides the general advantage of units giving the numeric labels physical meaning, handling of zero values for numeric labels subtly depends on the presence of a unit - https://github.com/google/pprof/pull/619.

profile/profile.go
```diff
@@ -715,6 +715,36 @@ func (s *Sample) HasLabel(key, value string) bool {
...
+// RemoveNumLabel removes all labels associated with the specified key for all
+// samples in the profile.
+func (p *Profile) RemoveNumLabel(key string) {
+	for _, sample := range p.Sample {
+		delete(sample.NumLabel, key)
+	}
+}
+
+// HasNumLabel returns true if a sample has a label with indicated key and value.
+func (s *Sample) HasNumLabel(key string, value int64) bool {
```
aalexand, line 739: I wonder if this helper is added for symmetry or there is a use case for it. I can see that for string labels checking the presence of a specific value of a specific label can be useful since string labels are basically enums and enum values are pets. For numeric labels values are more of cattle IMO and checking for a specific value seems exotic.

profile/profile.go
blakej11: Ah yeah, that makes sense! Fixed.

profile/profile.go
blakej11: I don't have a specific use case for it, and I'm fine with taking it out

profile/profile.go
```diff
@@ -715,6 +715,34 @@ func (s *Sample) HasLabel(key, value string) bool {
 	return false
 }
 
+// SetNumLabel sets the specified key to the specified value for all samples in the
+// profile. "unit" is a slice that describes the units that each corresponding member
+// of "values" is measured in (e.g. bytes or seconds).  If there is no relevant
+// unit for a given value, that member of "unit" should be the empty string.
```
aalexand, line 721: I would add a note that `unit` must either be nil or equal in length to `value`.

Added a few clarifying comments in this in https://github.com/google/pprof/pull/750.

blakej11: Done.

aalexand (approved):

PR #763, opened 2023-04-04
Add correct link for CI badge

hi-rustin:
We should only show badges for main branch workflows, not PRs.


aalexand:
I am not sure this change is necessary. I think the default is to show the status of the default branch, i.e. main. Do you have some evidence (documentation link or empirical) that this is not the case?

hi-rustin:
> I am not sure this change is necessary. I think the default is to show the status of the default branch, i.e. main. Do you have some evidence (documentation link or empirical) that this is not the case?

For example:
image

image

hi-rustin:
Oh, I see. It still failed. Because we canceled some tests on main branch workflows.

aalexand:
The PR does not update the image URL.

And https://github.com/google/pprof/workflows/ci/badge.svg?branch=main is red anyway.

aalexand:
I'm closing this. Totally feel free to re-open it with more evidence it is useful.

PR #773, opened 2023-04-23
Replace io.ReadFull to save an allocation

marselester:
A small change to save an allocation in `func (bu *Binutils) Open()`.

```
$ go test -bench=. -benchmem .
goos: darwin
goarch: amd64
pkg: tt
cpu: Intel(R) Core(TM) i5-10600 CPU @ 3.30GHz
BenchmarkReadFull-12    	  120055	      9929 ns/op	     120 B/op	       4 allocs/op
BenchmarkRead-12        	  117966	      9917 ns/op	     112 B/op	       3 allocs/op
PASS
ok  	tt	2.734s
```

<details>

```go
func BenchmarkReadFull(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if err := readFull(); err != nil {
			b.Fatal(err)
		}
	}
}

func BenchmarkRead(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if err := read(); err != nil {
			b.Fatal(err)
		}
	}
}

func readFull() error {
	f, err := os.Open("go.mod")
	if err != nil {
		return err
	}
	defer f.Close()

	var header [4]byte
	if _, err = io.ReadFull(f, header[:]); err != nil {
		return err
	}

	return nil
}

func read() error {
	f, err := os.Open("go.mod")
	if err != nil {
		return err
	}
	defer f.Close()

	var header [4]byte
	if _, err = f.Read(header[:]); err != nil {
		return err
	}

	return nil
}
```

</details>

### DIFF
--- a/internal/binutils/binutils.go
+++ b/internal/binutils/binutils.go
@@ -22,7 +22,6 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -300,7 +299,6 @@ func (bu *Binutils) Open(name string, start, limit, offset uint64, relocationSym
 	}
 
 	// Read the first 4 bytes of the file.
-
 	f, err := os.Open(name)
 	if err != nil {
 		return nil, fmt.Errorf("error opening %s: %v", name, err)
@@ -308,7 +306,7 @@ func (bu *Binutils) Open(name string, start, limit, offset uint64, relocationSym
 	defer f.Close()
 
 	var header [4]byte
-	if _, err = io.ReadFull(f, header[:]); err != nil {
+	if _, err = f.Read(header[:]); err != nil {
 		return nil, fmt.Errorf("error reading magic number from %s: %v", name, err)
 	}
 


internal/binutils/binutils.go
```diff
@@ -300,15 +299,14 @@ func (bu *Binutils) Open(name string, start, limit, offset uint64, relocationSym
...
-
 	f, err := os.Open(name)
 	if err != nil {
 		return nil, fmt.Errorf("error opening %s: %v", name, err)
 	}
 	defer f.Close()
 
 	var header [4]byte
-	if _, err = io.ReadFull(f, header[:]); err != nil {
+	if _, err = f.Read(header[:]); err != nil {
```
aalexand, line 309: I believe f.Read() does not guarantee it will read the specified number of bytes - the docs say "up to". So I think this change introduces a bug.

aalexand: Also, is there an example where this optimization matters? The time cost additional allocation (of four bytes?) should be negligible compared to the cost of the actual file read.

marselester: That's a good point! My assumption was that it should return those 4 bytes since it's reading from a disk and not from a terminal, a pipe, or a network socket.

According to https://man7.org/linux/man-pages/man2/read.2.html it could read less bytes from a disk than requested if `read()` syscall was interrupted by a signal.

> It is not an error if this number is smaller than the number of
> bytes requested; this may happen for example because fewer bytes
> are actually available right now (maybe because we were close to
> end-of-file, or because we are reading from a pipe, or from a
> terminal), or because read() was interrupted by a signal.

There is also an `EINTR` error that means the syscall was interrupted by a signal before any data was read. I assume the `Read` would return `n = 0` and a similar error in Go. In case it managed to read something and later was interrupted by a signal, there would be less bytes read than requested. I assume it should be able to read at least 512 bytes (a disk sector size) which covers the first 4 bytes.

That's a lot of speculations though on my side :D, it would be safer to close the PR.

aalexand: Yes, I think it's safer to leave the code as is - thanks! BTW, even though it's a read from file system, it can still be a network read if the path is network-mounted.

marselester: You're right.

PR #776, opened 2023-05-01
Add more clear err message upon disasm addr/regex

Louis-Ye:
b/255810947

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -435,16 +435,14 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	if len(syms) == 0 {
 		if address == nil { // User provides a symbol instead of an address
 			if len(symbols) == 0 {
-				return fmt.Errorf("no matches found for regexp %s in binaries", o.Symbol)
-			} else {
-				return fmt.Errorf("regex %s found in binaries, but not in the profile", o.Symbol)
+				return fmt.Errorf("no matches found for regexp %s in the binary", o.Symbol)
 			}
+			return fmt.Errorf("regex %s found in the binary, but not in the profile", o.Symbol)
 		} else { // User provides an address
 			if len(symbols) == 0 {
-				return fmt.Errorf("the address 0x%x is not found in binaries", *address)
-			} else {
-				return fmt.Errorf("address 0x%x found in binaries, but not in the profile", *address)
+				return fmt.Errorf("address 0x%x is not found in the binary", *address)
 			}
+			return fmt.Errorf("address 0x%x found in binary, not in the profile", *address)
 		}
 	}
 
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -515,8 +515,7 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 }
 
 // symbolsFromBinaries examines the binaries listed on the profile that have
-// associated samples, and returns the identified symbols matching rx, together
-// with a boolean indicating if the matched symbols have samples.
+// associated samples, and returns the identified symbols matching rx.
 func symbolsFromBinaries(prof *profile.Profile, g *graph.Graph, rx *regexp.Regexp, address *uint64, obj plugin.ObjTool) []*objSymbol {
 	// fileHasSamplesAndMatched is for optimization to speed up pprof: when later
 	// walking through the profile mappings, it will only examine the ones that have
@@ -532,7 +531,7 @@ func symbolsFromBinaries(prof *profile.Profile, g *graph.Graph, rx *regexp.Regex
 	var objSyms []*objSymbol
 	for _, m := range prof.Mapping {
 		// Skip the mapping if its file does not have samples or is not matched to
-		// the regexp (unless the regex is an address and the mapping's range covers
+		// the regexp (unless the regexp is an address and the mapping's range covers
 		// the address)
 		if !fileHasSamplesAndMatched[m.File] {
 			if address == nil || !(m.Start <= *address && *address <= m.Limit) {
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -433,11 +433,12 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	}
 
 	if len(syms) == 0 {
-		if address == nil { // User provides a symbol
+		// The symbol regexp case
+		if address == nil {
 			return fmt.Errorf("no matches found for regexp %s", o.Symbol)
 		}
 
-		// User provides an address
+		// The address case
 		if len(symbols) == 0 {
 			return fmt.Errorf("no matches found for address 0x%x", *address)
 		}
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -433,7 +433,19 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	}
 
 	if len(syms) == 0 {
-		return fmt.Errorf("no matches found for regexp: %s", o.Symbol)
+		if address == nil { // User provides a symbol instead of an address
+			if len(symbols) == 0 {
+				return fmt.Errorf("no matches found for regexp %s in binaries", o.Symbol)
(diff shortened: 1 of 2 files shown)


internal/report/report.go
```diff
@@ -433,7 +433,19 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	}
 
 	if len(syms) == 0 {
-		return fmt.Errorf("no matches found for regexp: %s", o.Symbol)
+		if address == nil { // User provides a symbol instead of an address
+			if len(symbols) == 0 {
+				return fmt.Errorf("no matches found for regexp %s in binaries", o.Symbol)
+			} else {
```
aalexand, line 439: nit: remove else and unindent its body. Same below.

internal/report/report.go
```diff
@@ -433,7 +433,19 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
...
+				return fmt.Errorf("regex %s found in binaries, but not in the profile", o.Symbol)
+			}
+		} else { // User provides an address
+			if len(symbols) == 0 {
+				return fmt.Errorf("the address 0x%x is not found in binaries", *address)
```
aalexand, line 444: nit: "the" is used here, but not in the message below, worth being consistent. I would remove it for brevity.

internal/report/report.go
```diff
@@ -433,7 +433,19 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	}
 
 	if len(syms) == 0 {
-		return fmt.Errorf("no matches found for regexp: %s", o.Symbol)
+		if address == nil { // User provides a symbol instead of an address
+			if len(symbols) == 0 {
+				return fmt.Errorf("no matches found for regexp %s in binaries", o.Symbol)
```
aalexand, line 438: I would use singular "in the binary", I think plural can be confusing to the user. When the user specifies the symbol or address they think of a single binary that should contain that pattern, not multiple binaries.

internal/report/report.go
```diff
@@ -433,7 +433,19 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
...
+			} else {
+				return fmt.Errorf("regex %s found in binaries, but not in the profile", o.Symbol)
```
aalexand, line 440: I wonder if it would make sense to simplify this message. It tries to be specific but I think it is may be too specific. Just saying something like "no data matched regexp %s" or something like that might be enough.

internal/report/report.go
```diff
@@ -433,7 +433,19 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	}
 
 	if len(syms) == 0 {
-		return fmt.Errorf("no matches found for regexp: %s", o.Symbol)
+		if address == nil { // User provides a symbol instead of an address
```
aalexand, line 436: Should we add tests covering the new code paths?

internal/report/report.go
Louis-Ye: The goal of this PR is to tell that user about the reason why there is no match: the provided symbol/address did not match the binary or did not match anything in the profile. I feel like simplifying the message would defeat the purpose

internal/report/report.go
Louis-Ye: Done.

internal/report/report.go
Louis-Ye: Done.

internal/report/report.go
Louis-Ye: Done.

internal/report/report.go
Louis-Ye: The report_test.go did not have this method PrintAssembly set up, so in order to test this code path, I think we need to mock the IO.writer and plugin.ObjTool. I was kind of hope that this PR being a simple one that just makes the error message more clear to the user. Do you think it is worth the effort?

internal/report/report.go
```diff
@@ -433,7 +433,17 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
...
+			if len(symbols) == 0 {
+				return fmt.Errorf("no matches found for regexp %s in the binary", o.Symbol)
+			}
+			return fmt.Errorf("regex %s found in the binary, but not in the profile", o.Symbol)
```
aalexand, line 440: nit:regexp

internal/report/report.go
```diff
@@ -433,7 +433,17 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	}
 
 	if len(syms) == 0 {
-		return fmt.Errorf("no matches found for regexp: %s", o.Symbol)
+		if address == nil { // User provides a symbol instead of an address
```
aalexand, line 436: "instead of address" in the comment reads a bit strange - as if address is what we expected but user entered symbol pattern. Both are OK and there is no primary expected input I think.

Also instead of "provides a symbol" it would be more accurate to say "symbol regexp".

Also "provides" should be "provided".

I maybe would simplify the comment to just: `// The symbol regexp case` and below to `// The address case`.

These are nits, but details are important in good engineering.

internal/report/report.go
aalexand: Ok, I guess. But I don't have a good way of knowing that this change works as intended.

Louis-Ye: I added the test, as I realized that there is a testdata that I can make use of, without too much effort to mocking it.

Also, I think due to some heuristic logic we place in to speed up the symbols finding process, I think essentially it is impossible to show matched symbols that do not have samples (unless it is an address). I have updated the comments in the code to make it more clear.

internal/report/report_test.go
```diff
@@ -506,3 +507,49 @@ func TestComputeTotal(t *testing.T) {
...
+				Symbol:       regexp.MustCompile(tc.symbol),
+
+				SampleValue: func(v []int64) int64 {
+					return v[1]
+				},
+				SampleUnit: profile.SampleType[1].Unit,
+			},
+		)
+
+		err := PrintAssembly(os.Stdout, rpt, &binutils.Binutils{}, -1)
```
aalexand, line 548: Can merge with if:

if err := ... ; err == nil || ...

internal/report/report_test.go
```diff
@@ -506,3 +507,49 @@ func TestComputeTotal(t *testing.T) {
...
+
+		if err == nil || err.Error() != tc.want {
+			t.Errorf("Got \"%v\", want \"%s\"", err, tc.want)
```
aalexand, line 551:
```
t.Errorf(`Got "%v", want %q`, err, tc.want)
```

internal/report/report_test.go
```diff
@@ -506,3 +507,49 @@ func TestComputeTotal(t *testing.T) {
...
+			symbol: "0x400000",
+			want:   "address 0x400000 found in binary, but the corresponding symbols do not have samples in the profile",
+		},
+	} {
+		rpt := New(
+			profile.Copy(),
+			&Options{
+				OutputFormat: List,
+				Symbol:       regexp.MustCompile(tc.symbol),
+
```
aalexand, line 540: nit nit: remove the empty line

internal/report/report.go
```diff
@@ -505,22 +514,26 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
...
 
-// symbolsFromBinaries examines the binaries listed on the profile
-// that have associated samples, and identifies symbols matching rx.
+// symbolsFromBinaries examines the binaries listed on the profile that have
+// associated samples, and returns the identified symbols matching rx.
 func symbolsFromBinaries(prof *profile.Profile, g *graph.Graph, rx *regexp.Regexp, address *uint64, obj plugin.ObjTool) []*objSymbol {
-	hasSamples := make(map[string]bool)
-	// Only examine mappings that have samples that match the
-	// regexp. This is an optimization to speed up pprof.
+	// fileHasSamplesAndMatched is for optimization to speed up pprof: when later
```
aalexand, line 520: Thanks for improving the comments!

aalexand (approved):

PR #782, opened 2023-05-08
Remove special coloring for self region in new flame graph view.

ghemawat:
Previously, the self cost part of a flame graph box  was shown with a tweaked color.

```
[<-- self --><-- rest of the cost -->]
             [child 1][child2][child3]
```

However this added clutter and was too subtle (I asked a couple of colleagues and they hadn't noticed it).

This change removes this coloring. The self cost is still available in the tooltip for each box and is also visible as the area to the left of the first child.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -490,11 +490,6 @@ Boxes are colored according to the name of the package in which the correspondin
 function occurs. E.g., in C++ profiles all frames corresponding to `std::` functions
 will be assigned the same color.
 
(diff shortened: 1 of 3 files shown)


aalexand:
@ghemawat Could you approve https://github.com/google/pprof/pull/780? This will fix the stuck CI in this and other PRs.

ghemawat:
Approved #780. Thanks.

aalexand (approved):

PR #783, opened 2023-05-14
Fixed a CSS bug that caused the upper right dialog box to be covered by the frame graph.

umlx5h:
Fixed a bug in the WebUI for new frame graphs where the dialog box in the upper right was displayed below the frame box.

before

after


aalexand:
@umlx5h From the screenshots looks like a good fix. Did you close it because the CLA check failed? You can sign it using the link in the failed CLA run FWIW.

PR #787, opened 2023-06-28
Hash-pin workflow Actions

### LINKED ISSUE #786: Hash-pin workflow Actions
When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash (or commit SHA) ensures the dependency is immutable and its behavior is guaranteed.

These dependencies can be kept up-to-date with dependabot. Whenever a new version of an Action is released, you'll receive a PR updating both its hash and the version comment.

I'll send a PR pinning the dependencies and adding dependabot along with this issue.

---

Sorry for not following the issue template, but I could figure out how to make it fit for this issue.

pnacht:
Fixes #786.

As described in the issue, this PR hash-pins all workflow Actions to protect the project from supply-chain attacks.

I've also set up dependabot to monitor the GitHub Actions. Whenever a new version is released, you'll receive a PR updating the Action's hash and the associated version comment.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,7 @@ jobs:
           xcode-version: '13.0'
     steps:
       - name: Update Go version using setup-go
-        uses: actions/setup-go@v4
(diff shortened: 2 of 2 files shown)


Louis-Ye:
Trying to get more context here. Is this a mandatory policy update to use workflow actions from Google internally (if so, is there a link for it; feel free to send it to yelouis@)? I can see that there is some mentioning at https://opensource.google/documentation/reference/github/services#actions, but it would be great to see if this is actually a policy. Thanks!

PR #790, opened 2023-07-27
Fix flamegraph display in the presence of --diff_base.

ghemawat:
Previously, the flamegraph display was broken in the presence of `--diff_base` because it did not account for the fact that some stacks could have negative values.  This change fixes the handling of negative values by tracking two numbers per displayed box that covers a set of stacks:

1. `sumpos`: the sum of the absolute values of increases in the stacks.
2. `sumneg`: the sum of the absolute values of decreases in the stacks.

The width of the box is proportional to `sumpos + sumneg`.

In addition, if a box has an overall decrease (`sumneg > sumpos`), the fraction of the box corresponding to the decrease is shaded green. (Similarly part of the box is shaded red if the box has an overall increase.)

Changed the display of legends and tooltips to clearly show the added and removed samples. E.g., the following shows that the cost of `Function` changed from `1.44s` to `1.3s` as well as the net change and the corresponding percentages.

```
0.14s (-1.8%) │ 1.44s (18.3%) 🠆 1.3s (16.5%) │ Function
```

Documented some of the Javascript types used for flamegraph rendering.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -485,6 +485,15 @@ Boxes are colored according to the name of the package in which the correspondin
 function occurs. E.g., in C++ profiles all frames corresponding to `std::` functions
 will be assigned the same color.
 
+When using the **--diff_base** option, box width is proportional to the sum of
+the increases and decreases in the sub-tree rooted at box. E.g., if the cost of
+one child of box decreases by 150 and the cost of another child increases by
+200, the box width will be proportional to 150+200. The net increase or decrease
+(the preceding example has a net increase of 200-150, i.e., 50) is indicated by
+a shaded region. The size of the shaded region is proportional to the net
+increase or net decrease. The shading is red for a net increase, and green for a
+net decrease.
+
 Inlining is indicated by the absence of a horizontal border between a caller and
 a callee. E.g., suppose X calls Y calls Z and the call from Y to Z is inlined into
 Y. There will be a black border between X and Y, but no border between Y and Z.
--- a/internal/driver/html/stacks.css
+++ b/internal/driver/html/stacks.css
@@ -28,7 +28,10 @@ body {
   position: absolute;
   overflow: hidden;
   box-sizing: border-box;
+  background: #d8d8d8;
 }
+.positive { position: absolute; background: #caa; }
+.negative { position: absolute; background: #aca; }
 /* Not-inlined frames are visually separated from their caller. */
 .not-inlined {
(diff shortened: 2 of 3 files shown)


s-kanev (commented):
LGTM overall, but I'll let @Louis-Ye chime in as the new pprof owner (we have an owner!).

ghemawat:
> LGTM overall, but I'll let @Louis-Ye chime in as the new pprof owner (we have an owner!).

Thanks for the quick review. Looking forward to Louis's comments.

Louis-Ye (commented):
LGTM for added code. And thanks for adding the detailed comments and explanation!

The only concern I have is that the test coverage decreases by 3%. 

Not a blocker, as I can see that test of the whole file is added in https://github.com/google/pprof/pull/716; but it would be nice if we can include test for this behaviour with a relatively low effort.

ghemawat:
Thanks.

> LGTM for added code. And thanks for adding the detailed comments and explanation!
> 
> The only concern I have is that the test coverage decreases by 3%.
> 
> Not a blocker, as I can see that test of the whole file is added in #716; but it would be nice if we can include test for this behaviour with a relatively low effort.

Yes, unfortunately none of the javascript in pprof is tested. Do you have any suggestions of how we would add javascript testing? I have a vague notion that we might need to install node in some github action and use that, but I wonder who here has the experience to figure out how to set that up. Maybe something like https://github.com/browser-actions/examples ?

Another alternative would be to do manually triggered tests: if pprof is invoked with a special flag, it passes that to the javascript, and the javascript code when it sees that flag runs self tests. Bit it would be better to figure out how to have real tests.

Should we continue the testing discussion outside this PR, either in an issue or in an email conversation?

Louis-Ye:
> none of the javascript in pprof is tested

Ah, that is right; giving another look to the code coverage report, it looks like the 3% of reduction is false alert: all the reductions are from go files, which I think it is probably due to versions different of the forked directory vs. the main directory.

> Do you have any suggestions of how we would add javascript testing

I don't have opinions on how to set this up neither. But definitely this is not a trivial effort in the scope of this PR.

> Should we continue the testing discussion outside this PR, either in an issue or in an email conversation?

Sure! I file another Github issue about this topic, we can discuss from there and with proper priority: https://github.com/google/pprof/issues/791

Louis-Ye (approved):
LGTM

PR #792, opened 2023-08-08
Prefix sample type menu entry ids with "sampletype-".

ghemawat:
Previously, we would use the sample type name as the HTML id for the corresponding menu entry. However some profiles contain sample type names that conflict with other HTML ids we use (in particular, the sample type name "sample" conflicted with a different id). Remove this name clash by prefixing every menu entry with "sampletype-".

some profiles caused the Javascript to break bec

### DIFF
--- a/internal/driver/html/common.js
+++ b/internal/driver/html/common.js
@@ -563,11 +563,11 @@ function viewer(baseUrl, nodes, options) {
     return str.replace(/([\\\.?+*\[\](){}|^$])/g, '\\$1');
   }
 
-  function setSampleIndexLink(id) {
-    const elem = document.getElementById(id);
+  function setSampleIndexLink(si) {
+    const elem = document.getElementById('sampletype-' + si);
     if (elem != null) {
       setHrefParams(elem, function (params) {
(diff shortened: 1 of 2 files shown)


internal/driver/html/common.js
```diff
@@ -563,11 +563,11 @@ function viewer(baseUrl, nodes, options) {
     return str.replace(/([\\\.?+*\[\](){}|^$])/g, '\\$1');
   }
 
-  function setSampleIndexLink(id) {
-    const elem = document.getElementById(id);
+  function setSampleIndexLink(si) {
+    const elem = document.getElementById('sampletype-'+si);
```
Louis-Ye, line 567: nit: I think the grammar convention in this file is having space before/after the operator +. Can you change it to that?

Louis-Ye (dismissed):
LGTM, with a minor comment.

internal/driver/html/common.js
ghemawat: Done. Thanks.

Louis-Ye (approved):
LGTM. Thanks!

Issue #777, opened 2023-05-02
Switch to the new flame graph implementation by default and remove d3 dependency


aalexand:
d3 dependency has been heavy toil to carry in pprof - version upgrades like #748 can be difficult when there is a version mismatch between the OSS version and the internal version used at Google. We also don't feel that the d3 usage provides enough value.

The current plan is to switch to the new flame graph implementation in the web UI, then after some time turn down the current d3-based flame graph implementation and the d3 dependency overall.

dhavaln-able:
Just checking, when the d3-dependency is supposed to be cleaned up?

aalexand:
@ghemawat We can probably finish this cleanup now - PTAL #825.

PR #801, opened 2023-09-01
Use go:embed for thirdparty

cristaloleg:
Simplifies handing of 3rd party files.

`npm install` also changed `package-lock.json` pretty heavily, not sure should I commit this too or not. WDYT ?


quasilyte (approved):

third_party/d3flamegraph/update.sh
```diff
@@ -21,36 +21,10 @@ D3FLAMEGRAPH_CSS="d3-flamegraph.css"
 
 cd $(dirname $0)
 
-generate_d3_flame_graph_go() {
+install_and_copy() {
```
aalexand, line 24: d3 based flame graph is going away in pprof - see #777.

So I would prefer to avoid any changes to it now.

cristaloleg: okay, so I will revert changes in `third_party/d3flamegraph` but will keep in `third_party/svgpan`. Can you confirm that?

aalexand: Yes, this sounds good.

cristaloleg: Done

aalexand (approved):

aalexand:
Will merge once CI passes.

PR #802, opened 2023-09-01
Add initial flamegraph web interface test.

ghemawat:
We now do some rudimentary testing of the flamegraph web interface. The test code is written in Javascript and uses a simple test fixture class.

### DIFF
--- a/internal/driver/testdata/testfixture.js
+++ b/internal/driver/testdata/testfixture.js
@@ -1,0 +1,30 @@
+// TestFixture records log messages and errors in an array that will
+// be returned to Go code. Each element in the result array is either
+// an array of the form ["LOG", ...]", or ["ERROR", ...].
+class TestFixture {
+  constructor() {
+    this.context = "";  // Added to front of all log and error messages.
+    this.result = [];
+  }
+
+  run(name, subtest) {
+    this.result.push(["LOG", "===", name]);
+    this.context = "";
+    subtest();
+    this.context = "";
+  }
+
+  // setContext arranges to add name(args) to all messages added
+  // until the next setContext or run call.
+  setContext(name, ...args) {
+    this.context = name + "(" + args.join(",") + ")"
+  }
+
+  log(...args) {
+    this.result.push(["LOG", this.context, ...args]);
+  }
+
+  err(...args) {
+    this.result.push(["ERROR", this.context, ...args]);
+  }
+};
--- a/internal/driver/testdata/testflame.js
+++ b/internal/driver/testdata/testflame.js
@@ -1,0 +1,90 @@
+function TestFlame() {
+  const test = new TestFixture();
+  const chart = document.getElementById("stack-chart");
+  if (!chart) {
+    test.err("could not find stack-chart");
+    return;
+  }
+  const chartRect = chart.getBoundingClientRect();
+
+  // Create map from box text to DOM element.
+  // TODO: Generalize to support multiple boxes for a given piece of text.
+  let boxMap;
+  fetchBoxes();
+  function fetchBoxes() {
+    boxMap = new Map();
+    const boxes = document.querySelectorAll(".boxbg");
+    for (let box of boxes) {
+      const text = box.innerText;
+      boxMap.set(text, box);
+    }
+  }
+
+  // rect gets the bounding box for box with text t.
+  function rect(t) {
+    const elem = boxMap.get(t);
+    if (!elem) {
+      test.err("did not find", t);
+      return null;
+    }
+    return elem.getBoundingClientRect();
+  }
+
+  // checkCalls checks that box with text a is positioned w.r.t. box with
+  // text b to indicate a call from a to b.  Expect a gap of the specified
+  // number of rows.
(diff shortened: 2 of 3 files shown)


internal/driver/testdata/testflame.js
aalexand, line 1: I assume the test is done in JS because writing it on the Go side using [chromedp API](https://pkg.go.dev/github.com/chromedp/chromedp) would be much more verbose?

ghemawat: I couldn't find chromedp apis for fetching dom contents and bounding boxes so would have to use JavaScript eval for that. Once I realized that it seemed there would be lots of boilerplate for each such fetch (documentation, type conversion, etc.) so I moved the test logic as well.

I think this will also help with testing individual js functions in future tests.

internal/driver/testdata/testfixture.js
```diff
@@ -0,0 +1,46 @@
...
+      str += sep;
+      str += arg;
+      sep = ",";
+    }
+    str += "):";
+    this.context = str;
+  }
+
+  log(...args) {
+    this.result.push(this.toArray("LOG", this.context, ...args));
```
aalexand, line 32: I don't know JS that well, but poking around a bit I think this can be just `this.result.push("LOG", this.context, ...args);`.

internal/driver/testdata/testflame.js
```diff
@@ -0,0 +1,86 @@
+(function () {
```
aalexand, line 1: nit: I would maybe give the function a name and then call it explicitly at the end. This syntax with parens gave me a bit of a pause when reading the code.

internal/driver/testdata/testflame.js
```diff
@@ -0,0 +1,86 @@
...
+    const ra = rect(a);
+    const rb = rect(b);
+    if (!ra || !rb) return;
+
+    const pixelGap = gap * rb.height;
+    if (rb.top != ra.bottom + pixelGap) {
+      test.err("not above");
+    }
+    // TODO: Allow checking boxes above pivots.
+    if (rb.left < ra.left || rb.right > rb.right) {
```
aalexand, line 44: `rb.right` comparison should be `rb.right > ra.right`?

internal/driver/testdata/testflame.js
```diff
@@ -0,0 +1,86 @@
...
+      test.err("horizontal span is not nested inside horizontal span of");
```
aalexand, line 45: The "inside horizontal span of" part reads unfinished. Or maybe I misread it.

internal/driver/testdata/testflame.js
```diff
@@ -0,0 +1,86 @@
...
+    for (let box of boxes) {
+      const text = box.innerText;
+      boxMap.set(text, box);
+    }
+  }
+
+  // rect gets the bounding box for box with text t.
+  function rect(t) {
+    const elem = boxMap.get(t);
+    if (!elem) test.err("did not find", t);
```
aalexand, line 26: Should we return here? Otherwise it seems we'd be dereferencing a null in the call in the next line.

internal/driver/testdata/testfixture.js
ghemawat: Thanks. You were close. I needed:

```
this.result.push(["LOG", this.context, ...args]);
```

internal/driver/testdata/testflame.js
ghemawat: Done.

ghemawat:
Cleaned up branch history by moving all changes made here into a single commit at the end.

internal/driver/testdata/testflame.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  // checkWidth checks that the width of the box with text t is approximately
+  // the specified fraction of the total width.
+  function checkWidth(t, fraction) {
+    test.setContext("checkWidth", t, fraction);
+    const r = rect(t);
+    if (!r) return;
+    const expect = (chartRect.width - 4) * fraction;
```
aalexand, line 58: nit: a named constant for the 4 might be good, it's a bit obscure as is.

aalexand (approved):

internal/driver/testdata/testflame.js
Louis-Ye: My 2 cents: personally I think amending the commit is better than squashing it into one, because it would be easier to relate the comments and CI runs to code in the PR.

Louis-Ye (approved):
LGTM

internal/driver/testdata/testflame.js
ghemawat: I think you meant "adding a new commit" instead of "amending the commit"?

I agree with the decision to not squash until merging into main. Also makes it easier for reviewers to see the changes since the last round of reviews.

Thanks.

aalexand: I don't have strong preference, separate commits are easier so probably the default choice. One added advantage is being able to move out a specific part of the change into a separate PR if requested by a reviewer - easier to do if it's a separate commit. Doesn't happen that often though, at least for this repo.

Issue #816, opened 2023-11-03
If the file under the current path contains a colon, go tool pprof will parse it into a url and will not work properly.


cuishuang:
Please answer these questions before submitting your issue. Thanks!

### What version of pprof are you using?

the output of go env:

```shell
GO111MODULE='on'
GOARCH='arm64'
GOBIN=''
GOCACHE='/Users/myname/Library/Caches/go-build'
GOENV='/Users/myname/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='arm64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/myname/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='darwin'
GOPATH='/Users/myname/go'
GOPRIVATE=''
GOPROXY=''
GOROOT='/Users/myname/mypath/go'
GOSUMDB='off'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/Users/myname/mypath/go/pkg/tool/darwin_arm64'
GOVCS=''
GOVERSION='devel go1.22-6dd7462a04 Fri Oct 13 17:10:31 2023 +0000' # I compiled it myself, it can be ignored and has nothing to do with the version.
GCCGO='gccgo'
AR='ar'
CC='clang'
CXX='clang++'
CGO_ENABLED='1'
GOMOD='/Users/myname/mypath/go/src/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/9t/839s3jmj73bcgyp5x_xh3gw00000gn/T/go-build3891298911=/tmp/go-build -gno-record-gcc-switches -fno-common'

```

### What operating system and processor architecture are you using?

'darwin'

'arm64' 

### What did you do?

I helped a colleague write a sample code to obtain the pprof file when the memory exceeds the threshold.

```go
package main

import (
	"fmt"
	"os"
	"runtime"
	"runtime/pprof"
	"time"
)

func main() {

[...]

PR #817, opened 2023-11-03
Fix exception when file type contains colon

cuishuang:
Fixes https://github.com/google/pprof/issues/816

### DIFF
--- a/internal/driver/fetch.go
+++ b/internal/driver/fetch.go
@@ -492,17 +492,23 @@ mapping:
 func fetch(source string, duration, timeout time.Duration, ui plugin.UI, tr http.RoundTripper) (p *profile.Profile, src string, err error) {
 	var f io.ReadCloser
 
-	if sourceURL, timeout := adjustURL(source, duration, timeout); sourceURL != "" {
-		ui.Print("Fetching profile over HTTP from " + sourceURL)
-		if duration > 0 {
-			ui.Print(fmt.Sprintf("Please wait... (%v)", duration))
+	// First determine whether the source is a file, if not, it will be treated as a URL.
+	if _, openErr := os.Stat(source); openErr == nil {
+		if isPerfFile(source) {
+			f, err = convertPerfData(source, ui)
+		} else {
+			f, err = os.Open(source)
 		}
-		f, err = fetchURL(sourceURL, timeout, tr)
-		src = sourceURL
-	} else if isPerfFile(source) {
-		f, err = convertPerfData(source, ui)
 	} else {
-		f, err = os.Open(source)
+		sourceURL, timeout := adjustURL(source, duration, timeout)
+		if sourceURL != "" {
+			ui.Print("Fetching profile over HTTP from " + sourceURL)
+			if duration > 0 {
+				ui.Print(fmt.Sprintf("Please wait... (%v)", duration))
+			}
+			f, err = fetchURL(sourceURL, timeout, tr)
+			src = sourceURL
+		}
 	}
 	if err == nil {
 		defer f.Close()
--- a/internal/driver/fetch_test.go
+++ b/internal/driver/fetch_test.go
@@ -192,12 +192,28 @@ func TestFetch(t *testing.T) {
 	type testcase struct {
 		source, execName string
 	}
-
-	for _, tc := range []testcase{
+	ts := []testcase{
 		{path + "go.crc32.cpu", ""},
 		{path + "go.nomappings.crash", "/bin/gotest.exe"},
 		{"http://localhost/profile?file=cppbench.cpu", ""},
-	} {
+	}
+	// Test that paths with a colon character are recognized as file paths
+	// if the file exists, rather than as a URL. We have to skip this test
+	// on Windows since the colon char is not allowed in Windows paths.
+	if runtime.GOOS != "windows" {
+		src := filepath.Join(path, "go.crc32.cpu")
+		dst := filepath.Join(t.TempDir(), "go.crc32.cpu_2023-11-11_01:02:03")
+		data, err := os.ReadFile(src)
+		if err != nil {
+			t.Fatalf("read src file %s failed: %#v", src, err)
+		}
+		err = os.WriteFile(dst, data, 0644)
+		if err != nil {
+			t.Fatalf("create dst file %s failed: %#v", dst, err)
+		}
+		ts = append(ts, testcase{dst, ""})
+	}
+	for _, tc := range ts {
 		p, _, _, err := grabProfile(&source{ExecName: tc.execName}, tc.source, nil, testObj{}, &proftest.TestUI{T: t}, &httpTransport{})
 		if err != nil {
 			t.Fatalf("%s: %s", tc.source, err)


internal/driver/fetch.go
```diff
@@ -589,7 +589,7 @@ func adjustURL(source string, duration, timeout time.Duration) (string, time.Dur
 		// url.Parse treats "hostname" as the scheme.
 		u, err = url.Parse("http://" + source)
 	}
-	if err != nil || u.Host == "" {
+	if err != nil || u.Host == "" || isFile(source) {
```
aalexand, line 592: I would put this check in the `fetch()` function, before we even calling `adjustURL()`. This will colocate it with other details of handling files vs URLs making it easier to read the code.

Can you also see if you can add a test that fails before the change and passes with the fix?

cuishuang: Of course. Thanks

cuishuang: Modified, please review again and give suggestions

internal/driver/fetch.go
```diff
@@ -614,3 +616,16 @@ func adjustURL(source string, duration, timeout time.Duration) (string, time.Dur
...
 	return u.String(), timeout
 }
+
+// isFile validates if the profile source is a file, but it will be recognized
+// as a link by url.Parse (when it contains `:`, like mem_2023-11-01_15:04:05).
+// If such a file exists in the current directory, it will be recognized as a
+// file rather than a link. Because colons are not allowed in file names on
+// Windows systems, the Windows case should be ignored.
+func isFile(source string) bool {
+	if runtime.GOOS == "windows" {
```
aalexand, line 626: What happens if we don't special-case Windows? I'm trying to understand why not just drop this check - if the path contains a colon this function will return false and the calling code will do what is needed I think?

internal/driver/fetch.go
```diff
@@ -492,7 +492,9 @@ mapping:
 func fetch(source string, duration, timeout time.Duration, ui plugin.UI, tr http.RoundTripper) (p *profile.Profile, src string, err error) {
 	var f io.ReadCloser
 
-	if sourceURL, timeout := adjustURL(source, duration, timeout); sourceURL != "" {
+	if isFile(source) {
```
aalexand, line 495: I think this code as written breaks opening perf.data files? We never get to that condition now.

internal/driver/fetch.go
cuishuang: In the Windows system, colons are not allowed in file names, so the Windows system naturally avoids the ambiguity of whether the source is a file or a URL. Here, if it is judged that the source has a colon, it will return false (but I understand it should be true?). If there is a case like 127.0.0.1:8080, there will be a problem. I think for a case like 127.0.0.1:8080, the processing steps should be: first try to see if there is a file with the same name in the current path, if not, then treat it as a url.

I don’t know if I expressed it clearly. Thanks very much for your review. I look forward to your reply and any discussion.

internal/driver/fetch.go
cuishuang: I think it shouldn't be the case. The previous logical sequence is to first judge adjustURL. If the conditions are not met, then judge isPerfFile. If the conditions are still not met, os.Open is executed at last. The changed logic is just to determine whether the source string corresponds to a file in the current directory before executing adjustURL. If so, it will be treated as a file first regardless of whether it contains a colon.  

I think it is equivalent to put the isFile(source) judgment in the adjustURL func, just like I did when I submitted it for the first time.

internal/driver/fetch.go
aalexand: But I think there is no harm in doing the check always? Having a file named "127.0.0.1:8080" is extremely unlikely. Once we make that change, I think there is not much value in having the `isFile()` function and I'd just inline it in the code of `fetch()` for simplicity.

internal/driver/fetch.go
aalexand: I'm talking just about the current code, for simplicity. Here is what I think happens now: We first check `isFile()` which will return true for a perf.data file and so we will never get to the `isPerfFile()` check or `convertPerfData()` call. Instead we will try to parse the perf.data file content as a profile.proto which will of course fail. Does this make sense?

cuishuang: This problem does exist. Previously, isPerfFile and convertPerfData were executed for local files in perf format that did not contain colons. But now it will go to isFile.

I should adjust the order of execution and put the isPerfFile check first if isFile is true, and execute adjustURL if it is false.

internal/driver/fetch.go
cuishuang: I would like to modify it this way, please see if it is possible. Thanks very much for your patient guidance

```go
func fetch(source string, duration, timeout time.Duration, ui plugin.UI, tr http.RoundTripper) (p *profile.Profile, src string, err error) {
	var f io.ReadCloser

	if isFile(source) {
		if isPerfFile(source) {
			f, err = convertPerfData(source, ui)
		} else {
			f, err = os.Open(source)
		}
	} else {
		sourceURL, timeout := adjustURL(source, duration, timeout)
		if sourceURL != "" {
			ui.Print("Fetching profile over HTTP from " + sourceURL)
			if duration > 0 {
				ui.Print(fmt.Sprintf("Please wait... (%v)", duration))
			}
			f, err = fetchURL(sourceURL, timeout, tr)
			src = sourceURL
		}
	}
	if err == nil {
		defer f.Close()
		p, err = profile.Parse(f)
	}
	return
}

func isFile(source string) bool {
	_, err := os.Stat(source)
	return err == nil
}
```

cuishuang: Thanks! Please review it again.

cuishuang:
Is there anything else you need me to do? I'm happy to do it.

internal/driver/fetch.go
```diff
@@ -492,17 +492,24 @@ mapping:
 func fetch(source string, duration, timeout time.Duration, ui plugin.UI, tr http.RoundTripper) (p *profile.Profile, src string, err error) {
 	var f io.ReadCloser
 
-	if sourceURL, timeout := adjustURL(source, duration, timeout); sourceURL != "" {
-		ui.Print("Fetching profile over HTTP from " + sourceURL)
-		if duration > 0 {
-			ui.Print(fmt.Sprintf("Please wait... (%v)", duration))
+	// first determine whether the source is a file, if not, it will be treated as a URL.
+	_, openErr := os.Stat(source)
```
aalexand, line 496: You can merge these lines:

```
if _, openErr := os.Stat(source); openErr == nil {
```

internal/driver/testdata/go.crc32.cpu_2023-11-11_01_02_03
aalexand, line 1: I don't think we need a new file. You can copy it programmatically to https://pkg.go.dev/testing#T.TempDir renaming as you need. Then you don't need to rename it back.

internal/driver/fetch_test.go
```diff
@@ -193,11 +193,28 @@ func TestFetch(t *testing.T) {
...
 		{path + "go.nomappings.crash", "/bin/gotest.exe"},
 		{"http://localhost/profile?file=cppbench.cpu", ""},
-	} {
+	}
+
+	if runtime.GOOS != "windows" {
+		ts = append(ts, testcase{path + "go.crc32.cpu_2023-11-11_01:02:03", ""})
+		err := os.Rename(path+"go.crc32.cpu_2023-11-11_01_02_03", path+"go.crc32.cpu_2023-11-11_01:02:03")
+		if err != nil {
+			t.Errorf("rename file err: %v", err)
```
aalexand, line 206: Fatalf(), then no return is needed.

cuishuang: Done

internal/driver/fetch.go
cuishuang: Done

internal/driver/testdata/go.crc32.cpu_2023-11-11_01_02_03
cuishuang: This way is much more elegant indeed. Grateful!

cuishuang:
It has been modified, please review again. Thanks so much!

internal/driver/fetch_test.go
```diff
@@ -193,11 +193,31 @@ func TestFetch(t *testing.T) {
...
 
-	for _, tc := range []testcase{
+	ts := []testcase{
 		{path + "go.crc32.cpu", ""},
 		{path + "go.nomappings.crash", "/bin/gotest.exe"},
 		{"http://localhost/profile?file=cppbench.cpu", ""},
-	} {
+	}
+
+	if runtime.GOOS != "windows" {
```
aalexand, line 202: nit: I think empty lines in the added code can be removed.

aalexand, line 202: Could you add a short comment to explain what case we are testing and why Windows is skipped? Something like "Test that paths with a colon character are recognized as file paths if the file exists, rather than as a URL. We have to skip this test on Windows since the colon char is not allowed in Windows paths." (wrap the comment at 80 columns as necessary)

aalexand (commented):
Looks good, just a couple minor comments.

PR #818, opened 2023-11-09
Add column numbers to profile.proto.

### LINKED ISSUE #687: Proposal: Add Column field to Line message
It would be really useful to have a `column` field along with the line number field in the `Line` message. This is especially useful for runtimes that use source code compacting/obfuscation like Javascript.

In `profile.proto` this would look like:

```diff --git a/proto/profile.proto b/proto/profile.proto
index ee0391f..92f9fef 100644
--- a/proto/profile.proto
+++ b/proto/profile.proto
@@ -195,6 +195,8 @@ message Line {
   uint64 function_id = 1;
   // Line number in source code.
   int64 line = 2;
+  // Column number in the source code.
+  int64 column = 3;
 }

 message Function {
```

Additional changes to the `pprof` tool:
* Output column information if it is present when `-lines` is used.
* Where `pprof` currently outputs `<path>:<line>` make it output `<path>:<line>:<column>` when column is available.
  * Columns are 1-indexed, so 0 column doesn’t exit. 
* `merge.go` needs updating to take into account column numbers, also `(*Profile).Aggregate(), aggregate()` in `driver.go`. We would need to check all the places where `Line.line` is used. 

Please let me know what you think.

Thanks!

snehasish:
This change adds column numbers to the line message in profile.proto. This allows users to distinguish between souce code locations on the same line. Only the llvm based addr2line is capable of reading column number information from dwarf and PE debug information. Other changes include:
* Add "columns" option for output granularity.
* Account for column numbers during profile merge.
* Update the encoder and decoder.
* Update golden test files for legacy profiles.

Fixes #687


internal/driver/config.go
```diff
@@ -124,7 +125,7 @@ func init() {
 	// take on one of a bounded set of values.
 	choices := map[string][]string{
 		"sort":        {"cum", "flat"},
-		"granularity": {"functions", "filefunctions", "files", "lines", "addresses"},
+		"granularity": {"functions", "filefunctions", "files", "lines", "columns", "addresses"},
```
aalexand, line 128: Would it make sense to instead make this an option like `columns` similar to `noinlines`? But I guess `pprof -lines -columns` would look weird.

profile/profile_test.go
```diff
@@ -997,7 +1075,7 @@ func locationHash(s *Sample) string {
 	var tb string
 	for _, l := range s.Location {
 		for _, ln := range l.Line {
-			tb = tb + fmt.Sprintf("%s:%d@%d ", ln.Function.Name, ln.Line, l.Address)
+			tb = tb + fmt.Sprintf("%s:%d:%d@%d ", ln.Function.Name, ln.Line, ln.Column, l.Address)
```
aalexand, line 1078: When columns are present, are they 1-based? If yes, maybe we can omit the ":0" part when the column is zero? This would make the `pprof -raw` output less chatty for the case when the column information is missing.

The downside is that if the column information is only partially present then the `-raw` output would partially have and partially haven't the column number which can be obscure to the user.

aalexand (commented):
Looks good for the most part, a couple questions.

profile/profile_test.go
snehasish: Yes, column numbers are 1-based when available (see https://dwarfstd.org/doc/DWARF5.pdf, Sec 6.2.2 pg 157). The value 0 is reserved to indicate that a statement begins at the “left edge” of the line, but implementations don't use this as far as I know. 

IMO it's clearer to retain the value, an additional zero per line is not too chatty. Happy to reconsider if you feel strongly.

internal/driver/config.go
```diff
@@ -157,6 +158,7 @@ func init() {
 		"sort":                 "sort",
 		"granularity":          "g",
 		"noinlines":            "noinlines",
+		"columns":              "columns",
```
aalexand, line 161: Maybe name it "showcolumns" so that `-lines -showcolumns` looks a bit better?

internal/driver/driver.go
```diff
@@ -246,6 +246,11 @@ func aggregate(prof *profile.Profile, cfg config) error {
 		function = true
 		filename = true
 		linenumber = true
+	case "columns":
```
aalexand, line 249: I think this shouldn't be in the granularity switch anymore and should be handled as option instead?

internal/driver/config.go
aalexand: We should remove it from the granularity list here then?

internal/driver/config.go
```diff
@@ -51,6 +51,7 @@ type config struct {
 	TagShow      string  `json:"tagshow,omitempty"`
 	TagHide      string  `json:"taghide,omitempty"`
 	NoInlines    bool    `json:"noinlines,omitempty"`
+	Columns      bool    `json:"columns,omitempty"`
```
aalexand, line 54: Maybe `ShowColumns`? See another comment.

aalexand (commented):
Could you check that pprof reports properly output the column when line granularity is used and the columns are enabled? I'm thinking scenarios like:

* `-top -lines -showcolumns`
* `-svg -lines -showcolumns`
* `-http $HOSTNAME:8080 -lines -showcolumns` then checking all views.

internal/driver/config.go
snehasish: Done.

internal/driver/config.go
snehasish: Done.

internal/driver/driver.go
snehasish: Done.

snehasish:
> Could you check that pprof reports properly output the column when line granularity is used and the columns are enabled? I'm thinking scenarios like:
> 
> * `-top -lines -showcolumns`
> * `-svg -lines -showcolumns`
> * `-http $HOSTNAME:8080 -lines -showcolumns` then checking all views.

Here is the output from top for a sample app (when non-zero column numbers are available via llvm-symbolizer):

```
$ go run pprof.go -top -lines -showcolumns sample profile.pb.gz

Type: cycles:u_event
Showing nodes accounting for 12157325838, 99.93% of 12165793058 total
Dropped 13 nodes (cum <= 60828965)
2413918557 19.84% 19.84% 2413918557 19.84%  main /tmp/sample.c:33:20
2289792795 18.82% 38.66% 2289792795 18.82%  main /tmp/sample.c:33:9
2051102236 16.86% 55.52% 2051102236 16.86%  compute_flag /tmp/sample.c:13:9
1185046082  9.74% 65.26% 1185046082  9.74%  main /tmp/sample.c:34:15
```
Also added a couple of screenshots which show column numbers in the web views.

profile/profile_test.go
snehasish: In the end I chose to suppress the zero column numbers while reporting to minimize updates to the tests.

snehasish:
Post-holiday ping :)

internal/symbolizer/symbolizer_test.go
```diff
@@ -242,23 +242,27 @@ func checkSymbolizedLocation(a uint64, got []profile.Line) error {
 		if g.Line != int64(w.Line) {
 			return fmt.Errorf("want lineno: %d, got %d", w.Line, g.Line)
```
aalexand, line 243: nit: "line" here or "columnno" below?

profile/profile_test.go
aalexand: I don't see it being suppressed, am I missing something?

snehasish: This is the test-only locationHash helper. Here I think we need the extra resolution here. My comment about suppressing the column number is implemented in `graph.go`, see L178-181 which updated the `NameComponents` method. 

```
s := fmt.Sprintf("%s:%d", i.File, i.Lineno)
if i.Columnno != 0 {
	s += fmt.Sprintf(":%d", i.Columnno)
}
```

internal/symbolizer/symbolizer_test.go
snehasish: Done.

aalexand (approved):

P403n1x87:
CPython >= 3.11 allows extracting line end and column end information. Would it be conceivable to have these extra fields to also carry this information in pprof files?

aalexand:
Unlikely. We are hesitant to add fields that are too niche. Even adding the column number was controversial enough. I don't anticipate adding more fields like that.

PR #825, opened 2023-12-17
Remove old d3 flamegraph code and d3 dependency.

### LINKED ISSUE #777: Switch to the new flame graph implementation by default and remove d3 dependency

aalexand:
Fixes #777.


internal/driver/webui.go
```diff
@@ -125,16 +124,16 @@ func serveWebInterface(hostport string, p *profile.Profile, o *plugin.Options, d
...
-			"/flamegraph2":   http.HandlerFunc(ui.stackView), // Support older URL
-			"/saveconfig":    http.HandlerFunc(ui.saveConfig),
-			"/deleteconfig":  http.HandlerFunc(ui.deleteConfig),
+			"/":           http.HandlerFunc(ui.dot),
+			"/top":        http.HandlerFunc(ui.top),
+			"/disasm":     http.HandlerFunc(ui.disasm),
+			"/source":     http.HandlerFunc(ui.source),
+			"/peek":       http.HandlerFunc(ui.peek),
+			"/flamegraph": http.HandlerFunc(ui.stackView),
+			// TODO: Remove supporting this older flamegraph2 URL at some point.
```
ghemawat, line 133: Should we perhaps change this to a permanent HTTP redirect to avoid people bookmarking /flamegraph2?

Also, do we want to preserve support for /flamegraphold (either redirect or an alias) to avoid breaking existing bookmarks, links, etc.

ghemawat (approved):
Thanks!

internal/driver/webui.go
aalexand: Yes, makes sense, but after spending 10 minutes on this it turned out to be trickier than I thought - http.Redirect("flamegraph") doesn't work since we have this higher level stripping of "/ui" and I assume we don't want to hardcode "/ui" in the redirect path since this defeats the purpose of this prefix being transparent to the implementation. I tried using a relative path in the redirect location but that didn't work either for some reason, need to look into this more. Maybe the StripPrefix() handler need to inspect the response and recognize redirects adding the /ui prefix back, but that means that whoever embeds this server in their server under a custom prefix will have to do the same and meanwhile the redirects will be broken for them. I might be missing a trivial solution? Anyway, will need to look more into this.

ghemawat: Can we get the current request's URL and replace a trailing /flamegraph2 with /flamegraph to form the new URL?

aalexand: As far as I can tell, StripPrefix() [modifies the request URL destructively](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/net/http/server.go;l=2165).

ghemawat: If I understand correctly, /foo redirects to /ui/foo. Then /ui/foo is handled by stripping the /ui prefix and then looking up the handler in the handler map.

What if we drop the StripPrefix call and have the handler map keys start with /ui? But perhaps some clients are doing something weirder?

Alternative: instead of calling http.StripPrefix, we provide our own wrapper that first examines the original URL and either (a) redirects from flamegraph2 to flamegraph, or (b) returns the precomputed http.StripPrefix() result.

aalexand: I ended up using a small bit of code to return a relative path redirect instead of using `http.Redirect()` which does not support that. This is similar to [this code](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/net/http/fs.go;l=607-612;drc=e973d242614d30b7a3aaf8f65d279b062976fa9f) in Go runtime itself so I think this is reasonable.

ghemawat: Looks great!

ghemawat (approved):

PR #831, opened 2024-02-23
Use html/template to produce source listing.

ghemawat:
We previously used an ad-hoc combination of calls to printf and HTML escaping routines to produce HTML for source code listing. We now use html/template to do the formatting. This should give us HTML generation that is less susceptible to escaping issues.

The change keeps the output very similar to the old output (mostly a small number of spacing changes that do not affect what a user can see).

Also added a test that does a quick check of the source view displayed in a browser.


internal/report/report.go
```diff
@@ -111,11 +111,10 @@ func Generate(w io.Writer, rpt *Report, obj plugin.ObjTool) error {
 		return printAssembly(w, rpt, obj)
 	case List:
 		return printSource(w, rpt)
-	case WebList:
-		return printWebSource(w, rpt, obj)
 	case Callgrind:
 		return printCallgrind(w, rpt)
 	}
+	// Note: WebList handling is in driver package.
```
Louis-Ye, line 117: If WebList is the only "enum" that is defined but not handled, maybe we should print out this enum value in the following "expected output format"? To make it easier debug if this method ss accidentally called for WebList.

internal/driver/driver.go
```diff
@@ -155,6 +163,18 @@ func generateReport(p *profile.Profile, cmd []string, cfg config, o *plugin.Opti
 	return out.Close()
 }
 
+func weblist(dst io.Writer, rpt *report.Report, obj plugin.ObjTool) error {
```
Louis-Ye, line 166: Can you rename this method to start with a verb as well to be consistent with the rest of the method? Maybe just the old printWebList can work, as it does print to the input io writer.

internal/report/source.go
```diff
@@ -752,89 +804,58 @@ func (sp *sourcePrinter) objectFile(m *profile.Mapping) plugin.ObjFile {
...
-	fmt.Fprintln(w, weblistPageCSS)
-	fmt.Fprintln(w, weblistPageScript)
-	fmt.Fprint(w, "</head>\n<body>\n\n")
-
-	var labels []string
-	for _, l := range ProfileLabels(rpt) {
-		labels = append(labels, template.HTMLEscapeString(l))
+// webListLine returns the contents of a single line in a web listing. This includes
+// the source line and the corresponding assembly.
+func webListLine(lineNo int, flat, cum int64, lineContents string,
```
Louis-Ye, line 809: makeWebListLine? To be naming consistent with the rest of method

internal/report/source.go
```diff
@@ -752,89 +804,58 @@ func (sp *sourcePrinter) objectFile(m *profile.Mapping) plugin.ObjFile {
...
-		printNested(w, srcIndent, assembly, reader, rpt)
+		line.Instructions = webListInstructions(srcIndent, assembly, reader, rpt)
 	}
-	fmt.Fprintln(w)
+	return line
 }
 
-func printNested(w io.Writer, srcIndent int, assembly []assemblyInstruction, reader *sourceReader, rpt *Report) {
-	fmt.Fprint(w, "<span class=asm>")
+func webListInstructions(srcIndent int, assembly []assemblyInstruction, reader *sourceReader, rpt *Report) []WebListInstruction {
```
Louis-Ye, line 839: makeWebListInstructions? To be naming consistent with the rest of method

Louis-Ye (approved):
LGTM in general, with a few minor comments.

ghemawat:
Oops. I missed the comments in the github UI. I will send a separate PR handling them.

PR #840, opened 2024-03-24
fix: change elfexec.GetBuildID() to accept ELF file instead of reader 

testwill:


internal/binutils/binutils.go
```diff
@@ -434,6 +434,7 @@ func (b *binrep) openELF(name string, start, limit, offset uint64, relocationSym
 
 	buildID := ""
 	if f, err := os.Open(name); err == nil {
+		defer f.Close()
 		if id, err := elfexec.GetBuildID(f); err == nil {
```
aalexand, line 438: Looking at the code, I actually propose we change `elfexec.GetBuildID()` to accept ELF file instead of reader. We already open the ELF file above, there is no need to open it again.

idamidzin (approved):

aalexand (approved):

PR #851, opened 2024-04-24
Flamegraph uses seconds instead of hours when appropriate.

ghemawat:
The Javascript code now behave the same as the Go code. (It switches to the next unit only if one unit is at least as big as the value being displayed.) The main effect is that we no longer use "hours" as the unit for time values in the range [36s,3600s). We only switch to using "hours" when the measured time is at least one hour. This is easier for users to understand: e.g., a time value of hundred seconds now displays as "100s", not "0.03hrs".

Also added a unittest for the Javascript unit picking code.

### DIFF
--- a/internal/driver/html/stacks.js
+++ b/internal/driver/html/stacks.js
@@ -13,23 +13,6 @@ function stackViewer(stacks, nodes) {
   const FONT_SIZE = 12;
   const MIN_FONT_SIZE = 8;
 
-  // Mapping from unit to a list of display scales/labels.
-  // List should be ordered by increasing unit size.
-  const UNITS = new Map([
-    ['B', [
-      ['B', 1],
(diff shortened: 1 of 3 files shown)


internal/driver/html/stacks.js
```diff
@@ -606,3 +574,40 @@ function stackViewer(stacks, nodes) {
     return hsl;
   }
 }
+
+// Mapping from unit to a list of display scales/labels.
+// List should be ordered by increasing unit size.
+const pprofUnits = new Map([
```
aalexand, line 580: I wish we switch to the unit definitions in google3/third_party/golang/pprof/internal/measurement/measurement.go (towards the end of that file is the table).

aalexand (approved):

PR #852, opened 2024-04-24
Share unit definitions between Go and Javascript.

ghemawat:
Go unit definitions (from measurement package) are now converted to JSON and delivered to Javascript code so that we do not need to duplicate these definitions in Javascript.

Had to make the unit type definitions exported with public fields.

Also had to change a test case to expect "us" instead of "µs".


aalexand (approved):
Thanks!

PR #854, opened 2024-04-27
Apply grab and grabbing cursor to graph UI element

### LINKED ISSUE #837: Nice to have: 'grab' cursor when dragging the graph

Github head.

### What UI question did you have in mind ?

Not a bug. This is merely UI idea that came to mind playing with the web-UI, specifically the 'graph' view.

When dragging the graph view, the mouse-cursor stays the same as regular pointer.
However, from a user feedback perspective -- seeing that this is holding the graph and moving it around -- it would be neat if the cursor was changing into the 'grab cursor' symbol (the little hand that 'holds on' to the thing being dragged). 

I think the [CSS cursor property](https://www.w3schools.com/cssref/pr_class_cursor.php) would be `cursor: grabbing`.

talohana:
Resolves #837

### DIFF
--- a/internal/driver/html/graph.css
+++ b/internal/driver/html/graph.css
@@ -1,0 +1,7 @@
+#graph {
+    cursor: grab;
+}
+
+#graph:active {
+    cursor: grabbing;
+}
--- a/internal/driver/html/graph.html
+++ b/internal/driver/html/graph.html
@@ -4,6 +4,7 @@
   <meta charset="utf-8">
   <title>{{.Title}}</title>
   {{template "css" .}}
+  {{template "graph_css" .}}
 </head>
 <body>
   {{template "header" .}}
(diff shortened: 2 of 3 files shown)


internal/driver/html/graph.css
```diff
@@ -0,0 +1,7 @@
+#graph {
```
aalexand, line 1: How about just including this in the existing css file? Not sure a new tiny file is worth having.

talohana: This will make `common.css` not really "common" (one-off style that is applied to a single page). Do you prefer to have this style in the common css file?

aalexand: Makes sense.

internal/driver/html/graph.css
```diff
@@ -0,0 +1,7 @@
...
+    cursor: grab;
+}
+
+#graph:active {
+    cursor: grabbing;
+}
```
aalexand, line 7: Could you add missing newline at the end of this file?

talohana: Done.

aalexand (approved):

PR #866, opened 2024-05-24
Remove pprof->chromedp dependency.

ghemawat:
Moved chromedp based tests to a separate module (in the same Git repository). This change signficantly reduces the dependency footprint of pprof.


browsertests/go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/google/pprof/browsertests
+
+go 1.19
+
+require (
+	github.com/chromedp/chromedp v0.9.2
+	github.com/google/pprof v0.0.0-20240521024322-9665fa269a30
```
liggitt, line 7: you'll want this to exercise the pprof code at HEAD, not this locked version

to do that, point this module at the local pprof module with something like this:

```
replace github.com/google/pprof => ../
```

and probably switch the require directive to `github.com/google/pprof v0.0.0` to make it clear this is not pinning to a specific pprof version

browsertests/browser_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
```
liggitt, line 14: can this be kept inside the internal package, possibly within `internal/driver/browsertests` or something similar?

go.mod
```diff
@@ -3,18 +3,8 @@ module github.com/google/pprof
 go 1.19
 
 require (
-	github.com/chromedp/chromedp v0.9.2
 	github.com/chzyer/readline v1.5.1
 	github.com/ianlancetaylor/demangle v0.0.0-20240312041847-bd984b5ce465
 )
 
-require (
```
liggitt, line 11: this reduction in transitive deps for downstreams is exactly what I was hoping to see 🎉

liggitt:
lgtm, thanks for carving this out

browsertests/browser_test.go
liggitt: > Plus internal/ means something special to Go, so placing a module crossing boundary inside an internal/... subtree may cause some confusion.

It does function properly to refer *up* the tree from an internal/... module (experiment at HEAD of https://github.com/liggitt/pprof/commits/chromedpdep/), and I thought putting this test-only module underneath `internal` would ensure nothing outside pprof could accidentally refer to it, which could reduce confusion.

But if you think it's less confusing as a visible test-only package, that's fine as well.

test.sh
```diff
@@ -29,6 +29,9 @@ PKG=$(go list -f '{{if .TestGoFiles}} {{.ImportPath}} {{end}}' ./...)
 
 go test $PKG
 
+# Browser tests are in a separate module, so do not show in $PKG.
+(cd browsertests && go test)
```
aalexand, line 33: Will we lose code coverage and race detection through browser tests with this change?

Plus `go vet` checks and formatting. And staticcheck above.

browsertests/testutils_test.go
```diff
@@ -0,0 +1,216 @@
+package browser_test
```
aalexand, line 1: Maybe "package browsertests" to match the directory name?

Same for browser_test.go file.

browsertests/testutils_test.go
```diff
@@ -0,0 +1,216 @@
...
+
```
aalexand, line 2: The name of this file testutils_test.go is a bit weird since this file is not a test. I think I'd name it either testutils.go or browsertestsutil.go or something like that.

test.sh
ghemawat: Good point. Added staticcheck, vet, race, lint, fmt.

Did not add coverage since browsertests only tests javascript code for which we cannot currently get any coverage info.

aalexand (approved):

aalexand:
Not merging yet since there is a pending discussion about whether we want to put these tests in internal.

I'm not entirely sure I would like them in internal since I wouldn't call integration tests like this an internal thing. At the same time more top-level directories is not exciting either. One thought I had is maybe we could have top-level `testing` directory and put `browsertests` and `fuzz` directories there.

ghemawat:
I don't like internal either since it does not work across module boundaries. If anything, we could if warranted, put an internal dir under browsertests to hold internal stuff. But seems unnecessary since that package exports nothing.

I like the idea of putting things under a testing directory: testing/fuzz, testing/browser are good names.

We will want a separate PR for moving fuzz anyway. We can move browsertests then, so maybe no reason to hold off on this PR?

PR #868, opened 2024-05-28
css: add font color for inlined function calls

maliyan:
After adding color to the text of the line number, it is easier to recognize which lines are inlined function calls

image

---
image


aalexand:
What if there are two or more function calls on a single line and only one / some of them are inlined?

Please also note that there is a CLA check error.

Issue #823, opened 2023-12-07
Set Function.start_line when symbolizing


prattmic:
Please answer these questions before submitting your issue. Thanks!

### What version of pprof are you using?

If you are using pprof via `go tool pprof`, what's your `go env` output?
If you run pprof from GitHub, what's the Git revision?

a5a03c77bf0842e0592d2be8129bca0b2bd3bb60

### What operating system and processor architecture are you using?

linux-amd64

### What did you do?

If possible, provide a recipe for reproducing the error.
Attaching a profile you are trying to analyze is good.

Analyze a `perf.data` profile, automatically converted via `perf_to_profile`.

```
$ pprof -raw perf.data                                                                                                            
Converting perf.data to a profile.proto... (May take a few minutes)                                                                           
[WARNING:src/quipper/perf_reader.cc:1323] Skipping 120 bytes of metadata: HEADER_CPU_TOPOLOGY                                                 
[WARNING:src/quipper/perf_reader.cc:1070] Skipping unsupported event PERF_RECORD_ID_INDEX                                                     
[WARNING:src/quipper/perf_reader.cc:1070] Skipping unsupported event PERF_RECORD_CPU_MAP                                                      
[WARNING:src/quipper/perf_reader.cc:1070] Skipping unsupported event UNKNOWN_EVENT_82                                                         
[INFO:src/quipper/perf_reader.cc:1061] Number of events stored: 2461                                                                          
[INFO:src/quipper/perf_parser.cc:272] Parser processed: 2 MMAP/MMAP2 events, 2 COMM events, 4 FORK events, 5 EXIT events, 2445 SAMPLE events, 2364 of these were mapped, 0 SAMPLE events with a data address, 0 of these were mapped
[INFO:src/perf_data_handler.cc:103] Using the build id found for the file name: [kernel.kallsyms], build id: ca67921314de38c21c74839716f10a65eba6f6ba.
[INFO:src/perf_data_handler.cc:103] Using the build id found for the file name: [kernel.kallsyms]_text, build id: ca67921314de38c21c74839716f10a65eba6f6ba.
[INFO:src/perf_data_handler.cc:103] Using the build id found for the file name: [kernel.kallsyms]_stext, build id: ca67921314de38c21c74839716f10a65eba6f6ba.
[WARNING:src/perf_data_handler.cc:577] stat: missing_callchain_mmap 2647/11713             
Comment: perf-version:6.5.6                                                                                                                   
[...]

PR #879, opened 2024-07-15
Use llvm-symbolizer's JSON output for symbolizing

LeszekSwirski:
In some edge cases (e.g. injected JIT symbols), function names can have
new lines. This breaks the llvm-symbolizer output parsing, and makes
pprof hang.

Conveniently, as of LLVM 13, llvm-symbolizer has a JSON output mode,
which is robust against all kinds of weirdness like new lines. We can
use this instead of the line-based parsing, and as a bonus we get much
simpler handling of multiple frames in a stack, as the JSON output
already returns these as an array.

This also requires splitting the CODE and DATA processing into separate
functions, since their JSON output is incompatible. For now, we keep the
DATA output as before, a slightly hacky but functional concatenation of
start + size, but this could be improved.

### DIFF
--- a/internal/binutils/addr2liner_llvm.go
+++ b/internal/binutils/addr2liner_llvm.go
@@ -16,6 +16,7 @@ package binutils
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


aalexand (approved):

insilications:
With this merged, I think [#823](https://github.com/google/pprof/issues/823) can be easily solved? The same JSON frame data from `llvm-symbolizer` that is being read can provide `StartLine` for `Function.start_line`.

PR #884, opened 2024-08-17
Use llvm-symbolizer's JSON output to provide function start lines

insilications:
When analyzing a perf.data profile converted automatically via perf_to_profile via `pprof -raw perf.data`, no function start lines (s=0) are present in any of the locations. With https://github.com/google/pprof/commit/813a5fbdbec8a66f7a5aedb876e1b2c3ee0f99ac, this can be easily solved by using the same JSON frame data from `llvm-symbolizer` to provide `StartLine` for `Function.start_line`. This solves https://github.com/google/pprof/issues/823.


aalexand (commented):
Thanks for adding this. Would it be possible to add a test that exercises the new code?

PR #895, opened 2024-09-18
Add Zircon kernel symbolization support.

piwicode:
In the context of Fuchsia Zircon kernel heap memory profiler I need symbolization support.


piwicode:
Hold on. I think I can use the first branch of that function to achieve the same result.

PR #897, opened 2024-09-25
Display hovered box details immediately.

### LINKED ISSUE #836: New flamegraph:  'hover shows details on the top' from old flame-graph missing

github head.

Relveant  in this contetxt Chromium 122.0.6261.69

### What did you do?

Explore the flamegraph in the http UI.

The new flamegraph is nice, but it is missing one feature that was neat in the 'old' version: showing detailed information about a particular node (such long form of demangled name Edit: CPU % is already on the top right).

Currently, this information _is_ there, but only as a hover tooltip, i.e. it takes a second or so to show up.

I like to navigate quickly and hover over interesting nodes and want to see the details immediately, e.g. in the left top corner where it used to be in the 'old' flamegraph. The space on the top is is unused (currently, I don't know if you plan anything there), but it would be nice to have the 'old' feature back (or maybe I already can configure it somewhere and I don't know about it ?)

hover-flamegraph2

ghemawat:
Display the details (name + samples) of the currently hovered box above the flame graph. If no box is currently being hovered over, we display the total samples in the flame graph.

Fixes #836.

image

### DIFF
--- a/internal/driver/html/stacks.css
+++ b/internal/driver/html/stacks.css
@@ -14,12 +14,26 @@ body {
   width: 100%;
   position: relative; /* Allows absolute positioning of child boxes */
 }
-/* Shows details of frame that is under the mouse */
+/* Holder for current frame details. */
 #current-details {
-  position: absolute;
-  top: 5px;
-  right: 5px;
+  position: relative;
+  background: #eee; /* Light grey gives better contrast with boxes */
   font-size: 12pt;
+  padding: 0 4px;
+  width: 100%;
+}
+/* Shows details of frame that is under the mouse */
+#current-details-left {
+  float: left;
+  max-width: 60%;
+  white-space: nowrap;
+  overflow-x: hidden;
+}
+#current-details-right {
+  float: right;
+  max-width: 40%;
+  white-space: nowrap;
+  overflow-x: hidden;
 }
 /* Background of a single flame-graph frame */
 .boxbg {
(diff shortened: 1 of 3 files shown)


aalexand:
I wonder if it would be useful if this new left details box would show the name of the pivot function when in pivot mode and not hovering any other node. This would give an easy way to copy to clipboard the name of a function of interest - pivot, move mouse to details, select, copy.

aalexand (approved):

ghemawat:
> I wonder if it would be useful if this new left details box would show the name of the pivot function when in pivot mode and not hovering any other node

In general, the pivot can be regexp so it can match many functions (useful for example to pivot on std::string to find all string usage). And the pivot is already fairly clear since we have an extra timing display just above the pivot(s). E.g., in the following we have pivoted on "swap":

image

aalexand:
> > I wonder if it would be useful if this new left details box would show the name of the pivot function when in pivot mode and not hovering any other node
> 
> In general, the pivot can be regexp so it can match many functions (useful for example to pivot on std::string to find all string usage). And the pivot is already fairly clear since we have an extra timing display just above the pivot(s). E.g., in the following we have pivoted on "swap":
> 

I was mostly thinking about the general ability to easily copy the function name for a given flame graph node. It's a separate feature perhaps, doesn't have to be discussed now or even prioritized until people ask for it.

hzeller:
Thanks Sanjay!

Issue #339, opened 2018-03-21
vsyscall mapping with remote gperftools profile causes error


toddlipcon:
Please answer these questions before submitting your issue. Thanks!

### What version of pprof are you using?

latest github (a74ae6fb3cd7047c79272e3ea0814b08154a2d3c)

### What operating system and processor architecture are you using?

Ubuntu 16 x64

### What did you do?

Tried to fetch a profile from a gperftools-based profile handler:
pprof http://my-host/pprof/profile?seconds=10

This produced the following:
pprof: unexpected negative adjusted address, mapping &{7 18446744073699065856 18446744073699069952 0 [vsyscall]  false false false false 0 0} source 18446744073699066188, offset 0

This seems to be an issue handling the following vsyscall mapping from /proc/self/maps:

ffffffffff600000-ffffffffff601000 r-xp 00000000 00:00 0           [vsyscall]

If possible, provide a recipe for reproducing the error.
Attaching a profile you are trying to analyze is good.

Attached the profile. It works without symbolization or if I download the binary and pass it on the command line. When I try to use the remote symbolz endpoint, though, it fails as above.
[broken-profile.bin.gz](https://github.com/google/pprof/files/1834944/broken-profile.bin.gz)

aalexand:
> Tried to fetch a profile from a gperftools-based profile handler.

How can I get a simple server with such handler up and running?

toddlipcon:
Unfortunately that's a bit easier said than done, since gperftools doesn't itself provide an HTTP server. I hit this while running against an Apache Kudu server (http://github.com/apache/kudu) but the build takes a while so may not be convenient for you.

In order to repro I wrote a quick little Python HTTP server that responds with the bad profile on any GET requests, and with some junk on any POST request (for symbols). Here's a tarball with a little Python HTTP server:
[test-server.tar.gz](https://github.com/google/pprof/files/1835647/test-server.tar.gz)

You can run it with 'python server.py 12345' and then reproduce the bug:
```
todd@todd-laptop:~/gopath/src/github.com/google/pprof$ ./pprof http://localhost:12345/
Fetching profile over HTTP from http://localhost:12345/
Local symbolization failed for kudu-tserver: stat /opt/cloudera/parcels/CDH-5.15.0-1.cdh5.15.0.p0.205/lib/kudu/sbin-release/kudu-tserver: no such file or directory
Local symbolization failed for libpthread-2.12.so: stat /lib64/libpthread-2.12.so: no such file or directory
[...]

Issue #280, opened 2017-12-08
Remote symbolization fails


ErwanLegrand:
### What version of pprof are you using?

Tested with Git revision 4b386c2. (With a patch to work around #276.)

### What operating system and processor architecture are you using?

Linux AMD64

### What did you do?

I tried to obtain a profile remotely. The server is written in C. Profiles are generated by [gperftools](https://github.com/gperftools/gperftools) 2.1 (for `/pprof/profile`) and [jemalloc](https://github.com/jemalloc/jemalloc) 3.4.1(for `/pprof/heap`). The request handler for `/prof/symbol` handles loader relocations/ASLR perfectly well.

### What did you expect to see?

Collection of profile succeeds, as it does with the old pprof.

### What did you see instead?

```
$ pprof --symbolize=remote https+insecure://XXX/pprof/heap
Fetching profile over HTTP from https+insecure://XXX/pprof/heap
pprof: unexpected negative adjusted address, mapping &{4 0 18446744073709551615 0 https+insecure://XXX/pprof/heap  false false false false 0 0} source 18446744073709551615, offset 0
```

ErwanLegrand:
@aalexand @rauls5382 I have opened this new issue to track the other problem I have mentioned in the comments of #276.

The profiles are in the old pprof format, as this is what libprofiler and libjemalloc generate. (Although I have not checked whether newer releases have switched to the newer format.)

Note that this also fails if `--symbolize=remote` is omitted, even though in this case no request is sent to `/pprof/symbol`.

I believe pprof must perform some computations to account for relocations when profiling locally, because it uses addr2line which does not include support for relocations. But I fail to see why this would be needed in the case of remote debugging. A server written in C can properly perform address to symbol resolution by using Ian Lance Taylor's glorious [libbacktrace](https://github.com/ianlancetaylor/libbacktrace). I would expect the Go runtime package to do the same. (Although I admit I have not actually tried using `Func.FuncForPC` or `Func.FileLine` within an executable built with `-buildmode=pie`.)

aalexand:
As far as I can tell, collecting data using gperftools and viewing that with pprof works fine for ALSR binaries. See example below. Closing the issue, feel free to re-open with more details on how to reproduce.

```
@aalexand:tmp$ cat busyloop.c 

int foo() {
  int sum = 0;
  for (int i = 0; i < 6; i++) {
    for (volatile int j = 0; j < (1 << 30); j++) {
      sum += j;
    }
  }
  return sum;
}

[...]

PR #907, opened 2024-10-22
Allow symbolz to try symbolize unsymbolizable mappings.

aalexand:
We added skipping unsymbolizable mappings in symbolz long time ago in PR #368 to address #339 where we saw error like "unexpected negative adjusted address".

But that error was fixed in a more proper way in subsequent #397 to fix another reported issue #280 (and internal b/32020573). So skipping unsymbolized mappings shouldn't be needed anymore.

I tested this by verifying that the test case from #339 still works fine with the proposed change. And that it does fail if I roll back #397 locally.

This change is useful as we experiment with using symbolz to symbolize JIT locations from //anon (which is unsymbolizable per current terminology).

### DIFF
--- a/internal/symbolz/symbolz.go
+++ b/internal/symbolz/symbolz.go
@@ -37,17 +37,13 @@ var (
 // Symbolize symbolizes profile p by parsing data returned by a symbolz
 // handler. syms receives the symbolz query (hex addresses separated by '+')
 // and returns the symbolz output in a string. If force is false, it will only
-// symbolize locations from mappings not already marked as HasFunctions. Never
-// attempts symbolization of addresses from unsymbolizable system
-// mappings as those may look negative - e.g. "[vsyscall]".
+// symbolize locations from mappings not already marked as HasFunctions. Does
+// not skip unsymbolizable files since the symbolz handler can be flexible
+// enough to handle some of those cases such as JIT locations in //anon.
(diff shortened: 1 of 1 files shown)


internal/symbolz/symbolz.go
```diff
@@ -37,19 +37,15 @@ var (
 // Symbolize symbolizes profile p by parsing data returned by a symbolz
 // handler. syms receives the symbolz query (hex addresses separated by '+')
 // and returns the symbolz output in a string. If force is false, it will only
-// symbolize locations from mappings not already marked as HasFunctions. Never
-// attempts symbolization of addresses from unsymbolizable system
-// mappings as those may look negative - e.g. "[vsyscall]".
+// symbolize locations from mappings not already marked as HasFunctions. Does
+// not skip unsymbolizable files since the symbolz handler can be flexible
```
gmarin13, line 41: I feel this comment may feel out of context after the patch is submitted and the check on m.Unsymbolizable() is not there anymore. Most people won't think "why aren't we skipping unsymbolizable mappings?".
Maybe just skip the comment here?

Optionally, perhaps the package comments could have a bit more details about symbolz and how it can get symbols for JITed code.

gmarin13: I don't have a strong opinion about it. We can submit as is.

PR #912, opened 2024-11-21
Fix 'show source [in new tab]' under -lines

tbg:
With this patch, it's possible to navigate from the (new and now only) flame graph to the source code for a box.


aalexand (commented):
Could you add a comment that explains the "why" behind the regexp? (no need to explain the regexp mechanics itself)

tbg:
Sure, done!

aalexand (approved):

tbg:
Thanks @aalexand - could you merge?

Issue #477, opened 2019-07-18
OS X symbolization still does not work


ezyang:
### What version of pprof are you using?

fe62a0baab87ba3abca12f4a621532bf67c9a7d2

### What operating system and processor architecture are you using?

OS X Mojave

### What did you do?

```
$ which addr2line
/usr/local/bin/addr2line
$ cat test.cpp
#include <vector>
#include <iostream>

int main() {
  std::vector<int> x(10000000, 2);
  for (int i = 0; i < 100000; i++) {
    x[i] *= x[(i+200) % 1000];
  }
  x.resize(1000000000);
  std::cerr << x[123] << "\n";
}

$ clang++ test.cpp -g3 -fno-pie -ltcmalloc
$ HEAPPROFILE=/tmp/hh ./a.out
Starting tracking the heap
Dumping heap profile to /tmp/hh.0001.heap (3852 MB allocated cumulatively, 3852 MB currently in use)
4
Dumping heap profile to /tmp/hh.0002.heap (Exiting, 0 bytes in use)
$ pprof a.out /tmp/hh.0001.heap
Using local file a.out.
Using local file /tmp/hh.0001.heap.
Welcome to pprof!  For help, type 'help'.
(pprof) top
Total: 3852.8 MB
  3852.8 100.0% 100.0%   3852.8 100.0% 0x00007fff64e72c05
```

### What did you expect to see?

Expected address to be symbolized

### What did you see instead?

Not symbolized

kalyanac:
What is the version of addr2line / binutils?
If it is less than 2.21, addr2line doesnt support the "-a" flag.

HDHarry:
I can second this. Does not work for me using cpu profiler.

ezyang:
addr2line 2.32

Excerpts from Kalyana Chadalavada's message of 2019-07-18 04:24:34 -0700:
> What is the version of addr2line / binutils?
> If it is less than 2.21, addr2line doesnt support the "-a" flag. 
>

aalexand:
We'll do our best to look at this some time in the future, but not sure how soon this is going to be. Your best bet is to debug this a bit yourself, as you have the source code.

ezyang:
My real plan is to just stop trying to run this on OS X :)

Excerpts from Alexey Alexandrov's message of 2019-07-19 15:48:03 -0700:
> We'll do our best to look at this some time in the future, but not sure how soon this is going to be. Your best bet is to debug this a bit yourself, as you have the source code.
>

[...]

PR #924, opened 2025-01-23
handle demangling OSX names with underscore prefix

alk:
See discussion at https://github.com/google/pprof/issues/477#issuecomment-2600988101

### DIFF
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -292,6 +292,16 @@ func demangleSingleFunction(fn *profile.Function, options []demangle.Option) {
 		fn.Name = demangled
 		return
 	}
+
+	// OSX has all the symbols prefixed with extra '_' so lets try
+	// once more without it
+	if strings.HasPrefix(fn.SystemName, "_") {
+		if demangled := demangle.Filter(fn.SystemName[1:], o...); demangled != fn.SystemName {
+			fn.Name = demangled
+			return
+		}
+	}
+
 	// Could not demangle. Apply heuristics in case the name is
 	// already demangled.
 	name := fn.SystemName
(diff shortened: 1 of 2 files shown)


internal/symbolizer/symbolizer.go
```diff
@@ -292,6 +292,16 @@ func demangleSingleFunction(fn *profile.Function, options []demangle.Option) {
 		fn.Name = demangled
 		return
 	}
+
+	// OSX has all the symbols prefixed with extra '_' so lets try
+	// once more without it
+	if strings.HasPrefix(fn.SystemName, "_") {
+		if demangled := demangle.Filter(fn.SystemName[1:], o...); demangled != fn.SystemName {
```
aalexand, line 299: Could we just pass `strings.StripPrefix(fn.SystemName, "_"), ...` in the `demangle.Filter` call above?

alk: Above? So just cut all the time no matter OS and everything? Then it won't work, since _ is usually used for mangled names for regular OS-es. And cutting it will simply break everything.

aalexand: Oh, got it.

Can we add a test that will exercise this code path?

alk: Ack. Added test case.

aalexand (approved):

alk:
The failure looks like a flake. The failed test looks unrelated to my change. But do let me know if I am mistaken. And if there is anything I can do to help.

aalexand:
> The failure looks like a flake. The failed test looks unrelated to my change. But do let me know if I am mistaken. And if there is anything I can do to help.

Right, it's a known recent flaky error, I filed #925 to track it. I restarted the failed test.

alk:
Thanks. Test passed now. So what happens next ?

aalexand:
merged

PR #929, opened 2025-02-08
In the tags command, print percentages relative to profile total.

aalexand:
Before this change percentages were relative to the tag total and this sometimes causes confusion for users as it makes it look like the tag is set for all samples when it's not.

Note that the percentage calculation as implemented does not attempt to be friendly to samples where a given tag has multiple values. Multiple values per tag per sample are discouraged and we should deprecate them. Maybe we should make pprof print a warning for such profiles.


ghemawat (approved):
Looks good.

Issue #835, opened 2024-03-04
Canvas size related to initial size of a graph, should use the full space


hzeller:
Please answer these questions before submitting your issue. Thanks!

### What version of pprof are you using?

github head.

### What operating system and processor architecture are you using?
x86_64
Relevant for this issue: Webbrowser Chromium 122.0.6261.69 

### What did you do?
Running pprof with --http option to get the web UI

The UI opens with the call graph, which looks nice and appears to be using the full available browser window as canvas:

![a](https://github.com/google/pprof/assets/140937/30a5a565-3b2a-4fa9-9f92-75d83b457cd7)

... however, when zooming, it is revealed that the canvas was, in fact, sized after the original aspect ratio of the overview graph. You see in this zoomed-in view of the graph, that the bottom is exactly cut at the border of the original overview canvas:

![b](https://github.com/google/pprof/assets/140937/3c708136-ee2e-4d37-8534-326dd5859675)

### What I would like to see

The canvas should fill the available space in the window, even if it is not fully needed in the vertical range initially ... after zooming it certainly is.

aalexand:
@hzeller I think it would be in internal/driver/webui.go and internal/driver/html, I can't say more concretely. It would be great if you can look into this and fix it!

hzeller:
This fixes it for me. Changelist or Pull-Request preferred ?

```patch
--- a/internal/driver/html/common.css
+++ b/internal/driver/html/common.css
@@ -218,10 +218,12 @@ a {
 }
 #graph {
   overflow: hidden;
+  width: 100%;
+  height: 100%;
 }
 #graph svg {
   width: 100%;
-  height: auto;
+  height: 100%;
   padding: 10px;
 }
 #content.source .filename {
```

aalexand:
Great! A pull request please. Please also include some sort of before/after pictures that will convince the reviewer the fix is working. Thanks!

PR #937, opened 2025-04-10
Fix canvas to fill the full height.

hzeller:
The canvas used to scale itself to the aspect ratio of the full graph, which might be fairly wide. That means even while zooming, the original aspect ratio is preserved and only a 'letterbox' is shown.

This CSS change fixes that.

Fixes: #835

### DIFF
--- a/internal/driver/html/common.css
+++ b/internal/driver/html/common.css
@@ -218,10 +218,12 @@ a {
 }
 #graph {
   overflow: hidden;
+  width: 100%;
+  height: 100%;
 }
 #graph svg {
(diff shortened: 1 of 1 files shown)


hzeller:
This shows a zoomed-in view of a graph that started out fairly wide and even after zooming occipies the same aspect ratio before, but uses the full available screenwidth afterwards:

Before:

pprof-before

After:

pprof-after

hzeller:
The Linux tests look like they fail due to some unrelated reason.

aalexand (approved):

PR #939, opened 2025-04-17
Fix the OSX demangling fallback comparison.

aalexand:
This is a follow-up to #924. In the comparison we should compare to the string with stripped underscore, otherwise the comparison will never be true and we strip the underscore in cases we shouldn't. See the added test case.

Also, while we are here, add copying demangling options that we also missed. Hopefully we can get
https://github.com/ianlancetaylor/demangle/pull/24 in and this won't be needed.

### DIFF
--- a/internal/symbolizer/symbolizer.go
+++ b/internal/symbolizer/symbolizer.go
@@ -296,7 +296,10 @@ func demangleSingleFunction(fn *profile.Function, options []demangle.Option) {
 	// OSX has all the symbols prefixed with extra '_' so lets try
 	// once more without it
 	if strings.HasPrefix(fn.SystemName, "_") {
-		if demangled := demangle.Filter(fn.SystemName[1:], o...); demangled != fn.SystemName {
+		// Copy the options because they may be updated by the call.
+		o := make([]demangle.Option, len(options))
+		copy(o, options)
+		if demangled := demangle.Filter(fn.SystemName[1:], o...); demangled != fn.SystemName[1:] {
 			fn.Name = demangled
 			return
 		}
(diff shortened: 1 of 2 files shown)


aalexand:
@alk @ianlancetaylor FYI

gmarin13 (approved):

alk:
Thanks for catching this. I apologize for the trouble.

alk:
So this "broke" gperftools tools tests on OSX. Basically there are some unmangled symbols that we expected to be hidden when processing legacy profiles. In this case it was tc_newarray. Which of course on OSX is _tc_newarray.

So this "side effect" of removing '_' that was removed in this fix, we kinda needed it. For now I am manually passing --hide= flag for fix the test.

But looks like the underlying issue is that we really should un-tangle underscore prefix removal on OSX from demangling.

Or most ideally when extracting raw symbols whatever OSX-specific thing we do (well we don't, we just invoke usual set of tools; some of those are actually already automagically strip this stuff).

Curious to see people's opinion on this.

ianlancetaylor:
If I'm reading the code correctly, it tries to demangle the symbol, and if the demangling fails and the symbol starts with an underscore, it tries demangling the symbols without the underscore.

In my opinion that is a bad approach. Whether symbols have a leading underscore or not is determined by the platform&mdash;the combination of `runtime.GOARCH` and `runtime.GOOS`. The code should have a list of platforms that have a leading underscore. On that platforms, a leading underscore should be removed. On other platforms, it should not. Either way, the demangler should only be called once.

For example, see https://go.googlesource.com/go/+/3432c68467d50ffc622fed230a37cd401d82d4bf/src/cmd/link/internal/ld/ar.go#227 for the (rather simple) code that strips a leading underscore for targets supported by Go.

If there are cases where pprof doesn't know the appropriate `runtime.GOOS` and `runtime.GOARCH` to use, because the profile comes from a different system, then a pprof option can be added saying whether to strip underscores or not. As far as I know not many people would have to understand that option, as most people run pprof on profiles generated on the same system.

PR #943, opened 2025-04-28
Enable net/http/pprof handlers in pprof web UI.

aalexand:
This is so that pprof itself can be profiled when needed.

Tested: started the web UI, collected /debug/pprof/heap and /debug/pprof/profile profiles.


internal/driver/webui.go
```diff
@@ -210,6 +210,9 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
 	mux := http.NewServeMux()
 	mux.Handle("/ui/", http.StripPrefix("/ui", handler))
 	mux.Handle("/", redirectWithQuery("/ui", http.StatusTemporaryRedirect))
+	// Enable self-profiling handlers if net/http/pprof is imported by the main
+	// program.
```
ghemawat, line 214: Are there going to be users who are unhappy about having this access point opened?

internal/driver/webui.go
```diff
@@ -210,6 +210,9 @@ func defaultWebServer(args *plugin.HTTPServerArgs) error {
...
+	mux.HandleFunc("/debug/pprof/", http.DefaultServeMux.ServeHTTP)
```
ghemawat, line 215: Should we apply the same auth checks (localhost etc.) that we apply to other handlers?

pprof.go
```diff
@@ -24,6 +24,8 @@ import (
 
 	"github.com/chzyer/readline"
 	"github.com/google/pprof/driver"
+
+	_ "net/http/pprof" // Enable profiling pprof itself in the web UI mode.
```
ghemawat, line 28: In the past I have wanted to profile startup performance itself. This approach does not allow profiling startup time. An alternative would be flags (separate ones will be needed for cpu and memory profiling).

Not sure how we exit cleanly though to allow the profile to be saved if we are using flags.

PR #948, opened 2025-06-29
Treat memfd files as unsymbolizable.

aalexand:


profile/profile_test.go
```diff
@@ -1940,3 +1940,71 @@ func BenchmarkWrite(b *testing.B) {
...
+			wantUnsymbolizable: true,
+		},
+		{
+			desc:               "vsyscall mapping is unsymbolizable",
+			file:               "[vsyscall]",
+			wantUnsymbolizable: true,
+		},
+		{
+			desc:               "kernel mapping is unsymbolizable",
+			file:               "[kernel.kallsyms]_text",
```
gmarin13, line 1967: This is pre-existing behavior, but why are we treating the kernel mappings as unsymbolizable?
We should be able to symbolize them if we have the debug symbols for the kernel.

aalexand: Good question. I think we only recognize vmlinux name [here](https://github.com/google/pprof/blob/e611ec304b22ed2bc39c0fd534f71243962cdd8a/internal/binutils/binutils.go#L444). I think for the local symbolization of the kernel we expect that user configures `$buildid/$name` kind of search dir which renames the mapping name to the name, vmlinux in case of the kernel. See b/168662041#comment2. But this is definitely an interesting case.

gmarin13: I see. Actually, it looks like the name needs to include substring `vmlinux`, not be exactly `vmlinux`.

gmarin13 (approved):

profile/profile_test.go
aalexand: > Actually, it looks like the name needs to include substring `vmlinux`, not be exactly `vmlinux`.

Ah, right - I don't remember why...

PR #951, opened 2025-08-10
profile: optimize Parse allocs

pymq:
`go test -bench='^\QBenchmarkParse\E$' -run='^$' -count=10 -benchtime=5s`

```
goos: linux
goarch: amd64
pkg: github.com/google/pprof/profile
cpu: 13th Gen Intel(R) Core(TM) i7-1360P
         │ old-parse.txt │            new-parse.txt             │
         │    sec/op     │    sec/op     vs base                │
Parse-16    62.07m ± 13%   55.54m ± 13%  -10.52% (p=0.035 n=10)

         │ old-parse.txt │            new-parse.txt             │
         │     B/op      │     B/op      vs base                │
Parse-16    47.56Mi ± 0%   41.09Mi ± 0%  -13.59% (p=0.000 n=10)

         │ old-parse.txt │            new-parse.txt            │
         │   allocs/op   │  allocs/op   vs base                │
Parse-16     272.9k ± 0%   175.8k ± 0%  -35.58% (p=0.000 n=10)
```

### DIFF
--- a/profile/proto.go
+++ b/profile/proto.go
@@ -36,6 +36,7 @@ package profile
 import (
 	"errors"
 	"fmt"
+	"slices"
 )
 
 type buffer struct {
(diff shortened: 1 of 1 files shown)


profile/proto.go
```diff
@@ -187,6 +188,19 @@ func le32(p []byte) uint32 {
 	return uint32(p[0]) | uint32(p[1])<<8 | uint32(p[2])<<16 | uint32(p[3])<<24
 }
 
+func decodeVarintsCount(data []byte) (n int) {
```
aalexand, line 191: nit: Can this be written a bit shorter as

```go
func peekNumVarints(data []byte) int {
	for numVarints := 0; len(data) > 0; numVarints++ {
		var err error
		if _, data, err := decodeVarint(data); err != nil {
			break
		}
	}
	return numVarints
}
```

PR #976, opened 2025-12-06
Add -all_frames flag that allows disabling the drop / keep frames logic

aalexand:
This is occasionally useful for debugging and to inspect raw profile stacks.

### DIFF
--- a/internal/driver/cli.go
+++ b/internal/driver/cli.go
@@ -37,6 +37,7 @@ type source struct {
 	HTTPHostport       string
 	HTTPDisableBrowser bool
 	Comment            string
+	AllFrames          bool
 }
 
 // parseFlags parses the command lines through the specified flags package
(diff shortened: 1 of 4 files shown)


internal/driver/cli.go
```diff
@@ -51,7 +52,8 @@ func parseFlags(o *plugin.Options) (*source, []string, error) {
 	flagSymbolize := flag.String("symbolize", "", "Options for profile symbolization")
 	flagBuildID := flag.String("buildid", "", "Override build id for first mapping")
 	flagTimeout := flag.Int("timeout", -1, "Timeout in seconds for fetching a profile")
-	flagAddComment := flag.String("add_comment", "", "Annotation string to record in the profile")
+	flagAddComment := flag.String("add_comment", "", "Free-form annotation to add to the profile")
+	flagNoDropFrames := flag.Bool("no_drop_frames", false, "Ignore drop_frames directive in the profile (rare, mainly for debugging)")
```
ghemawat: Since this affects both drop-frames and keep-frames, maybe we can have a better name. How about --all_frames with comment

"Keep all frames in profile, ignoring drop_frames and keep_frames regular expressions (rarely needed, mainly used for debugging pprof itself)"

But I am fine with your current naming if you prefer it.

ghemawat (dismissed):

internal/driver/cli.go
aalexand: Makes sense, done.

ghemawat (approved):

PR #977, opened 2025-12-12
internal/report: use better heuristic for locating the source code

valyala:
The current heuristic for locating the source code for profile functions doesn't work in the following cases when running pprof with default options:

1. It cannot find source code for standard Go packages. Users need to specify -source_path=`go env GOROOT` as a workaround.

2. It cannot find source code for external modules. Users need to specify -source_path=`go env GOMODCACHE` as a workaround. But this workaround stops working if external module names contain uppercase chars. For example, github.com/FooBar/Baz, because Go stores sources for such module names into $GOMODCACHE/github.com/!foo!bar/!baz directory according to https://go.dev/ref/mod#module-cache , while pprof is unaware of such a conversion.

3. It cannot find source code at the vendor directory. Users need to specify -source_path=`pwd`/vendor as a workaround. But this workaround doesn't work for vendored modules, since their names contain module version in the form module_name@version, while the source code for the module_name is stored inside vendor/module_name directory.

4. It cannot find source code for profiles obtained from Go executables built with -trimpath command-line flag, if their sources are put in the directory with the name other than the module name.

This significantly reduces pprof usability, since it is very hard to inspect sources with `list` command. Users have to create horrible hacks with -source_path and -trim_path command-line flags such as https://github.com/VictoriaMetrics/VictoriaLogs/pull/893 , but these hacks do not cover properly all the cases mentioned above.

This commit covers all the cases mentioned above, so pprof properly detects source code with default settings, without the need to specify -trim_path and -source_path command-line flags.


jespino:
@aalexand Hi, is there anything that I can do to help to get this in? It would help improve the experience in some go projects.

aalexand:
@jespino @valyala One thing that would help is to create an issue with the description from the PR and then land it in a few smaller independent changes. The current PR is not huge but is not easy to untangle in terms of why this or that part of code was changed and how this maps to the four things being addressed.

PR #987, opened 2026-01-15
Support multiple -add_comment arguments

camillobruni:
This makes it a bit nicer to annotate profiles when using external tools.


aalexand (commented):
I think the test implementation testFlags is different from internal/driver/flags.go. I'm not sure GoFlags there support multiple flags properly. Could you test that manually? Please also check that this change does not make the comma a separator as we want something like `-add_comment=hello,world` to add one comment, not two.

PR #1008, opened 2026-07-15
internal/graph: escape double quotes in tag names in DOT output

### LINKED ISSUE #769: quotes in CPU profile tags not ecsaped properly

00490a63f31712a6991b73391dd3decdada278d0 (i.e. main as of today), but the problem exists in all prior versions. I first stumbled upon it in Go 1.19.

Irrelevant

### What did you do?

Run `pprof -dot` on the provided file (this is a CPU profile taken of a CockroachDB 23.1 node, but these details don't matter). Note how the resulting `.dot` file is invalid since the `"` occurring in a label was not properly escaped:

```
N36_0 [label = "range_str:12419/2:/Table/136/1/"{NHCH-…-PWN-a"}" id="N36_0" fontsize=8 shape=box3d tooltip="0.01s"]
```

[pprof.pb.gz.txt](https://github.com/google/pprof/files/11230678/pprof.pb.gz.txt)
[pprof.dot.txt](https://github.com/google/pprof/files/11230679/pprof.dot.txt)

The `"` in the label should have been escaped, resulting in a valid `.dot` file.

### What did you see instead?

An invalid `dot` file, which in particular implies that the graph view in pprof isn't working. The line above should have been

```
N36_0 [label = "range_str:12419/2:/Table/136/1/\"{NHCH-…-PWN-a\"}" id="N36_0" fontsize=8 shape=box3d tooltip="0.01s"]
```

which [would work](https://dreampuf.github.io/GraphvizOnline/#digraph%20G%20%7B%0A%20%20start%20-%3E%20a0%20%5Blabel%3D%22range_str%3A12419%2F2%3A%2FTable%2F136%2F1%2F%5C%22%7BNHCH-%E2%80%A6-PWN-a%5C%22%7D%22%5D%3B%0A%7D).

I gave it a half-hearted attempt at fixing, but it seemed like there might be more than one place where the escaping isn't happening properly and it wasn't super clear to me where we'd need to throw in an `escapeForDot`[^1]  on top of here:

https://github.com/google/pprof/blob/a41b82acbcb1f6a6d0998b0fddeb1f1adb7d8fc1/internal/graph/dotgraph.go#L251

Also, it seems useful to add testing for these sorts of things.

[^1]: https://github.com/google/pprof/blob/a41b82acbcb1f6a6d0998b0fddeb1f1adb7d8fc1/internal/graph/dotgraph.go#L490-L495

LeSingh1:
Fixes #769

## Problem

Tag (nodelet) labels in the DOT renderer interpolate the tag name directly into the `label = "..."` attribute without escaping:

```go
nodelets += fmt.Sprintf(`N%d_%d [label = "%s" ...]`, nodeID, i, t.Name, ...)
```

Every other label in `dotgraph.go` (node names, edge labels, legend title) is passed through `escapeForDot`, but the two tag-nodelet paths were not. A tag name containing a double quote — common in CPU-profile string tags, e.g. `range_str:.../"{...}"` — produces invalid DOT that graphviz cannot render, breaking the whole graph (see the example in #769).

## Fix

Apply the existing `escapeForDot` helper to `t.Name` in both the label-tag path (`tagsToNodelets`) and the numeric-nodelet path (`numericNodelets`). No other behavior changes.

## Test

Added `TestComposeWithTagsThatNeedEscaping` (golden `compose10.dot`): a label tag whose name contains `"` now renders as `label = "range_str:\"{quote}\""`. The test **fails before** this change (unescaped quote → malformed DOT) and **passes after**. Full `internal/graph` package tests pass; `go vet` and `gofmt` are clean.

### DIFF
--- a/internal/graph/dotgraph.go
+++ b/internal/graph/dotgraph.go
@@ -247,7 +247,7 @@ func (b *builder) addNodelets(node *Node, nodeID int) bool {
 			continue
 		}
 		weight := b.config.FormatValue(w)
-		nodelets += fmt.Sprintf(`N%d_%d [label = "%s" id="N%d_%d" fontsize=8 shape=box3d tooltip="%s"]`+"\n", nodeID, i, t.Name, nodeID, i, weight)
+		nodelets += fmt.Sprintf(`N%d_%d [label = "%s" id="N%d_%d" fontsize=8 shape=box3d tooltip="%s"]`+"\n", nodeID, i, escapeForDot(t.Name), nodeID, i, weight)
 		nodelets += fmt.Sprintf(`N%d -> N%d_%d [label=" %s" weight=100 tooltip="%s" labeltooltip="%s"]`+"\n", nodeID, nodeID, i, weight, weight, weight)
 		if nts := lnts[t.Name]; nts != nil {
 			nodelets += b.numericNodelets(nts, maxNodelets, flatTags, fmt.Sprintf(`N%d_%d`, nodeID, i))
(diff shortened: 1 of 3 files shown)


aalexand:
It doesn't look like the proto "officially" disallows double-quote characters in the label/tag names, but it's pretty odd to have them there. What is the use case for this?

Issue #1009, opened 2026-07-15
pprof cannot locate source code for the list command in common Go setups


jespino:
The current heuristic for locating the source code for profile functions doesn't work in the following cases when running pprof with default options:

- It cannot find source code for standard Go packages. Users need to specify -source_path=`go env GOROOT`  as a workaround.
- It cannot find source code for external modules. Users need to specify -source_path=`go env GOMODCACHE`  as a workaround. But this workaround stops working if external module names contain uppercase chars. For example, github.com/FooBar/Baz, because Go stores sources for such module names into $GOMODCACHE/github.com/!foo!bar/!baz directory according to https://go.dev/ref/mod#module-cache, while pprof is unaware of such a conversion.
- It cannot find source code at the vendor directory. Users need to specify -source_path=`pwd`/vendor  as a workaround. But this workaround doesn't work for vendored modules, since their names contain module version in the form module_name@version, while the source code for the module_name is stored inside vendor/module_name directory.
- It cannot find source code for profiles obtained from Go executables built with -trimpath command-line flag, if their sources are put in the directory with the name other than the module name.

This significantly reduces pprof usability, since it is very hard to inspect sources with list command. Users have to create horrible hacks with -source_path and -trim_path command-line flags such as VictoriaMetrics/VictoriaLogs#893, but these hacks do not cover properly all the cases mentioned above.

It would be great if pprof could properly detect source code in all the cases mentioned above with default settings, without the need to specify -trim_path and -source_path command-line flags.

jespino:
This has a PR trying to fix this issue in #977 but I'll split it in multiple smaller PRs with very focused fixes.

PR #1010, opened 2026-07-20
internal/report: fall back to $GOROOT/src when locating source files

jespino:
Profiles from Go programs refer to standard library sources under the GOROOT of the build machine, which breaks the source and weblist views when that path does not exist locally (e.g. a profile from another machine), or when the program was built with -trimpath and the files are recorded relative to $GOROOT/src.

Teach openSourceFile to fall back to the local $GOROOT/src after the regular search fails: relative paths resolve against it directly, and absolute paths resolve by matching the components after a /src/ one. The fallback only runs when the existing lookup finds nothing, so previously resolved files are unaffected.

Also introduce tryOpenFile, which rejects directories that os.Open would happily open, so a directory matching a candidate path no longer wins the search and then fails when its lines are read.

This is the first PR of 4 to address the #1009

### DIFF
--- a/internal/report/source.go
+++ b/internal/report/source.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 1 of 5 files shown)


internal/report/source.go
```diff
@@ -1007,28 +1008,95 @@ func openSourceFile(path, searchPath, trim string) (*os.File, error) {
...
-		// Search up for every parent of each possible path.
-		for {
-			filename := filepath.Join(dir, path)
-			if f, err := os.Open(filename); err == nil {
-				return f, nil
-			}
-			parent := filepath.Dir(dir)
-			if parent == dir {
-				break
+		if f, err := tryOpenFile(path); err == nil {
```
jespino, line 1011: All this changes are round allowing to fallback later to other "handlers", for now we are adding the goroot handler, but in other PRs we will add more special handlers like gomodcache or vendor.

internal/report/source.go
```diff
@@ -1007,28 +1008,95 @@ func openSourceFile(path, searchPath, trim string) (*os.File, error) {
...
+	if f, err := openGorootSourceFile(path, gorootSrc); err == nil {
+		return f, nil
+	}
 
 	return nil, fmt.Errorf("could not find file %s on path %s", path, searchPath)
 }
 
+// tryOpenFile opens the file at filename if it exists and is not a directory,
+// which os.Open would also happily open.
+func tryOpenFile(filename string) (*os.File, error) {
```
jespino, line 1043: Fail fast when you try to open a directory.

internal/report/source.go
```diff
@@ -1007,28 +1008,95 @@ func openSourceFile(path, searchPath, trim string) (*os.File, error) {
...
+}()
+
+// openGorootSourceFile tries to open a Go standard library source file under
+// gorootSrc. Binaries built without -trimpath record standard library files
+// under the build machine's GOROOT (e.g. /usr/local/go/src/runtime/proc.go),
+// which may not exist locally, so resolve the path components after a "/src/"
+// component against gorootSrc. Binaries built with -trimpath record them
+// relative to $GOROOT/src (e.g. runtime/proc.go), so resolve relative paths
+// against gorootSrc directly.
+func openGorootSourceFile(path, gorootSrc string) (*os.File, error) {
```
jespino, line 1078: this is the special handling for goroot, only executed if the openSourceFile previous machinery doesn't find anything.

internal/report/source.go
```diff
@@ -1007,28 +1008,95 @@ func openSourceFile(path, searchPath, trim string) (*os.File, error) {
...
+	}
+	return f, nil
+}
+
+// gorootSrc is the directory holding the Go standard library sources, if
+// known. runtime.GOROOT honors the $GOROOT environment variable and otherwise
+// reports the GOROOT this binary was built with, which is the Go installation
+// used to `go install` pprof in the common case.
+var gorootSrc = func() string {
+	if goroot := runtime.GOROOT(); goroot != "" {
```
aalexand, line 1065: I think pprof binary's `runtime.GOROOT()` can be different from the goroot of the profiled binary? Not sure we can guarantee that this is the right default.

internal/report/source.go
aalexand: Did you see cases where this practically happens? Asking because it seems unlikely that a source file path matches a directory name because source file names usually have an extension and directory names do not.

internal/report/source.go
```diff
@@ -1007,28 +1008,95 @@ func openSourceFile(path, searchPath, trim string) (*os.File, error) {
...
+	} else {
+		// Scan each component of the path.
+		for _, dir := range filepath.SplitList(searchPath) {
+			// Search up for every parent of each possible path.
+			for {
+				filename := filepath.Join(dir, path)
+				if f, err := tryOpenFile(filename); err == nil {
+					return f, nil
+				}
+				parent := filepath.Dir(dir)
```
aalexand, line 1023: Note to self: This is historical but I think we should remove using even parent path of the search paths. When I specify /foo/bar as the search path, I do not expect that the tool will attempt /foo/src/baz.cc paths or something like that. In particular, /foo/src can be a slow NFS / remote mounted directory and trying to look up those paths will slow down the scan.

internal/report/source.go
```diff
@@ -1007,28 +1008,95 @@ func openSourceFile(path, searchPath, trim string) (*os.File, error) {
...
+var gorootSrc = func() string {
+	if goroot := runtime.GOROOT(); goroot != "" {
+		return filepath.Join(goroot, "src")
+	}
+	return ""
+}()
+
+// openGorootSourceFile tries to open a Go standard library source file under
+// gorootSrc. Binaries built without -trimpath record standard library files
+// under the build machine's GOROOT (e.g. /usr/local/go/src/runtime/proc.go),
```
aalexand, line 1073: Is there a possibility we could instead 1) modify `trimPath()` below to trim the absolute path prefix; 2) add $GOROOT/src to the default search path? I assume that would be enough to locate the runtime files.

aalexand:
Please sign the CLA.

internal/report/source.go
jespino: the runtime.GOROOT() is used as the "fallback" in case the original reported file in the pprof hasn't been found. It is not "ideal", but is a reasonable fallback.

internal/report/source.go
jespino: Fair enough. it was more a defensive measure imported directly from the pre-existing PR, I found it interesting, but we can remove it, from this PR, and if it is needed, we can add it back in the future.

PR #1014, opened 2026-08-02
internal/report: guard against nil Symbol regexp in disasm and source rendering

### LINKED ISSUE #789: disasm: no matches found for regexp
pprof latest, go tip

Ubuntu 22.04.2, amd64

### What did you do?

Tried to open a pprof profile capture on my mac (darwin/arm64) on my linux/amd64 machine.

With `go tool pprof` on Go 1.17 I see:

image

### What did you see instead?

Starting with go 1.18 the disasm page crashes:

image

Same for google/pprof latest

On console it says: `no matches found for regexp:`

anishesg:
The disasm page crashes when accessed without a filter parameter because the Symbol regexp is nil. The code in PrintAssembly (report.go:400) and printSource (source.go:52) attempted to dereference the nil pointer when calling o.Symbol.String() or o.Symbol.MatchString().

This fix adds nil checks before accessing the Symbol regexp in three locations: PrintAssembly, printSource, and MakeWebList. When Symbol is nil, the code now treats it as matching all symbols (for MatchString) or uses an empty string in error messages (for String()).

The issue manifests when viewing profiles from different architectures (e.g., darwin/arm64 on linux/amd64) because the web UI calls the disasm endpoint without a symbol filter, causing a nil pointer dereference that displays as "no matches found for regexp:" with an empty regexp value.

Fixes #789

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -397,8 +397,10 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	// If the regexp source can be parsed as an address, also match
 	// functions that land on that address.
 	var address *uint64
-	if hex, err := strconv.ParseUint(o.Symbol.String(), 0, 64); err == nil {
-		address = &hex
+	if o.Symbol != nil {
+		if hex, err := strconv.ParseUint(o.Symbol.String(), 0, 64); err == nil {
+			address = &hex
+		}
 	}
 
 	fmt.Fprintln(w, "Total:", rpt.formatValue(rpt.total))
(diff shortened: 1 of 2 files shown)


aalexand:
Do you mind attaching before/after screenshots or outputs from the tool to demonstrate the effect of the fix?

Also a question on this:

> The issue manifests when viewing profiles from different architectures (e.g., darwin/arm64 on linux/amd64) because the web UI calls the disasm endpoint without a symbol filter,

Could you elaborate in the description why different architectures affect this?

internal/report/report.go
```diff
@@ -397,8 +397,10 @@ func PrintAssembly(w io.Writer, rpt *Report, obj plugin.ObjTool, maxFuncs int) e
 	// If the regexp source can be parsed as an address, also match
 	// functions that land on that address.
 	var address *uint64
-	if hex, err := strconv.ParseUint(o.Symbol.String(), 0, 64); err == nil {
-		address = &hex
+	if o.Symbol != nil {
```
aalexand, line 400: I don't think `o.Symbol` is nil in the described scenario, I think it points to a regexp for empty string.

If `o.Symbol` was nil, `o.Symbol.String()` would panic.

I don't think this PR fixes anything, so I'll close it.

PR #1022, opened 2026-08-29
internal/driver: close saved profile files

cuishuang:
Remote profiles are saved to an `*os.File` returned by `newTempFile`, but the file was not closed after `Profile.Write` returned.

This is mostly hidden for short-lived CLI invocations, but leaks file descriptors when pprof is used repeatedly from a long-lived process.

Close the saved profile file after writing, treat close failures as save failures, and only print the success message when both writing and closing succeed. Preserve the write error if both operations fail.

Add tests covering successful writes, write failures, close failures, and combined write and close failures.

Tests:
- `./test.sh`


internal/driver/fetch_test.go
```diff
@@ -42,6 +42,53 @@ import (
 	"github.com/google/pprof/profile"
 )
 
+type testWriteCloser struct {
```
aalexand: Not sure if this test is worth it. It's pretty shallow in what it actually tests and the LOC of the test for the triviality of the logic it tests seems questionable.

cuishuang: Thanks for the review. I agree that the test was shallow and disproportionately large for this small piece of logic.

I removed the dedicated test and the helper function, and kept the fix inline in the profile-saving path. The code now always closes the saved profile file, preserves write errors over close errors, and only reports success when both operations succeed.

aalexand (approved):

PR #1023, opened 2026-08-29
driver: pass an empty window title to start on Windows

HuzaifaAbdulRehman:
## What is broken

On Windows, `pprof -web` and `weblist` never open the graph when the temp file
path contains a space, which is every account like `C:\Users\Ada Lovelace\...`.
A blank console window appears instead, titled with that path, and pprof exits 0
without printing anything. The Go toolchain vendors this code at
`src/cmd/vendor/github.com/google/pprof/internal/driver/commands.go:350`, so
`go tool pprof -web` behaves the same way.

## Root cause

`commands.go:350` returns the Windows browser command as `cmd /c start`. The
call sites in `invokeVisualizer` and `openBrowser` split it on spaces and append
the target, and `os/exec` quotes any argument containing a space, so `cmd`
receives:

```
cmd /c start "C:\Users\Ada Lovelace\AppData\Local\Temp\pprof001.svg"
```

`start` reads its first quoted argument as the window title. The path is
consumed as a title, a console opens, and the file never reaches a handler.
`Start()` returns nil, so pprof records a success.

## What changed

The Windows command carries a trailing space, so splitting it yields the empty
title argument ahead of the target. `browsers()` delegates to `browsersForOS`
so the expansion can be tested from any platform.

Nothing exported changes. Non-Windows commands are unchanged, which the test
pins for darwin.

## Test plan

`TestBrowsersForOS` pins what each command expands to, and fails without the
trailing space:

```
$ go test ./internal/driver/ -run TestBrowsersForOS -count=1
--- FAIL: TestBrowsersForOS (0.00s)
    commands_test.go:31: browsersForOS("windows") expands to ["cmd" "/c" "start" "C:\Users\Ada Lovelace\...\pprof001.svg"]; want ["cmd" "/c" "start" "" "C:\Users\Ada Lovelace\...\pprof001.svg"]
```

End to end on Windows 10, Go 1.26.5: driving `invokeVisualizer` with `%TMP%` set
to a directory containing a space, the target is never opened before the change
and opens after. `gofmt -d -s .`, `go vet -all ./...`, `staticcheck ./...`,
`golangci-lint run -D errcheck` and `go test -race ./internal/driver/` are
clean, and the full suite matches the pre-change baseline.


aalexand:
The fix is pretty invasive so I'd love to find a simpler option.

Can you try this one line change:

```go
	case "windows":
		// Trailing space causes strings.Split(v, " ") to produce an empty argument (""),
		// which 'start' requires as the window title so paths with spaces are not consumed as the title.
		cmds = append(cmds, "cmd /c start ")
```

?

aalexand:
Or alternatively a local hack like

```go
args := strings.Split(v, " ")
if len(args) == 0 {
    continue
}
if runtime.GOOS == "windows" && v == "cmd /c start" {
    args = append(args, "")
}
viewer := exec.Command(args[0], append(args[1:], target)...)
```

HuzaifaAbdulRehman:
Switched to your one line version, trailing space and comment. `webui.go` is
untouched again, so the change is 10 lines plus a test.

One deviation worth flagging: `browsers()` now calls `browsersForOS(goos)`, four
lines, so the test can check the Windows expansion while running on Linux. Gated
on `runtime.GOOS` it would never run in this CI. The test fails if the trailing
space is ever deleted, which seemed worth guarding. Happy to drop the seam if
you would rather keep the literal one liner.

Verified on Windows 10 by opening a file under a `%TMP%` path containing a
space. vet, staticcheck, golangci-lint and the race tests are clean.

HuzaifaAbdulRehman:
Correction to the description, which I have edited: I wrote that pprof blocks in
`viewer.Wait()` until the stray console window is closed. That is wrong. Running
the unpatched binary with `%TMP%` under a path with a space, `weblist` exits 0
after about three seconds and leaves the console window behind. The hang I saw
came from `go test` holding the inherited pipe, not from pprof.

The rest stands: the target is never opened, nothing is printed, and the window
title is the path that should have been opened.

aalexand (approved):

Issue #707, opened 2022-06-26
pprof source view overcounts recursive functions


kvakil:
Please answer these questions before submitting your issue. Thanks!

### What version of pprof are you using?
`go tool pprof` with go1.18.3.

<details>
<summary>Full <code>go env</code> output</summary>
<pre>
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/keyhan/.cache/go-build"
GOENV="/home/keyhan/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/keyhan/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/keyhan/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/snap/go/9848"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/snap/go/9848/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.18.3"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/dev/null"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build2938140119=/tmp/go-build -gno-record-gcc-switches"
</pre>
</details>

### What operating system and processor architecture are you using?

Linux, x86-64 Intel.

### What did you do?

I opened the attached profile with `go tool pprof -http=: fac.pb.gz`,
and navigated to the "Sources" page. Here is the output:

![Screenshot 2022-06-26 at 13-03-34 e cpu](https://user-images.githubusercontent.com/22647613/175831921-c2df2c83-5486-495f-9bfc-5746594d1f06.png)

As you can see, it is claimed that `fac` took 6193% of the total time.
It is also said that it took 150 seconds, which is wrong since the
program took only a couple of seconds.

[...]

PR #1030, opened 2026-09-21
internal/report: count each address once per sample in weblist

Hee-San:
Updates #707. This is the first of two PRs; the analysis is in https://github.com/google/pprof/issues/707#issuecomment-5758777928.

`newSourcePrinter` did `cum[addr] += value` for every frame of every sample. With recursion the same address is on the stack many times within one sample, so the sample was added to that address once per frame. On a CPU profile of a recursive `fib` (total 310ms), one 10ms sample with 29 `fib` frames added 190ms to one call instruction and 90ms to the other.

This counts each address at most once per sample, the same way `newGraph` does with its `seenNode` set. It affects `-weblist` and the web UI Source page, which both go through `MakeWebList`.

It looks like a regression from #599 (cbba55b). Before that, weblist went through `rpt.newGraph` and got the deduplication for free. The README documents the rule this restores: "Samples that include a location multiple times (e.g. for recursive functions) are counted only once per location."

Effect on the `fib` profile:

```
                          before              after
weblist header            8.55s  2758.06%     900ms  290.32%
call fib(n-1)             5.20s               300ms
call fib(n-2)             3.05s               300ms
```

The values after the change are the ones `-disasm` already reports for the same profile. They are still above 100% because the header is a sum over instructions that one recursive sample contributes to more than once. That is a separate cause with a separate fix (the second PR), so this one says `Updates`, not `Fixes`.

What doesn't change: `flat`, which is only added for the leaf frame; weblist output for profiles without recursion, which I checked is byte-for-byte identical on testdata/sample.bin + sample.cpu; and every other view, since nothing else uses `newSourcePrinter`.

The test has two samples. In one the recursive call address is the leaf as well as an outer frame, which also pins down that `flat` is still counted when the address has already been seen. Moving the `flat` update inside the new `if` would make it fail.


Hee-San:
Screenshots of `-weblist main.fib` with the assembly expanded. They are from macOS, where llvm-symbolizer gives no line numbers for Go binaries, so all instructions land under one entry.
prA-before
prA-after

